PR #32, opened 2016-01-11
Update Instagram.php

FFServices:
- added getUserFollowings
- added users list pagination 
if a user list (getUserFollowers or getUserFollowings) > 200,than there is a next_max_id in the end of the array, so you can pass it to the next request and it will continue from the previous request (user 201,202,..)

### DIFF
--- a/src/Instagram.php
+++ b/src/Instagram.php
@@ -856,6 +856,26 @@ class Instagram {
     return $popularFeed;
   }
 
+/**
+  * Get user followings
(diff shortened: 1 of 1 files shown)


milesrichardson:
I am scraping a friend list of size > 200 and it only ever returns 101 results. I do not see next_max_id value in the response (I've intercepted the traffic to confirm). How do you get the next_max_id value?

PR #53, opened 2016-02-08
Patch 1

cloudever:


mgp25:
@cloudever got your idea, but it needs some fixes + the short login validation commit doesnt work, it required to be like that.

Regards

PR #80, opened 2016-03-07
Better DeviceID generation

### LINKED ISSUE #79: User with infinity phones
This is bad idea, on every single request change phone id.
It suspicious. And detectable.

Enelar:
Fixing #79

### DIFF
--- a/src/Instagram.php
+++ b/src/Instagram.php
@@ -5,7 +5,7 @@ require_once 'InstagramException.php';
 
 class Instagram
 {
-    protected $username;            // Instagram username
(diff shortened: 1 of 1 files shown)


djstanley:
Have you implemented a fix that uses a static uuid per "device"? Are the fixes you've made above sufficient? Don't want to get banned whilst following users.

Enelar:
@djstanley i do not know. I could make uuid static, but i not sure if i should to.
Maybe instagramm expects new uuid per each login. I do not know.

If something forcing you, i could investigate it. On weekends.

PR #81, opened 2016-03-07
Improve coding style

Enelar:
DRY, and other ancient knowledge. Just first step.


mgp25:
@Enelar instead of creating `enforced request` to throw an exception, better add it the request function.

Regards

PR #100, opened 2016-03-17
Downgrade php version dependency

Koc:
No sense require 5.6, it works ok on 5.5.9


mgp25:
All the tests i do is under 5.6, this way i can guarantee all users that the API works under this PHP version, because i wont support any other, thats why.

Regards

PR #120, opened 2016-04-03
Update Version

wehmoen:
Update to Instagram v.7.20.0 to avoid "Your Instagram app is outdated" Messages


mgp25:
@wehmoen you need to update also IG SIG KEY, otherwise it wont work

Regards

PR #121, opened 2016-04-03
Fix typo, remove line.

RageBrahimi:
1. Typo in `function direct_share();` Replace `$recipeint_users` with `$recipient_users` as it prevented direct share to assigned recipient(s) **(would only send to self)**.

2. `'caption_text'   => $captionText,` unnecessary in `function deleteComment();`


Mamedov1985:
@RageBrahimi When did you get the commentId ??? 
I tried the command getMediaComments($mediaId) and some other comments but did not see this parameter.

PR #154, opened 2016-05-03
It's duplicated and make confusion

mkanker:
It was duplicated with the complete method: getUserFollowings($usernameId, $maxid = null)
Everyone should use only this, because the only way to get all users is using the maxid concept.


mgp25:
@mkanker different functions, the one with parameters if for the self user, and the other function for the rest of the users, notice the endpoint.

Maybe im wrong, since i dont remember the response, let me know if you try it

Regards

mkanker:
You're right, I guess we could just name the method with
GetSelfFollowing()  then to don't make confused, what do u think?

On Tue, May 3, 2016, 03:27 mgp25 <notifications@github.com> wrote:

> @mkanker <https://github.com/mkanker> different functions, the one with
> parameters if for the self user, and the other function for the rest of the
> users, notice the endpoint.
>
> Maybe im wrong, since i dont remember the response, let me know if you try
> it
>
> Regards
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/mgp25/Instagram-API/pull/154#issuecomment-216448356>
>
-- 
Michel Ank

mgp25:
@mkanker oks, send a pull request

PR #155, opened 2016-05-03
Changes the method name to doesn't confuse

mkanker:


BurakDev:
@mgp25 I think you need update the Wiki

PR #165, opened 2016-05-06
Feature added separate signup method without calling other api methods

ruscon:
BC

your login() method sometime return $login[1]; sometime null;
but not used anywhere after return

i return boolean true/false for signUp
false - if user are already logged in

and return $this for login() for chaning


ruscon:
BC

your login() method sometime return $login[1]; sometime null;

i return boolean true/false
false - if user are already logged in

mgp25:
https://github.com/mgp25/Instagram-API/commit/8039c9345bec7c39c9432092e7fea9ae1dbbcd29

PR #235, opened 2016-06-24
Fix query string based errors on search

ihsangulumser:
basicly, space (" ") support for location search and fb user search.


mgp25:
@ihsangulumser have you tried this and worked?

ihsangulumser:
did you see any problems at code?
both urlencode and rawurlencode working fine at my personel projects

PR #248, opened 2016-07-04
Support get and set $ch

vkill:
Hi @mgp25 

In this PR, I add two functions(getCh and setCh) into Instagram class.
Now, we can set http proxy for $ch .
Please review it. thanks.


techinpark:
Very good update.

mgp25:
Although is good, i dont want proxy support in the api. The aim is to avoid spammers

Regards

PR #281, opened 2016-07-18
Added constant to ffmpeg path and adapted functions to it

lucasalveslm:
I was having some problems with ffmpeg path on OSX, so I managed to use a constant to let the ffmpeg path configurable 😄


mgp25:
am using OS X and i dont face that issue :/

Regards

PR #298, opened 2016-07-26
Update IG SIG KEY + version

RiTu1337:


GuilhermeFM:
I've tested the new key here and its not working.

mgp25:
@RiTu1337 it doesnt work to me either, please check it

Issue #302, opened 2016-07-27
Video cropped to square on mobile devices after uploading


mewlabs:
This is my comment from #180, but seems that there it remains without attention :(

Please capture video configure request after tapping the "format icon", to specify landscape orientation instead of square. It would be great, because without it, uploading of video is useless in most cases, it displays ok in web, but cropped to square on mobile devices...

I can capture traffic in app up to v7.0.0 but they don't have this feature. I guess that it should be just several additional options on video configure step. Thanks!

mewlabs:
Found reason: there is no need to create square icon for video, because in this case video will be cropped to square on mobile devices. PR created

PR #305, opened 2016-07-28
#302 Video cropped to square on mobile devices after uploading

mewlabs:
There is no need to create square icon for video, because in this case video will be cropped to square on mobile devices


mgp25:
@mewlabs pushed commit

Regards

PR #381, opened 2016-08-23
Response for getProfileData() and editProfile()

jpmoney5065:


jpmoney5065:
I did the get function as one catch-all function instead of typing out 19 different functions, so it's a little different from the rest of the code.

It works like this:
```
$pf_data = $i->getProfileData();

$pf_data->getProperty('username')
$pf_data->getProperty('phone_number')
$pf_data->getProperty('has_anonymous_profile_picture')
```
etc, etc

mgp25:
@jpmoney5065 im going to merge this PR and add the functions so it match with the other classes, once we finish doing all changes in the API we can improve the calls and make a better interface

Regards

mgp25:
@jpmoney5065 for now ill leave like that, but we need to improve all the classes etc and make better interface

PR #416, opened 2016-09-04
Add missing getter.

aryanbaghi:


mgp25:
@aryanbaghi hello, thanks for the PR. Can you instead of `getPk()` rename it to `getUsernameId()` or `getMediaId()`

Media ids are a much longer string that username id just so you can difference them.

Regards

Issue #485, opened 2016-09-21
Continue) A hashtag searching problem


secretcau:
Hi. I wrote a hashtag searching problem before.
And then, i have researched the problem.

Now, i am facing a problem related the hashtag searching problem.

> I can search when i logged-in for uploaded photo's owner account to instagram.
> However, I couldn't find if i logged-in for other account.  

Now, i posted the issue to instagram cs team. But i'm not expect a solution from their.

Sorry to my poor English.

jpmoney5065:
No solution right now. Someone will get to it when they get to it.

secretcau:
Uhm... Isn't it my mistake? Your means 'Current searching problem is normality at now although i set by the standard using this API'. Is it right?

secretcau:
@jpmoney5065 I found definite difference.
I could get it **without** log-in on instagram web-page.
But, I couldn't get it **with** log-in instagram web.

Is there a clue?

mgp25:
#470

Issue #470, opened 2016-09-15
Publications and hastags


weeh89:
Hi, I make the publication by this script.
There are about two weeks hashtags publications are not public lists
_Example: https://www.instagram.com/explore/tags/HASHTAG/_

**If you are logged in with the same account that published, or is not logged in to any account the image appears correctly in the hashtag list.

If you are logged in with another user the image does not appear to him in the hashtag list**

Is there any way for this fix to this problem?

mgp25:
True, i just tested it.

Yes, there is a way to fix it. When someone or i is in the mood to work on that we'll send a pull request, meanwhile do not open more issues, we know the thing and thats it.

- Is there any way for this fix to this problem?

Yes, research, compare, try and error until success

Regards

oratiostuff:
we are working to fix

also already for more than 14 days I'm trying to get this setting

honeydewkorea:
@weeh89 Hello. i have a issue that yours.
Did you solve the problem? Could you help me?

PR #488, opened 2016-09-21
To solve #470, #485

secretcau:
I use old version because relation of my web site.
Uploading function is well done, but i get a return seraching hashtag
with login on instagram public app or web.

But i could get all list without login. Why it is occured?
Is it any solution. I have wander to this issue


mgp25:
@secretcau there is many not useful code in this PR. Send or post here the update

Regards

honeydewkorea:
@mgp25 I know but these codes are related to my app. However I sended this pull request because i understood that you mentioned posting pull request to help in #470. Am i mistake?!

PR #519, opened 2016-09-28
Extract  overall average image color.

beshoo1:
When class fix photo ratio to meat instagram rules , it create a white background color , in general it is ok, but what if we get overall average image color. and make it as background color ! it will be more suitable.

hope this help !


mgp25:
@beshoo1 ill improve this when i finish some few things

Thanks anyway!

PR #536, opened 2016-10-01
Check array elements exist

shadowofb:
Check array elements exist


mgp25:
@shadowofb not all vars require isset, rebase it and send it again

Regards

shadowofb:
@mgp25 how to understand what variables need to be checked and which are not?

PR #551, opened 2016-10-05
Edit media

Vasiliy-Bondarenko:
EditMediaResponse was not working. 
Fixed 2 errors.

### DIFF
--- a/src/Instagram.php
+++ b/src/Instagram.php
@@ -750,7 +750,7 @@ class Instagram
             'caption_text' => $captionText,
         ]);
 
-        return new EditMediaResponse($this->http->request("media/$mediaId/edit_media/", SignatureUtils::generateSignature($data))[1]['media']);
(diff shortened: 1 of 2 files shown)


mgp25:
@Vasiliy-Bondarenko full response should be managed by `EditMediaResponse`

```php
return new EditMediaResponse($this->http->request("media/$mediaId/edit_media/", SignatureUtils::generateSignature($data))[1]); 
```

If you want, make changed to the class instead of changing the code i quoted above

Regards

Vasiliy-Bondarenko:
it's not possible to just edit the EditMediaResponse.
if you don't give full response to the class EditMediaResponse - how it could manage it? :)

it was 
```php
return new EditMediaResponse($this->http->request("media/$mediaId/edit_media/", SignatureUtils::generateSignature($data))[1]['media']);
```
it's only a part of the response. for example it does not have status, which is needed by EditMediaResponse. so i removed ['media'] at the end.

PR #690, opened 2016-11-22
Fix getUsernameId

frrakn:
I don't understand the point of adding the JSONMapper. Kills code readability and introduces bugs like this.


frrakn:
Can you elaborate? I'm not getting any on my side.

PR #730, opened 2016-12-03
Fix video upload id issue

msmoosa:
`uploadVideo` function uses an `uploadId` implementation that generates the following value `1.48073646786E+12`. Such a value is rejected by instagram api and returns 500 internal server error.

The number needs to be formatted as done in uploadPhoto and other places.

This PR refactors the working uploadId implementation using `number_format` into a new `getUploadId()` and consumes this wherever being used including `uploadVideo`


msmoosa:
Not quite sure what the styleCI error is

mgp25:
@msmoosa im going to close the PR, but im adding this. Im going to move the function to `Utils.php` instead

Regards

msmoosa:
Cool

PR #770, opened 2016-12-12
Update X-IG-Capabilities header for v10.1.0

ping:
New value for for 10.1.0/10.2.0


ping:
@RiTu1337 This repo is already using the 10.1.0 key since f9a24d9730012e4ec3bcbb8c8c43b701a86bd052

RiTu1337:
@ping ok you are right. and it is a correct header.

PR #884, opened 2017-01-27
fixes weird rare bug during editProfile

Overtorment:
Would not mind if anyone could test this out.


mgp25:
@Overtorment It doesnt happen to me, anyway if you face it again, send more information about it and ill research it

Regards

PR #893, opened 2017-01-30
same type as old_stories

Vasiliy-Bondarenko:
this hint is used by jsonmapper


frrakn:
Haven't inspected, but would `friend_request_stories` also need the same comment?

PR #902, opened 2017-01-31
Allow pagination for getUserTags function

yuvalino:
this basically just adds the parameters, but on top of that, enables JSON_BIGINT_AS_STRING in the json_decode function.

The json_decode change is required since the returned next_max_id for the getusertags request is a very long numerical string - without JSON_BIGINT_AS_STRING, the code will convert the ID to int and turn it into an exponential number due to 32bit limitations in ints.

### DIFF
--- a/src/Instagram.php
+++ b/src/Instagram.php
@@ -1157,8 +1157,8 @@ class Instagram
         return $this->request("usertags/$usernameId/feed/")
         ->addParams('rank_token', $this->rank_token)
         ->addParams('ranked_content', 'true')
-		->addParams('max_id', (!is_null($maxid) ? $maxid : ''))
-		->addParams('min_timestamp', (!is_null($minTimestamp) ? $minTimestamp : ''))
+	->addParams('max_id', (!is_null($maxid) ? $maxid : ''))
+	->addParams('min_timestamp', (!is_null($minTimestamp) ? $minTimestamp : ''))
         ->getResponse(new UsertagsResponse());
     }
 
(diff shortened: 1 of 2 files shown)


monajalal:
Can you please suggest correction so that this function will store tags in $tags array?
```
 47     $tag_helper = null;
 48     $tags = [];
 49
 50     do {
 51         if (is_null($tag_helper)) {
 52             $tag_helper = $i->getUserTags(53112486);
 53         } else {
 54             $tag_helper = $i->getUserTags($tag_helper->getNextMaxId());
 55         }
 56         $tags = array_merge($tags, $tag_helper->getUser());
 57     } while (!is_null($tag_helper->getNextMaxId()));
 58
 59     foreach ($tags as $tag){
 60         echo '- '.$tag->getTagInfo()."\n";
 61     }
```

I was not sure what to add here `$tags = array_merge($tags, $tag_helper->getUser());` instead of `getUser()` and couldn't find an example of usage of getUserTags on the Web. I tried to use it like `getUserFollowers` function as below:
```
29     $helper = null;
 30     $followers = [];
 31
 32     do {
 33         if (is_null($helper)) {
 34             $helper = $i->getSelfUserFollowers();
 35         } else {
 36             $helper = $i->getSelfUserFollowers($helper->getNextMaxId());
 37         }
 38
 39         $followers = array_merge($followers, $helper->getUsers());
 40     } while (!is_null($helper->getNextMaxId()));
 41
 42     echo "My followers: \n";
 43     foreach ($followers as $follower) {
 44        # echo '- '.$follower->getUsername()."\n";
 45     }
```

Issue #910, opened 2017-02-02
Path to settings files is broken.


Vasiliy-Bondarenko:
This change is breaking whole system!!!

https://github.com/mgp25/Instagram-API/commit/b7a222397fb227fbfe800ce93a8cad1666da724b
```
const DATA_DIR = __DIR__.DIRECTORY_SEPARATOR.'data'.DIRECTORY_SEPARATOR;
```
and in current codebase... 
```
$this->setting = new SettingsFile($username, Constants::DATA_DIR);
```

You fixed the path to settings file, and nobody can change it because it's a constant and you give no way to supply different path.
This path points to vendors/... folder and will be rewritten after every update and all settings for all accounts will disappear. 
Am i missing something or it's a huge fail????

mgp25:
No reason to be dramatic, go back a couple commit until i fix it

Regards

Vasiliy-Bondarenko:
Well, I have some trouble with users today. So it's kind of dramatic for me :))
Thanx.
You make some breaking changes. It's hard to keep up :)
I think i will end up writing tests for everything... I can not test manually every time or just throw the code to production and hope it will work :)

mgp25:
Rolled back to previous code, does it works now?

Vasiliy-Bondarenko:
yes. sure.

mgp25:
@Vasiliy-Bondarenko oks let me know if everything is okay now

Vasiliy-Bondarenko:
well, it's working! thanx.

mgp25:
@Vasiliy-Bondarenko ill be more careful next time so users dont merge code that breaks backwards compatibility

Regards

Vasiliy-Bondarenko:
@mgp25 fixed. now i can update and switch the path. actually you renamed it - it was settingsAdopter :) but it's an easy fix. no problem.
Thanks a lot. You doing a great job. Really appreciated.

frrakn:
Out of curiosity - how were you setting this?

I set it as a constant because aside from actually changing the code I didn't see a way to pass in the settings path. I'm happy to fix my commit - just not sure where you were passing this in.

Vasiliy-Bondarenko:
@frrakn I just extend Instagram class and making all the tweaks i need. In this case - change public property settingsAdapter.

```
class Insta2 extends Instagram
{
    /**
     * @var Account
     */
    protected $acc;
    /**
     * @var Instagram
     */
    protected $insta;

[...]

PR #913, opened 2017-02-03
Fix backwards compatibility issues

frrakn:
Addresses #910


mgp25:
@frrakn i created a dev branch, so we can test things there before pushing it to master

PR #914, opened 2017-02-03
Check response not null before all JSONMapper calls, throw EMPTY_RESPONSE exception

frrakn:
@mgp25 wasn't able to find a dev branch, but I'm ok if you want to make one and merge this to it

### DIFF
--- a/src/Instagram.php
+++ b/src/Instagram.php
@@ -33,7 +33,6 @@ class Instagram
     public function __construct($debug = false, $truncatedDebug = false, $settingsAdapter = null)
     {
         self::$instance = $this;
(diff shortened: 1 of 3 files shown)


src/Instagram.php
```diff
@@ -33,7 +33,6 @@ class Instagram
     public function __construct($debug = false, $truncatedDebug = false, $settingsAdapter = null)
     {
         self::$instance = $this;
-        $this->mapper = new \JsonMapper();
```
frrakn: grepped the codebase - this isn't used nor is listed as a variable in the class declaration.

Issue #919, opened 2017-02-05
Class 'InstagramAPI\Settings' not found


mangledbottles:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer completely.
- Put an `x` into all the boxes [ ] relevant to your issue (like so [x]).
- Use the *Preview* tab to see how your issue will actually look like.

---

### Before submitting an issue make sure you have:
- [x] [Searched](https://github.com/mgp25/Instagram-API/search?type=Issues) the bugtracker for similar issues including **closed** ones
- [x] [Read the FAQ](https://github.com/mgp25/Instagram-API/wiki/FAQ)
- [x] [Read the wiki](https://github.com/mgp25/Instagram-API/wiki)
- [x] [Reviewed the examples](https://github.com/mgp25/Instagram-API/tree/master/examples)
- [x] [Installed the api using ``composer``](https://github.com/mgp25/Instagram-API#installation)

### Purpose of your issue?
- [x] Bug report (encountered problems/errors)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

Code:

```php

include __DIR__.'../../../../autoload.php';
require_once __DIR__.'/../src/ChallengeSMS.php';

$debug = true;

echo "#####################\n";
echo "#                   #\n";
echo "#   SMS CHALLENGE   #\n";
echo "#                   #\n";
echo "#####################\n";

echo "\n\nYour username: ";
$username = trim(fgets(STDIN));
echo $username."\n";

if ($username == '') {
    echo "\n\nYou have to set your username\n";
    exit();
}

echo "\n\nYour settings path folder ([ENTER] if dedault): ";
$settingsPath = trim(fgets(STDIN));

if ($settingsPath == '') {
    $settingsPath = null;
}

$c = new \InstagramAPI\ChallengeSMS($username, $settingsPath, $debug);

$c->startChallenge();

while ($c->getStep() < 3) {
    switch ($c->getStep()) {
        case 1:
            echo "\n\nInsert Phone number: ";
            $code = trim(fgets(STDIN));
            $c->setPhone($code);
            break;

        case 2:
            echo "\n\nInsert Security Code (leavy empty to reset): ";
            $code = trim(fgets(STDIN));
            if ('' == $code) {
                $c->reset();
            } else {
                $c->setCode($code);
            }
            break;

        default:
            echo "No function for this! Press Y to reset, enter to retry\n";
[...]

PR #920, opened 2017-02-05
Update src/ChallengeSMS.php - Add Settings Class

mangledbottles:
Added settings class that was missing


frrakn:
I see this is part of the issue report #919, tagging so it's linked :)

PR #926, opened 2017-02-07
Media comments : pagination issue

JulesDe:
Add of the rank token parameter for the `getMediaComments` function
It seems like it is mandatory to ensure a working pagination of the results.


mgp25:
@JulesDe have you checked that is doesnt work without this?

mgp25:
@JulesDe i think its okay now since nobody has opened an issue related to this yet. I'll let you know if this is required in the future.

Python version is a port of this one, so maybe it isnt at the same version yet

Regards

PR #970, opened 2017-02-24
Improving settings adapter

ilyk:
This pull request adds ability to create your own settings adapters.
The main rules are
1. Implement `InstagramAPI\SettingsAdapter\SettingsInterface` interface
2. Specify *settingsAdapter* array key **type = custom** and **class = className** of needed adapter;

example:
```
$instagram = new \InstagramAPI\Instagram(true, false, [
    'type'  => 'custom',
    'class' => \InstagramAPI\SettingsAdapter\Memcached::class
]);
```

### DIFF
--- a/examples/usingMemcachedSettingsAdapter.php
+++ b/examples/usingMemcachedSettingsAdapter.php
@@ -1,0 +1,31 @@
+<?php
+
+include __DIR__.'/../vendor/autoload.php';
+
+
+// Example using Memcached for storing settings and cookies
+$i = new \InstagramAPI\Instagram(true, false, [
+    'type'  => 'custom',
+    'class' => \InstagramAPI\SettingsAdapter\Memcached::class,
+    'persistent_id' => 'instagram',
+    'memcache_options' => [
+        Memcached::OPT_PREFIX_KEY => 'settings_'
+    ],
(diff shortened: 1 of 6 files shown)


mgp25:
@ilyk is this line okay?

```php
$settings = new $customClass($username, $config);
```
Can you give more details?

```php
'class' => \InstagramAPI\SettingsAdapter\Memcached::class
```

You can change this `\InstagramAPI\SettingsAdapter\Memcached::class` to any class you want to manage settings?

And you give me more detailed instructions, i will add this info to the wiki

Regards

ilyk:
@mgp25 
1. yep, this is correct for sure: http://php.net/manual/en/language.variables.variable.php
2. *::class* allows for fully qualified class name resolution at compile: http://php.net/manual/en/language.oop5.constants.php
3. I can give you detailed instructions just tell what kind of details you need :)

the fact is that you simply can 
1. create a class which implements given interface
2. put full class name (with namespace) into 'class' key of settings array
3. the settings property will contain object of the specified class

mgp25:
@ilyk can you add to this PR an example file (places in examples folder)? That would help users using this feature

ilyk:
Example added

PR #982, opened 2017-02-27
Select inerface

ildarsaitkulov:
added select output network interface + curl options


mgp25:
@ildarsaitkulov due to the latest http interface rewriting, this need a rebase

Regards

PR #988, opened 2017-02-27
update Dockerfile

mattes:
* switch to ubuntu:xenial for easy ffmpeg installation
* remove comments


ilyk:
what's the problem to install `ffmpeg` on debian instead of ubuntu?

mattes:
In debian:jessie ffmpeg is only available in backports. On top of that `php:7.0-apache` installs apache, which is not needed.

SteveJobzniak:
Docker is not something I've ever used so I cannot review this contribution.

We need more feedback from docker users explaining why this is a good idea. To me it looks like the new Dockerfile contributed here forces users to be running Ubuntu Xenial and that's terrible if that's the case. There are thousands of Linux distros.

SteveJobzniak:
Again: This won't get merged without more explanation. So we'll have to close this pull request for now.

PR #1010, opened 2017-03-03
Pass PDO into settings adapter

jpmoney5065:
Basic usage, passed into main contructor: `['type'=>'mysql' , 'pdo'=>$yourPdoVariable]`
To change from default schema:  `['type'=>'mysql' , 'pdo'=>$yourPdoVariable , 'schema'=>'differentDB']`


jpmoney5065:
Should be 100% backwards compatible

SteveJobzniak:
Thank you very much for doing this work.

I am reviewing the code, and see some issues.

1. Move the PDO class/connection back into SettingsMysql.php. Meaning, restore its connect() function which you deleted. It's better that way. We shouldn't clutter the SettingsAdapter.php file with per-backend things like how to connect to them. And your change also removed its connection fallbacks such as "localhost", "instagram", etc.

2. You have some unrelated change in autoInstall(). Not sure what that is about. Please do that as a separate pull request to keep the tasks separate.

3. Change the SettingsMysql constructor to accepting something like an array, as follows: `['type'=>'pdo', 'pdo'=>$pdo]` and `['type'=>'regular', 'username'=>..., 'password'=>..., 'host'=>..., 'dbName'=>...]`.

4. Make SettingsAdapter use the appropriate SettingsMysql constructor value in point 3, depending on whether the user passed a PDO object or regular connection parameters.

PR #1012, opened 2017-03-03
Pass PDO to settings adapter

jpmoney5065:
`$settingsAdapter = ['type'=>'mysql', 'pdo'=>$pdo]`

- Change in `autoInstall()` is because we don't know the db name when we pass the in the PDO variable, so we need to look it up.


SteveJobzniak:
@jpmoney5065 Please test https://github.com/mgp25/Instagram-API/commit/839e772f893f822f5f77885bd24b6157aa00fdf8. I don't have MySQL on this machine so I wrote it blind but tried to be sure that it'll act exactly like in the past if given legacy parameters. And if you give it the "pdo" config option it should only use PDO.

And you should also specify the dbName as a config option, even in your PDO case. So no risky "select database()".

PR #1043, opened 2017-03-07
Phpdoc fixes

SubodhDahal:
Fix parameter order, add missing parameters in function docblocks


SteveJobzniak:
Thank you so much for taking the time to go through all of this. Years of sloppy contributions where people rewrote functions without updating the docblock has lead to this. I've gone in and edited a few things to make sure it's up to date and to remove a bug you had introduced (a parameter was set to "20" instead of null for some reason).

All good now. Thanks a lot!

PR #1049, opened 2017-03-08
Added new objects and props to classes

zuexo:
Added Broadcast object and few props to other objects.
Also added SuggestedUsers object


SteveJobzniak:
@zuexo Thanks a lot for contributing all these new response objects! Some of them need typefixes but I'll take care of it after the current work.

For future reference:

Every ID (almost everything with "pk" or "id" in keyname): string, to avoid overflowing int/floats

Every lat & lng: float, for proper location precision.

SteveJobzniak:
Here's how it was fixed:

https://github.com/mgp25/Instagram-API/commit/68c07cf22d300add8f037b7a6176486976d2be6b

Thanks again for running with the apiDeveloperDebug mode and finding all of these in the first place. It certainly helped a lot!

zuexo:
My contribution is all thanks to you. You changed api debuging to be less complex.
And for the future I'll keep that in mind and not forget setting types in annotations.

SteveJobzniak:
@zuexo Happy to hear that. And thank you very much. I'm really loving your contributions. :)

SteveJobzniak:
@zuexo By the way, you should really fix your Git config. Everything you submit is being contributed as `Matej Mihevc <Matej Mihevc>` which means it cannot be clicked to lead to your GitHub profile.

Do `git config user.name zuexo` and `git config user.email zuexo@users.noreply.github.com` while in your local copy of our repo. :)

Or fix it globally on your whole machine by adding `git config --global` to those commands above.

zuexo:
Thanks for the info. It probably broke when I started using Gitlab. :)
Will fix it right now!

SteveJobzniak:
@zuexo We've been looking at your contributions. You've clearly got intelligence and a sense for quality code. We decided to invite you as a collaborator so that you can commit directly to the project with any other improvements you come up with. 👍

PR #1051, opened 2017-03-09
Add new properties to response objects

zuexo:
New properties and new objects in connection with ExploreResponse.
Also added few other properties to Reel.


SteveJobzniak:
@zuexo Excellent work, thanks a lot! And all fields have correct types this time! <3

PR #1054, opened 2017-03-10
cleanImageUrl helper function

joseflorido:
Helper function to remove ig_cache_key from the image URLs.


SteveJobzniak:
Hi. Thanks but this has nothing to do with the API. If you want the ability to remove that from your URLs in your own end-user project, then feel free to have such code there.

I have personally got similar code in my own personal project. But I use http_build_url and parse_url instead, to parse it, unset the query portion, and rebuild it. Which is much better than this hardcoded string name thing you did.

But anyway, it's unfortunately not something that we need in this project. It's for the API. Not for tricks like editing URLs. :\ :)

joseflorido:
Thanks. I thought maybe someone could need this too, but I understand your
point about not being part of the API.

SteveJobzniak:
@joseflorido I appreciate the good intentions. :-)

PR #1060, opened 2017-03-10
Add optional count parameter for searchFBLocation

gepo:
fbsearch/places supports count parameter. If pass something like 1000 it returns more than 30 items (77 in my tests).

### DIFF
--- a/src/Instagram.php
+++ b/src/Instagram.php
@@ -2084,17 +2084,23 @@ class Instagram
      * Search for Facebook locations by name.
      *
      * @param string $query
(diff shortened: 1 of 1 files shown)


SteveJobzniak:
Thank you for discovering this. This looks good and clean to me.

Thoughts, @mgp25 ? We'll have to be sure that this is how the official client does it, to avoid being detected as a fake client.

mgp25:
The number of items returned in the response are the same as the value set in `count`, if you set a big number and there are less items, only those items will be returned.

PR #1069, opened 2017-03-11
Fix getting adapter from options array

Alfredao:
I just fixed this typing error :)


SteveJobzniak:
@Alfredao Thank you! Nothing to complain about here. That was a pure typo. :D

PR #1080, opened 2017-03-14
use random UUID for directShare

valga:
UUID used as boundary in directShare must be random. Otherwise, it will be impossible to send more than one message per thread: Instagram silently rejects subsequent messages.


SteveJobzniak:
@valga Interesting discovery... @mgp25 knows all about the boundary parameter and what it's used for. I am sure he had a reason to use the phone's UUID instead of a random UUID every time. So let's wait for his analysis.

dyar95:
@SteveJobzniak This is just the boundary for a multipart form. It should be random every single time :-)

valga:
@SteveJobzniak, it was random (per session) until a23fcbccf9eb707f56d1cb716e95b2861b21abba. It seems, in some cases (like boundaries) we need to generate random UUIDs per request.

@dyar95, nope, it is also an extra field in POST data.

dyar95:
Looks like a normal boundary for multipart/form-data to me. In the case of `client_context` it adds it to the body as well, but that's about it.

valga:
I have double-verified the issue, and found that `client_context` must be random per every request. Static boundary is ok.
I will open new pr.

SteveJobzniak:
@valga The commit you pointed to is ultra important. Before, Instagram would shadowban you (mark you as an "open proxy") if they saw too many device IDs from a single IP. So we made UUID + Device ID static per-account.

Anyway, @mgp25 is the one who reverse engineered the API so he knows why he made boundary static. Going to guess because the real client always made it static. He takes a lot of care to act exactly like the client.

SteveJobzniak:
The new PR (for reference): https://github.com/mgp25/Instagram-API/pull/1083

SteveJobzniak:
Okay I see what went on now, explained here:

https://github.com/mgp25/Instagram-API/commit/eb0714be5c8d3b5bd5dfded1232eccee715c5369

https://github.com/mgp25/Instagram-API/commit/22c936ee9b3769b46ec3e5b68c08e9ac51f12a18

PR #1083, opened 2017-03-14
use random client_context in directShare

valga:


dyar95:
Verified, this is correct. Should be merged asap.

Issue #1072, opened 2017-03-12
Uploaded image isn't in an allowed aspect ratio :: 1.6.1


beshoo1:
@mgp25  I am optimizing the resize class that i sent  you before. but i need to understand what is the correct allowed aspect ratio that instagram likes .

since while i am trying to upload  640x320 return an error !!

```
POST:  media/configure/
DATA: ig_sig_key_version=4&signed_body=778e3f3d227a681b7b680b414e812d699621024a2e2eb8ef8ef1d5a6c8481464.{"_csrftoken":"OLHpWCvpUzYnuY1NIXNYhvs2oeSP5zS6","media_folder":"Instagram","source_type":4,"_uid":"3986477616","_uuid":"fcea4419-e1b1-434d-9cb4-e6592e1f62ec","caption":"","upload_id":"1489328979093","device":{"manufacturer":"Fly","model":"FS405","android_version":18,"android_release":"4.3"},"edits":{"crop_original_size":[640,640],"crop_center":[0,0],"crop_zoom":1},"extra":{"source_width":640,"source_height":640}}
← 400    80B
RESPONSE: {"status": "fail", "message": "Uploaded image isn't in an allowed aspect ratio"}



What is the best guideline  for this ?

SteveJobzniak:
640x320 is ratio 2:1. Instagram allows 1:1 and SOME slightly wider aspects, but definitely not media that is TWICE as wide as it is tall.

Instagram places all media in 1:1 aspect ratio boxes (squares). Recently they added support for landscape and portrait (non-square) images.

The longest side is never allowed to anything other than these dimensions: `16x16`, `30x30`, `32x32`, `48x48`, `50x50`, `60x60`, `64x64`, `75x75`, `80x80`, `96x96`, `100x100`, `128x128`, `150x150`, `160x160`, `192x192`, `200x200`, `240x240`, `320x320`, `480x480`, `600x600`, `640x640`, `750x750`, `800x800`, `960x960`, `1080x1080`. There _may_ be more dimensions that I haven't discovered, but it's unlikely.

That means that the allowed 1:1 square aspect ratios are the EXACT dimensions above.

If an image does not exactly fit that size, then it must be resized and possibly also cropped.

Uneven aspects (portrait/landscape) are allowed to have one side shorter than that. But by how much? I don't know. Perhaps 1.6:1 is the max they allow. So 1:1 to 1.6:1.

[...]

PR #1085, opened 2017-03-15
ImageAutoResizer Class

valga:
see #1072


SteveJobzniak:
@valga Excellent work. Thank you! I will read it in depth soon.

valga:
@SteveJobzniak, ok. 

I made some adjustments to the code. Main one is to use 87 as JPEG quality, because 100 produces huge files, and also we have `{"lib_name":"jt","lib_version":"1.3.0","quality":"87"}` string somewhere in code.

SteveJobzniak:
@valga Agreed, 100 is overkill. But I suggest 95. That's the most common "almost perfect quality but not as big as 100" setting.

Instagram will recompress our media after we upload it, so better give it closer to 100% quality.

Don't mind what our headers say in another place, because they cannot see from our jpgs what quality we used when compressing a file.

SteveJobzniak:
@valga Another thing that worries me is the use of tempnam. That one will fail if the user has 65535 files in their temp folder, which some users do. We had bug reports from some people who had too many temp folder.

I would prefer if you add a final $tempFolder = null parameter, and if null use tempnam(sys_get_temp_dir()), otherwise use tempnam($this->_tempFolder). That way the user can control where images get stored and manually delete them all from that folder via some other code if they need.

So set quality to 95 and make temp folder configurable. I'll review this more in depth when I have time. Very busy atm.

valga:
@SteveJobzniak, no worries. I've also added another cool feature - detect image orientation and keep it if needed to.

SteveJobzniak:
@valga Fantastic! Great discovery about the image orientation, which would be lost during the EXIF loss from the recompression otherwise.

I'll probably have time to do a full review and test today.

SteveJobzniak:
@valga Fantastic! Great discovery about the image orientation, which would be lost during the EXIF loss from the recompression otherwise.

I'll probably have time to do a full review and test today. There's no rush to merge this since 2.x itself isn't ready yet, so let me know when you feel like it's ready for general use and I'll go through the code.

valga:
@SteveJobzniak, I am out of ideas for fixing or adding something, so I think it is ready for review.

SteveJobzniak:
@valga I am super busy working on the critical User-Agent fix, but will review this after that.

I did find something interesting which helps though:

https://help.instagram.com/1469029763400082

> Once you share it, the photo or video will appear as a portrait or landscape in Feed. You can share photos and videos with aspect ratios between 1.91:1 and 4:5.

SteveJobzniak:
@valga I am super busy for a few days working on the critical User-Agent fix, but will review this after that.

I did find something interesting which helps though:

https://help.instagram.com/1469029763400082

> Once you share it, the photo or video will appear as a portrait or landscape in Feed. You can share photos and videos with aspect ratios between 1.91:1 and 4:5.

That's for portrait/landscape content. Square content ratio is always 1:1.

So allowed photo ratios are:

- Exactly 1:1.
- Between 1.91:1 and 4:5.

I suggest adding this to the code/function comments somewhere. Quote the supported ratios and add a "@see https://help.instagram.com/1469029763400082". Perhaps even make the function handle ratios outside that automatically (I don't have time to read the code to see if it already does that).

SteveJobzniak:
@valga Excellent! :-) I didn't have time to review yet but I did two small tweaks here via github, which you'll want to pull if you are going to do other edits.

SteveJobzniak:
It's looking like I'll only have time to review this tomorrow, because the user agent solution research was slow and excruciating work, and 5 hours later I am finally onto the implementation steps.

valga:
@SteveJobzniak, does Instagram have a limit for minimum image size? If yes, what should we do when image is too small after all manipulations?

And for e59c756 I think that ratio range is more important than tmp directory.

Also, it turned out that "Orientated" images are tough.

SteveJobzniak:
@valga

> does Instagram have a limit for minimum image size? If yes, what should we do when image is too small after all manipulations?

Good thing you thought about that! Minimum allowed width is 320 and maximum allowed width is 1080.

https://help.instagram.com/1631821640426723

_"When you share a photo that has a width between 320 and 1080 pixels, we keep that photo at its original resolution as long as the photo's aspect ratio is between 1.91:1 and 4:5 (a height between 566 and 1350 pixels with a width of 1080 pixels)."_

_"If you share a photo at a lower resolution, we enlarge it to a width of 320 pixels. If you share a photo at a higher resolution, we size it down to a width of 1080 pixels."_

And I am pretty sure that they require us (the "app") to do those resizes before uploading. Their server refuses media outside the specs.

> And for e59c756 I think that ratio range is more important than tmp directory.

Wasn't the ratio range only configurable because we hadn't confirmed Instagram's range for sure? Now we know the exact range, and don't need those parameters at all. Unless they're serving some other purpose (like intentionally controlling cropping ratio?). In that case yeah move tmpfolder to the last position.

> Also, it turned out that "Orientated" images are tough.

Are they? There's just an EXIF attribute. Read it, rotate the pixelbuffer if it's meant to be rotated. If that EXIF is missing, we don't rotate at all. What's the issue? :)

SteveJobzniak:
With the huge Device rewrite done, I'm ready to review this, try the code, read through it all in depth, etc. It's the first task I'll get to when I wake up. So please make sure to do any necessary edits in time for that. ;-)

valga:
@SteveJobzniak 
> And I am pretty sure that they require us (the "app") to do those resizes before uploading. Their server refuses media outside the specs.

Hmm, I will verify it later. If this is true, we need to make some adjustments.

> Unless they're serving some other purpose (like intentionally controlling cropping ratio?).

Yep, they are for setting specific cropping ratio(s), like squaring images, etc.

> What's the issue? :)

There is no issue, if you do pre-rotation, i.e. read exif, rotate original if needed, crop-resize it. But rotation doubles memory usage for image in some cases (like 90 degrees turn). Assuming, that original are bigger than output in general, post-rotation is preferred. So, I read exif, crop-resize original and rotate output (if needed). And it is tough. Or post-rotation is overkill, I don't know.

SteveJobzniak:
> Hmm, I will verify it later. If this is true, we need to make some adjustments.

Seems that way. A guy tried to upload a 640x320 image and got this API reply:

`{"status": "fail", "message": "Uploaded image isn't in an allowed aspect ratio"}`.

By the way, Instagram's server-side aspect ratio verification formula is:

```
aspect_ratio = width/height
ok: aspect_ratio >= 0.8 && aspect_ratio <= 1.91
reject: anything else
```

And furthermore it seems (https://help.instagram.com/1631821640426723) that they also verify image size as follows:

```
ok: width >= 320 && width <= 1080
reject: anything else

**Comment by SteveJobzniak** [comment]
> Yep, they are for setting specific cropping ratio(s), like squaring images, etc.

Ahh that's fantastic. I moved those parameters back.

> There is no issue, if you do pre-rotation, i.e. read exif, rotate original if needed, crop-resize it. But rotation doubles memory usage for image in some cases (like 90 degrees turn). Assuming, that original are bigger than output in general, post-rotation is preferred. So, I read exif, crop-resize original and rotate output (if needed). And it is tough. Or post-rotation is overkill, I don't know.

I love how you're paying attention to details like that. That's what I do as well and it's rare to see other developers with that much attention to detail. You've done a fantastic job with this class. I look forward to reviewing and merging it today after I wake up.

As for the post-rotate: Perhaps we can just achieve it by setting the rotation flag in EXIF in the output image instead of actually rotating any pixels at all. I have not checked if PHP's image functions can append EXIF though.

**Comment by valga** [comment]
>Perhaps we can just achieve it by setting the rotation flag in EXIF in the output image instead of actually rotating any pixels at all.

We can't just set rotation flag, as we do user-specified cropping (cropFocus), so when we are asked to crop image from top, we need to know, where top is actually located.

Without cropFocus (crop always from center) we can just set orientation flag.

**Comment by valga** [comment]
@SteveJobzniak, I have checked limits on image dimensions and found, that Instagram accepts both 1x1 and 1280x1280 images (at least for timeline). So, the only thing matters is aspect ratio (although I did not find largest accepted resolution, but see below).

However, I enforce a limit of 1280 pixels to largest side to keep output file size sane.

**Comment by SteveJobzniak** [comment]
@valga Thank you for all of this hard work. I will now download this class, read through it and test it in depth.

**Comment by SteveJobzniak** [comment]
Review nearly complete. You've done a great job. I'm just fixing some bugs in the aspect ratio calculation. It's not right.

```php
        if ($aspectRatio > 1 && $width > self::MAX_WIDTH) {
            $width = self::MAX_WIDTH;
            $height = floor($width / $aspectRatio);
        } elseif ($height > self::MAX_HEIGHT) {
            $height = self::MAX_HEIGHT;
            $width = floor($height * $aspectRatio);
        }
```

Does not handle < 1, does not handle square, etc.

valga:
Yep, square images have been incorrectly resized to 1035 (instead of 1080). It is fixed now.

valga:
Also, some GD functions can result in fatal error, so they have been unmuted.

SteveJobzniak:
@valga Alright. I didn't see your edit while I did a rewrite. I've rewritten it to just strictly follow Instagram's specification, and to always derive height from width&aspect (the way they do it). And commented things clearly for future contributors.

I also found 2 more bugs in other areas, will fix now.

SteveJobzniak:
Alright @valga the second issue was that it's pointless to "imagefill" with white, since we never display the background anyway. And the "imagecreate" functions always create a pre-filled black background.

But I'm looking at http://php.net/imagerotate and they're requiring a bgcolor attribute. So I guess the only fix necessary is to remove the pointless and wasteful "imagefill()" call.

Thoughts on that?

SteveJobzniak:
As you see here:

http://php.net/imagecreatetruecolor

> imagecreatetruecolor() returns an image identifier representing a **black** image of the specified size.

There's no reason to use this "white" imagefill stuff. Especially since we never see the white.

SteveJobzniak:
To use black in rotation, we simply have to "imagerotate($source, $degrees, 0)".

valga:
White is used as nice background for transparent png/gif images. Perhaps we should call `imagefill` only if source image has transparency.

SteveJobzniak:
Ohh. Interesting idea. Instagram themselves don't support transparency. Only JPGs. So if the user inputs a PNG/GIF file, we'd want something nice instead of black. I agree with that decision.

But no, it's not worth searching for transparency. GD has no easy way to check that. You must scan every single pixel, look it up in the palette, and check the alpha of that color, hehe. Definitely way more costly than an imagefill() call, which is probably very cheap.

I agree with the decision to keep imagefill() then. But I'll add some comment to explain why it's there.

SteveJobzniak:
@valga Alright I am sure we've caught every issue now. Everything looks clean and works perfectly. I'll do a squashed merge (as a single commit) if you're happy with the results, to avoid bloating the history. :-)

Thank you very much for doing this work! It's excellent. It was something we really needed and was unavoidable in time for 2.x, and you helping out like this certainly took off pressure from all the other hard work I've been doing every day for the past 3 weeks. I'm so happy that 2.x is nearly ready for release now. Phew. We're getting there!

valga:
It turned out that `imagefill` is super costly (runs like 0.1s), so I replaced it with `imagefilledrectangle` (0.0001s).

SteveJobzniak:
@valga Good catch! imagefilledrectangle() just draws a box instead of slowly flood-filling and checking every pixel. Nice.

SteveJobzniak:
Woops about the consts. ;-)

valga:
You may append `if ($width > self::MAX_WIDTH) {` to upper `else` after cefb72c

SteveJobzniak:
@valga Are you sure about that last change? Looks like a bug. Due to that change, nothing is verifying the aspect ratio anymore whenever the width is less than max_width.

SteveJobzniak:
Or perhaps it's always done perfectly in those minAspect/maxAspect checks higher up?

SteveJobzniak:
Hmm I am not sure but maybe a bug was introduced by that last change after all?

The maxAspect check higher up just changes the width, not the height: https://github.com/mgp25/Instagram-API/commit/cefb72c4755562d97172857176d46f2bd8dc27db#diff-ee4e09f3eadece98106a4290dd42f48aL484

So with the last commit, if the maxAspect codepath is taken, then the $height value never gets changed.

SteveJobzniak:
Either way I would say that the last change is dangerous against people's future contributions and patches. If someone edits the max/min aspect algorithms, and doesn't realize how tightly they're tied to the resize calculations, then they'll break the class. A simple floor(width / aspect) call is something we can do a few billion of per second, so let's move it back just to always be sure we'll get a result in the correct aspect ratio.

valga:
> So let's move it back just to always be sure we'll get a result in the correct aspect ratio.

Ok.

valga:
And I have nothing to say, except for it's ready to be merged.

SteveJobzniak:
@valga Awesome. Thank you very much for all your work on this class. Срасибо. Very fun class to use and excellent job with the algorithm! :-) I've been resizing a picture of kittens all evening. ;-)

jpmoney5065:
https://s-media-cache-ak0.pinimg.com/236x/4d/f5/0d/4df50d14122d1ea02311bbc1e924be01.jpg

The lurkers appreciate the work!

SteveJobzniak:
@jpmoney5065 Haha. Glad you like it. Have fun! :-D

PR #1099, opened 2017-03-19
Add two extra Instagram response errors

jpmoney5065:
11. Following the max limit of users
 12. Not authorized to view user


SteveJobzniak:
Hi. Thanks! I need to think about this... The current exceptions are all critical errors. General errors like those you're adding are usually handled by simply checking the `$e->getMessage()` value. But of course, the exception code only runs when an API returns an error, which may not be very often in well-written user applications, so it should be fine to add even more regular expressions. But I'll have to think for a moment.

SteveJobzniak:
Hi. Thanks! I need to think about this... The current exceptions it auto-maps to values are all critical errors. General FUNCTION-SPECIFIC errors like those you're adding are usually handled by simply checking the `$e->getMessage()` value to see if it contains a certain string. Regular expressions are slow. But of course, the exception constructor code only runs when an API returns an error, which may not be very often in well-written user applications, so it should be fine to add even more regular expressions. But I'll have to think for a moment.

For example, I personally haven't added errors like "User does not exist." Instead, my own personal application simply checks $e->getMessage() and does a string-search for that, to find specific problems with specific API calls that failed.

We may possibly have to reject these and suggest the exact same solution there. I'm leaning towards that.

jpmoney5065:
Ok, I understand the logic...

I was catching the `\InstagramException` errors this app throws then also catching the `code 0: UNKNOWN` errors via $e->getMessage, so was thinking any error that the Instagram server throws our way, we should try and get with the api.

SteveJobzniak:
@jpmoney5065 It would be nice to have all codes, but we don't control their server's messages, and Instagram doesn't give us any errorcodes, so we can literally only parse strings and hope for the best. And those strings can change at any time. So the less we hardcode the better.

SteveJobzniak:
Generally the usage right now is as follows:

Catch InstagramException. If IG_API_THROTTLED (the most common critical error), handle that. If not, check getMessage() to look for API-specific problems such as "User not found" in the message string.

jpmoney5065:
Not doubting your judgements, but throwing my hat in the ring for users like me, that in about a year of using this API, have never -not once- received a `throttled` or `429` error... but have had dozen of accounts caught on different types of checkpoints or disabled for whatever reasons... And sent way too many selfies of myself to IG HQ holding my username on a 'handwritten piece of paper'  to prove that I am a real person.

SteveJobzniak:
@jpmoney5065 That's because 1.x of the API sucked and didn't tell the users about throttling. Instead it just got stuck in a loop and kept spamming Instagram's server with the same request over and over until it succeeded (which could take anywhere from 5 seconds to an hour). That was really dumb, so throttling was changed to an exception that the API user _must_ handle instead. We don't want users to overload Instagram and we don't want the API to get stuck in invisible loops for minutes at a time. So while working on 2.x I introduced the throttling exception.

As for the old exception system we discussed earlier in this thread, I actually completely rewrote it from the ground up tonight, to get away from the stupid "everything globbed into 1 exception, using its getCode" madness.

https://github.com/mgp25/Instagram-API/tree/master/src/Exception

Now you can catch ThrottledException very easily without having to check getCode.

And the one you want to catch is FunctionException, which is for general functions. getMessage on it to get the actual error from the server. Check strpos() on it for various strings that you're interested in.

The rule about critical core exceptions remains. But all of those are now individual exceptions.

If you don't care about exception type, just catch \Exception or \InstagramAPI\Exception\InstagramException, since everything derives from those.

Enjoy.

SteveJobzniak:
Hey just FYI: 2.x is super close to release-ready now, and I decided to revise the new FunctionException name to a less clashy name.

https://github.com/mgp25/Instagram-API/commit/8687d0460d4039297aeca4052ca7160440167581

PR #1100, opened 2017-03-19
Improve README.md

lex111:


SteveJobzniak:
Thank you! 👍  :)

lex111:
Okay, thanks, usually put a dollar sign to show that this is a command performed in the terminal, it's certainly a matter of taste, but again, so often do, although this makes copying difficult.

SteveJobzniak:
@lex111 Yeah `$` makes copying harder, and it's [not common](https://github.com/guzzle/guzzle) to include it. And if someone doesn't understand that something is a terminal command then we really don't want them as users, because they'll just flood the ticket system with things like "What is PHP?". We get _tons_ of that kind of stupidity here already on a daily basis, just look at all the closed tickets marked as "invalid" or "not related", it's 90% of all tickets. ;-)

PR #1105, opened 2017-03-20
RealtimeClient Implementation

valga:
#### What is RealtimeClient?
In short, it is like Push Notifications service: when something interesting happens, you will receive a notification from Instagram. It is used mostly for Direct Messages. Also, it is bi-directional: starting from April 2017 Instagram uses RealtimeClient for sending text messages to chats.

#### How can I listen to incoming messages?
We use [Événement](https://github.com/igorw/evenement) library. It is very simple to use, check out [an example](https://github.com/igorw/evenement#adding-listeners).

#### Where can I find a list of supported events and their parameters?
You can find the list of events in `src/Realtime.php`. Up-to-date parameters for all events are in `examples/realtimeClient.php`

#### I checked out an example. What is that loop magic?
We use [ReactPHP](http://reactphp.org/). It is based on something like JavaScript Event Loop: it allows you to write async code easily. Of course, it comes with some overhead, but it also offers an ease of use.

#### But wait, PHP is single-threaded.
Yes, you are right. And your event listeners will block everything else, so don't run any time-consuming tasks there. Also, you should write asynchronous code to interact with all third-party servers.

#### I started your example, and it just writes some debug info like forever. Is it ok?
It is totally ok. RealtimeClient is designed to run as a daemon. So it never stops, until something bad (like auth error) happens.

#### How can I interact with it, if it runs as daemon?
Check out `examples/realtimeHttp.php`. It offers a simple HTTP server, that implements some commands like:
```bash
# send typing notification to thread 123
$ curl -i 'http://127.0.0.1:1307/activity?threadId=123&flag=1'
# remove typing notification from thread 123
$ curl -i 'http://127.0.0.1:1307/activity?threadId=123&flag=0'
# stop realtimeclient
$ curl -i 'http://127.0.0.1:1307/stop'
```
Of course, you can use other methods, like Pub/Sub with Redis, etc. Just attach you favorite client to main loop and control everything.

#### Does it support proxy?
Yes, it supports HTTP(S) CONNECT and SOCKS proxies.

#### What happens when network connection is gone?
We will keep trying to reconnect with Instagram.


mgp25:
I have also this, but we need to talk about few things. Email on profile

Regards

SteveJobzniak:
Hi again @valga. The implementation is very clean, so this is definitely the right one to merge and work on!

Just ensure first to update it to follow the code format of the project:
https://github.com/mgp25/Instagram-API/blob/master/CONTRIBUTING.md

From a quick glance, you'll mainly need to change the functions so that their parameters follow the 1 argument per line rule which the rest of the code in this project uses. It's not enforced by StyleCI so that has to be done manually.

SteveJobzniak:
Also, since this is only 90% done and "may get people banned", I'd like to ensure that nobody merges this until it's 100% safe.

valga:
@SteveJobzniak, yeah, I remember about that. But speaking of code format, main thing is to write walls of documentation.

Also, overall progress is about 50%.

SteveJobzniak:
@valga Hehe, yes, your code itself is always very clean. <3 Just needs to have docs to explain unexpected/expected behaviors and exceptions.

But what you did in ImageAutoResizer is fine. A logical variable name + its data type is enough, when you did things like `@param int $src_x` etc. Nobody who is a programmer is going to misunderstand what that means.

And as for the new code here, I don't know what exceptions the functions throw (I mean the one from the 3rd party lirary), if any, but just document/catch those and change the function arguments to one per line. Everything else looked great at a quick glance. :-) As always!

mgp25:
@valga there are some constants you can place in `Constants.php` like:

```php
const PACKAGE_NAME = 'com.instagram.android';
const APPLICATION_NAME = 'InstagramForAndroid';
const CPU_ABI = 'armeabi-v7a:armeabi';

const APPLICATION_ID = '567067343352427';
const PLATFORM = 'android';
```

They are also used in some functions, in the latest commit i did you can notice:

```php
    const VERSION_CODE = '54119071';
```

Which is your `BUILD_VERSION` i guess. The reason i named it `VERSION_CODE` it is because there is a direct relation with it:

```
versionInfo:
  versionCode: '54119071'
  versionName: 10.15.0
```

I also added some Facebook Constants:

```php
    // Facebook Constants.
    const FACEBOOK_OTA_FIELDS = 'update%257Bdownload_uri%252Cdownload_uri_delta_base%252Cversion_code_delta_base%252Cdownload_uri_delta%252Cfallback_to_full_update%252Cfile_size_delta%252Cversion_code%252Cpublished_date%252Cfile_size%252Cota_bundle_type%252Cresources_checksum%257D';
    const FACEBOOK_ORCA_PROTOCOL_VERSION = 20150314;
    const FACEBOOK_ORCA_APPLICATION_ID = '124024574287414';
```

`FACEBOOK_OTA_FIELDS` is for a function that is used for updated over the air.

I also saw you are using a `APPLICATION_ID`, so maybe its the same as this one. I added the prefix `FACEBOOK_ORCA_` because thats how the APK reference it and makes more easy to differenciate it.

If you have any recommendation, send a PR and ill merge it.

Regards

valga:
@mgp25, I picked up your `VERSION_CODE` constant, thanks.

As for `APPLICATION_ID`, it seems they are using 2 different FB applications: official one (124...) and hidden one (567...). Hidden one is "Analytics Only", so may be it is the reason they use it in MQTT client.

Other constants are safe to include into `Constants.php`, except for `CPU_ABI`, as it is related with `Device` class. @SteveJobzniak, do you have any idea of how to include `CPU_ABI` into `Device` class safely?

mgp25:
@valga i pushed some commits to `Instagram.php` that makes conflicts with your PR. So you can update it.

I added two factor authentication.

Regards

valga:
@mgp25, yeah, I saw it. Two factor authentication is great!

By the way, RealtimeClient progress is about 80%.

aitte2:
Wow incredible work valga! Thank you! :)

MrLuit:
This looks very promising! I'm looking forward to see this being implemented.

SteveJobzniak:
@valga I'm temporarily home for easter. Saw your question now.

> Other constants are safe to include into Constants.php, except for CPU_ABI, as it is related with Device class. @SteveJobzniak, do you have any idea of how to include CPU_ABI into Device class safely?

Well, yeah that definitely belongs in the device class. It's a list of Android ABIs supported by the device. But I don't know how we can get that (safely) without actually having the devices.

User agents (which all other Device class info is built on) is easy, since those are logged from people who own the devices.

But the actual ABI list can only be queried if we have the device in our hands and make a test app to read the CPU_ABIS value...

Perhaps all modern devices have the same list of ABIs? Or perhaps it's simply based on the android version? In that case we could do something in the Device class which looks at android version and returns a list of ABIs for that version of Android. Hmm... Do you have any ideas?

Anyway, one more thing: I can see that we need to store 'experiments' in the database. Any reason why we can't just use the normal Experiments used by the regular login? If they differ, then I suggest naming the storage value "rtc_experiments" to not mix the two terms.

SteveJobzniak:
@valga Okay I've found the answer to CPU_ABI:

http://handstandsam.com/2016/01/28/determining-supported-processor-types-abis-for-an-android-device/

ABIs found on Android are in 3 categories:

ARM:
- armeabi
- armeabi-v7a (Typically also supports armeabi)
- arm64-v8a (Typically also supports armeabi and armeabi-v7a)

X86:
- x86
- x86_64 (Typically also supports x86)

MIPS:
- mips
- mips64 (Typically also supports mips)

And the author writes:

> “armeabi-v7a” is BY FAR the most widely seen ABI on Android. About 90+% of all phones made in the last few years supports this.  x86 is only on a few devices including the The Asus Zenphone 2, Genymotion Emulator and some Android Emulator images.   I’ve never even heard of any modern android device using a “mips” based processor.

That list above gives us the following conclusion: Ignore MIPS, Ignore X86, and note that the person said that 64-bit ARM typically always supports 32-bit ARM and old-school ARM too.

I've checked all of the GoodDevices and saw that heir Snapdragon, Samsung and Huawei CPUs are all 64-bit ARMv8a.

Meaning that our list for modern devices is as follows, and is identical for all devices:

`arm64-v8a, armeabi-v7a, armeabi` (in that order).\

This means we do not need per-device CPU_ABI values. They all use the same ones.

valga:
@SteveJobzniak 
> This means we do not need per-device CPU_ABI values.

But we need to change it with new devices added. So I will move `CPU_ABI` constant into `GoodDevices`.

> I can see that we need to store 'experiments' in the database, which adds like 4 kilobytes to the storage (no big deal but still weird). Any reason why we can't just use the normal Experiments constants used by the regular login or a new static RTC_Experiments constant, or perhaps just keeping the RTC_Experiments in memory every time the client connects instead of saving them to storage at all?

These are 2 different things: `Constants::EXPERIMENTS` is a list of experiments supported by application, and stored experiments are approved by Instagram on per-account basis. So the whole experiments thing is like a tool to test and roll out new features, used by Instagram. For, example, MQTT is in test phase, so it is unavailable for most accounts: they use WebSocket. Also, Instagram can tune experiments via params: for example, they can enable MQTT just for receiving.

Also, it seems, that official client syncs experiments every 2 hours, but I did not verify that yet.

Speaking of storage issue, we can save only experiments used in our code.

anonyzm:
```php
"repositories": [
 +        {
 +            "type": "git",
 +            "url": "https://github.com/valga/php-http-proxy-react"
 +        },
 +        {
 +            "type": "git",
 +            "url": "https://github.com/valga/Pawl"
 +        }
 +    ]
```
Don't u think it is okey to use your noname repos in this project? What if u abandon them?

Also u are using https://packagist.org/packages/binsoul/net-mqtt-client-react
It has 81 installs... really? Why not to take industry standart like rabbitmq and take a lot popular and well contributed https://github.com/php-amqplib/php-amqplib library?

valga:
@anonyzm

> Don't u think it is okey to use your noname repos in this project? What if u abandon them?

These repos are indeed temporarily and I will definitely abandon them. As soon as their owners accept my pull requests, I will switch to official packages.

> Why not to take industry standart like rabbitmq and take a lot popular and well contributed php-amqplib library?

I think you need to call Facebook and ask them about their choice of protocols.

anonyzm:
> I think you need to call Facebook and ask them about their choice of protocols.
I don't advice u to change protocol, but to use industry-standart-like packages like rabbitmq ampqlib, that also is able to work via mqtt.

SteveJobzniak:
> But we need to change it with new devices added. So I will move CPU_ABI constant into GoodDevices.

I'm not sure we need to ever change it, because 64-bit ARM is here to stay for a long time (probably the life of this project). But putting the constant in that file is fine since it's related to the devices. Do something like:

```php
class GoodDevices
{
    /**
     * List of supported binary architectures for the device CPUs.
     *
     * NOTE TO COLLABORATORS: Currently all devices use the same 64-bit ARM list,
     * but if future added devices have different CPU capabilities, this will need a rewrite
     * to be able to specify different CPU_ABI values per-device.
     *
     * @var string
     */
    const CPU_ABI = '...';

    /*
     * LAST-UPDATED: MARCH 2017.
     */
    const DEVICES = [
```

As for experiments: Ah I see, so each client is meant to keep track of it per-account lists. I see. Well, the encoding is pretty okay, and people won't get this multi-kilobyte blob if they don't run MQTT. :) It's okay. Do you think it should be named rtc_experiments or experiments in storage? I guess experiments, since you're saying it is a generic value for the whole account and isn't just for rtc.

SteveJobzniak:
@anonyzm Hmm:

"RabbitMQ is the most widely deployed open source message broker. With more than 35,000 production deployments of RabbitMQ world-wide at small startups and large enterprises, RabbitMQ is the most popular open source message broker. RabbitMQ is used at companies like: ... Instagram".

Funny seeing their name there.

valga:
@anonyzm 
> php-amqplib
This library is a pure PHP implementation of the **AMQP** protocol. 

So, it is useless for us, as we use different protocols.

RabbitMQ is a message broker, i.e. a **server**. This pull request is about Realtime**Client**. I see no reason to use message broker in this pull request.

@SteveJobzniak 

Here is an example of saved experiments: https://gist.github.com/valga/58262c9043ecfe9cee5e5928399ce6ec

Actually, people will get this blob after every `syncFeatures()` call, so I am thinking of some whitelist for stored experiments. Or we can just take rtc experiments and save them into `rtc_experiments`, I don't know.

But if we want to act like official client, we must check enabled features when needed. For example, experiments has `ig_android_2fac` option, which enables (or disables) 2-factor auth. So, if we just call 2-factor methods without any checks, we may look suspicious for Instagram.

SteveJobzniak:
@valga Ahh you hooked it into syncFeatures. Well it makes sense I guess.

But in that case *definitely* go for the whitelist approach, where _saveFeatures makes a copy of the specific array keys we care about (such as ig_android_2fac and the one for mqtt support). Because I'd estimate that most users have the default "file" backend and writing 4 kb of data every time a slightest change happens is pretty hefty.

So perhaps this would work for our needs:

- Add "experiments" to storagehandler's list of `PERSISTENT_KEYS` (as you've already done).
- Add a new const to StorageHandler: `EXPERIMENT_KEYS`, which is an array with `ig_android_2fac` and a few more we need to be able to check. That way we can expand that list over time
- Add two funcs to StorageHandler: `packExperiments($experiments)` and `unpackExperiments($experiments)`. The packer creates a new array, loops through `EXPERIMENT_KEYS` and looks for all of those keys in `$experiments`, copies them to the new array, then JSON-encodes, zips and base64 encodes (although zip+base64 may possibly lead to BIGGER data when we only save a few keys, so maybe only JSON-encode). And unpack does the opposite.
- Lastly, add an `$this->experiments` to the main `Instagram` class which holds the array of interesting per-account experiment keys, and which is updated every time syncFeatures happens.

Benefits of this design:

- We can watch any experiment keys we want and can find them all with a global `if (isset($this->experiments['somekey']))` and then check its value.
- We don't write 4kb of data to disk every time.
- We don't cause as many writes to disk in general, since with less experiment fields the data will often be the same, which means the storage backend detects no change in the value, so it won't have to write anything to disk.

Btw, about MQTT: I see now that RabbitMQ is just a server. I did a quick search and guess there isn't any huge MQTT client library, so the one with 81 installs may be the only choice...? I saw one with like 43 installs which is even worse. Doesn't seem like there is a big community of MQTT clients.

valga:
@SteveJobzniak 

I built RealtimeClient on top of the ReactPHP library. It simplifies development A LOT, but also it narrows the choice for 3rd party libraries. So I actually had to choose between one and half MQTT libraries. But despite of being over engineered, `binsoul/net-mqtt-client-react` is ok.

But sad thing about MQTT is we don't actually have MQTT. We have FMQTT (stands for Facebook MQTT). FMQTT v3 is compatible with MQTT v3, but FMQTT v4 is not compatible with MQTT v4. For now, Instagram uses v3 for RealtimeClient and v4 for analytics. But it can be changed in future.

SteveJobzniak:
@valga Ah I see, well ReactPHP is great. And I'd rather see an over-engineered library like binsoul's than something badly engineered. Let's just pray that FMQTT 4 is never used for the chat.

SteveJobzniak:
@valga I had some free time and merged the master changes to StorageHandler into your branch. So just remember to pull it later. ;)

valga:
@SteveJobzniak, I ended up with setter and getter for expriments in `StorageHandler`.

@mgp25, thanks for the tip. I can confirm, that they are using RealtimeClient for direct messages now. Just for text-only so far. 	f66ea14 adds support for this feature.

And I hope I will finish RTC soon.

SteveJobzniak:
@valga Hi. The changes to StorageHandler look good. And I did a test of json_encode vs json_encode+gzip+base64, and it seems that despite the 30% base64 overhead it still ended up smaller than just json_encode. Interesting...

As for CPU_ABI, it's in the right place but wrong value:

https://github.com/mgp25/Instagram-API/pull/1105/commits/a152241b05285430359d539a677f84cf765203c8#diff-4c27747c9d768037369035b7b146521eR104

The current value is for a 32-bit device (or for a 64-bit device queried about its 32-bit list). Was your constant captured on a 32-bit device?

For ARM64 the ABI list is `arm64-v8a, armeabi-v7a, armeabi` (in that order; it's 64-bit, 32-bit, and old 32-bit, respectively). So since it seems Instagram just puts the ABIs in a colon-separated list, I suspect that the correct string is supposed to be `const CPU_ABI = 'arm64-v8a:armeabi-v7a:armeabi';`.

This depends on how Instagram is querying the device hardware when making their list. Here's examples of what a 64-bit device says when queried in different ways:

```> getprop | grep abi                                           
[ro.product.cpu.abi]: [arm64-v8a]
[ro.product.cpu.abilist32]: [armeabi-v7a,armeabi]
[ro.product.cpu.abilist64]: [arm64-v8a]
[ro.product.cpu.abilist]: [arm64-v8a,armeabi-v7a,armeabi]
```

Anyway, I just want to say: You've been working incredibly hard on this, and everything so far looks amazing. There isn't much thanks in this community so I just want to thank you sincerely for this incredible work. Great code quality too. :-)

valga:
> I guess gzipping is a good thing?

GZIP is bad for empty experiments. So I am thinking about comparing both variants before save and using first char as a trigger like Z => gzipped, J => plain JSON. But it is not so important.

> Was your constant captured on a 32-bit device?

Nope, I have Snapdragon 820, which is ARMv8-A. So it is all about source. Instagram uses `Build.CPU_ABI` and `Build.CPU_ABI2`, so it is basically about APK, rather than device itself.

SteveJobzniak:
Hi @valga,

> I am thinking about comparing both variants before save and using first char as a trigger like Z => gzipped, J => plain JSON. But it is not so important.

That's a great idea. It would let us upgrade the storage format for experiments in the future if necessary. Please do that.

Perhaps in the far-future we could add a "P" method (packed) which is a bitmask of the enabled/disabled flags of the experiments we care about, etc. That solution would be just 4 bytes long or so. (The "P" character + 1 byte to describe which list of experiments are packed (what the packed value represents, since the list will grow over time) + 2 bytes to hold the bit-flags for those experiments). But that's a later question. The important thing is that having a leading character to trigger encoding detection is a great way to be able to change algorithm in the future, in a backwards-compatible way.

> Instagram uses Build.CPU_ABI and Build.CPU_ABI2, so it is basically about APK, rather than device itself.

Oh okay. And if I look at Android's source, https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/core/java/android/os/Build.java, I can see that CPU_ABI and CPU_ABI2 are "the names of the first (CPU_ABI) and second (CPU_ABI2) instruction sets of native code". And those values are set further down in that file, and are set to entry 0 and 1 of either "SUPPORTED_32_BIT_ABIS" (ro.product.cpu.abilist32) or "SUPPORTED_64_BIT_ABIS" (ro.product.cpu.abilist64). And it says that 32-bit processes will always see 32-bit ABI lists for backwards compatibility.

So as long as Instagram's app remains 32-bit (which it is now), then `armeabi-v7a:armeabi` will remain correct.

Okay so everything is correct. Excellent!

valga:
@SteveJobzniak, done.

@mgp25, I added some info for your wiki.

Also, overall progress is about 95%. I just need to double-verify some MQTT-related stuff.

SteveJobzniak:
@valga Brilliant work as always. This is looking awesome!

I also like the 32-bit hack (as far as "hacks" can be liked). It's a fascinating way of getting a 64-bit int JSON value instead of outputting the large number as a JSON string.

SteveJobzniak:
@valga One thing that's sorta-confusing is this:

```php
InstagramAPI\Realtime\Client
InstagramAPI\RealtimeClient
```

Can that be cleaned up somehow perhaps by renaming `InstagramAPI\RealtimeClient` to `InstagramAPI\Realtime` (not saying this is the best name, just something to consider).

Since that file "combines" all parts of the `InstagramAPI\Realtime\` namespace into a single usable master-interface, it's both a "client" and a "master/handler/whatever". So we can get a bit liberal with the name.

Here are some ideas:

```php
new InstagramAPI\Realtime();
new InstagramAPI\RealtimeLayer();
new InstagramAPI\RealtimeCore();
new InstagramAPI\RealtimeCommunication();
new InstagramAPI\RealtimeHandler();
new InstagramAPI\RealtimeConnection();
new InstagramAPI\RealtimeSocket();
```

I don't know which would be preferred. But I do know that it's a bit weird to have both `InstagramAPI\Realtime\Client` and `InstagramAPI\RealtimeClient`.

I'll leave the judgment call up to you. At least I fully agree with keeping the main-client into the main `InstagramAPI\` namespace, so that it's easy for users to reach it (as easy as `InstagramAPI\Instagram`, `InstagramAPI\InstagramID`, etc). :-)

valga:
@SteveJobzniak, I agree with you about current naming is being a bit confusing. Instagram app has `RealtimeClientManager` class name, because it really manages multiple realtime clients (WebSocket and MQTT). But I guess it is too long.

`Realtime` is the best choice from your list of suggestions. Will rename it just before merging.

SteveJobzniak:
@valga Ahh, `InstagramAPI\RealtimeClientManager()` makes sense but it's super long. So yeah I guess `InstagramAPI\Realtime()`. That'll be nice and short. :-)

SteveJobzniak:
@valga I saw the new commits. This is seriously impressive. Excellent work.

valga:
@mgp25, @SteveJobzniak, it is ready for your review.

SteveJobzniak:
@valga This is fantastic news. I should tell you that I am going back and forth to treatment and am too exhausted to check it now, but I will check it when I feel better. I am going to guess 3-5 days. I need sleep. I think you need rest too after this hard work. ;-)

mgp25:
@valga incredible work, so far seems everything is good! if you can add one more example, using for for example `markDirectThreadItemSeen`

And everything seems so clear that i think every user can understand it reading your code. Nevertheless if you want to add any info to the wiki, you are welcome.

10/10 

💯 

Regards

valga:
@mgp25, done. Also I think it is safe to include `sendDirectText` into examples, because it requires valid `threadId`. So, you can't open new threads with it.

SteveJobzniak:
@valga I'm getting closer to having enough energy to review and test this.

In the meantime I have a question... In our main class we have some direct sharing/messaging functions, which are causing troubles to users since they were sort of half-finished. They used "recipients" instead of thread_id concept, etc. And they don't work very well. Sometimes they can post/create a thread. Other times they fail, and their API wants a valid thread_id (but our functions only take a recipient list, which was the wrong implementation but we're stuck with it now).

I see that `sendDirectText` is implemented here in MQTT as a replacement. Are the other functions also able to run over MQTT instead? If you look in src/Instagram.php and search for the words "function direct" you'll see all photo/media sharing functions.

Here are the questions (for mgp25 too):

1. What do you think we should do with the old HTTP API direct-functions? As far as I am aware, the official app no longer uses those HTTP API endpoints for direct sharing/messaging at all.
2. Therefore, one option is that we can delete the old direct-functions and bump the API to 3.x (because deleting a function is a backwards-incompatible change, and some people use those current, broken "direct" functions already).
3. Or, we can try to fix the old functions by making them use `thread_id` as a parameter instead of `recipients`, but that would also be backwards-incompatible... So we'd still end up as 3.x if we do that.
4. Or, add them to the realtime client so that it handles all sharing too. Is that how the official app does it now?

Well, just some things to think about. It's an issue we'll have to figure out, and depending on what the best solution is we'll end up with Realtime released as either 2.1 or 3.0 depending on if the old HTTP api functions will be rewritten/rescued, or deleted/replaced with MQTT.

😱

valga:
In short, HTTP API is used as a fallback, when Realtime connection is not available (because of network issues, etc). So we must keep HTTP methods for sure. 

Also, Realtime has only a few methods implemented:
- mark item in direct thread as seen,
- indicate activity in direct thread,
- send text message to direct thread.

SteveJobzniak:
@valga Ahh okay, good to hear we don't have to throw away the old code. Thanks for that information. I'll think about it and try to figure out some way to patch the existing functions. It's probably going to end up being some _hopefully backwards-compatible_ change to the parameter lists of those functions.

mgp25:
@valga this is already done right? Check the conflicts of `composer.json`. Im going to review it this weekend, and if everything is okay, then ready to merge it!

Regards

valga:
@mgp25, yep, it is done. At least I have nothing to add or fix.

Conflict is resolved.

SteveJobzniak:
@valga Thank you, I will review this tomorrow (saturday) or sunday. Together with mgp25's review, we'll be ready to merge by sunday.

SteveJobzniak:
@valga I am downloading this now to test the code (all functions, examples, etc). Then I will do a readthrough of all changes. I'll be done in a few hours.

SteveJobzniak:
@valga Please verify if this interpretation is right in this commit:

https://github.com/mgp25/Instagram-API/pull/1105/commits/6ed536677d3aba470c93794dcf79760a46e1a0d2

Note that event-loop can use many backends:
```
react/event-loop suggests installing ext-libevent (>=0.1.0)
react/event-loop suggests installing ext-event (~1.0)
react/event-loop suggests installing ext-libev (*)
```

Including pure PHP.

SteveJobzniak:
Is this intentional or a typo somewhere?

```
[webs] Sending command X{...}
```

The "X" character precedes every JSON object output by the realtimeHttp daemon. I did a search for the X and can't find it anywhere in the code that outputs the word "Sending command", so just asking if you know what that's about.

valga:
> Including pure PHP.

Pure PHP implementation was buggy until 0.4.3 was released 3 weeks ago. This is the bug: https://github.com/reactphp/event-loop/issues/42

So it is ok to remove `ext-event` requirement.

> The "X" character precedes every JSON.

This is intentional. Each _user_ command must be prefixed with `X` when running WebSocket client. MQTT client doesn't require that.

SteveJobzniak:
Okay the "does it work?" review is now complete, to the extent that my health allows!

And I am very, **very** impressed! This is fantastic! So, examples/realtimeClient is basically how to write a low-level event daemon yourself (great for power users). And examples/realtimeHttp is a fantastic HTTP daemon which makes it trivial to call its localhost APIs from other scripts to do actions. That daemon is a great way to have an infinite event loop which handles all messages (which ties up that PHP process), without tying your own main script.

That's fantastic. Users who want specific needs could easily, for example, make their own modified realtimeHttp daemon which saves every realtime notification to a database, and then lets them call an API to retrieve all new events that have happened since timestamp X. So it's all the benefits of realtime communication, but with the ability to keep it in a separate PHP process and just get its info anytime you want via your own project's main script, by calling a local API.

That's really, really good! The examples are *perfect* and will really help people with all the tools they need for writing their own personal implementations.

You have REALLY created something very beautiful here! It's amazingly fun to see and send messages coming in realtime, and to get story/live video notifications instantly when they happen.

And I am very impressed with the experiment caching. Thanks to the whitelist, the experiemnts string for me is just 114 bytes (which is small enough to be practically "nothing"), and it doesn't seem like they're re-ordering the list of experiments, so that compressed string remains the same value all the time, which means that the experiments saving doesn't cause a bunch of needless settings storage rewrites every time we login (since the new compressed experiments value is the same as last time, so no write takes place :)).

This is 10/10. Perfect work. Everything is awesome about this. And your code quality is fantastic as always. It's so rare to meet perfectionists. You're awesome. :)

I will now read through the actual code, but apart from the ext-event thing, I don't think there'll be anything at all to do :)

valga:
A few things just popped out in my head:
1) We must use random `%ACCOUNT_ID%` and `%SESSION_ID%` placeholders in MQTT client, as Instagram may detect us.
2) Experiments sync must be done in random intervals.
3) MySQL storage handler uses cache, so it is cookie/settings loss is possible with 2 script running in parallel. It seems, we need to use `last_modified` column for cache refresh, like ```SELECT ... WHERE `last_modified` > :our_last_modified```.

And speaking of experiments, RealtimeClient has 5 modes controlled by them:
1) WebSocket: read/write, MQTT: don't connect.
2) WebSocket: read/write, MQTT: just connect.
3) WebSocket: write, MQTT: read.
4) WebSocket: read, MQTT: write.
5) WebSocket: just connect, MQTT: read/write.

Also, upcoming Direct Messaging update will use experiments too.

SteveJobzniak:
@valga

1:

I did worry a little bit about that earlier when we discussed the 32-bit JSON trick... Thinking that `json_encode('%ACCOUNT_ID%')` could possibly fail in some unexpected way and not properly replace the placeholder values, if it doesn't match what was put in the JSON blob. But then I thought "... but JSON is a strict format... strings are always encoded inside of double quotes, so it should be fine...". I don't know. If we use random numbers I'd worry more about accidentally replacing the text in more than 1 place.

I think I have a solution... 

```php
$accountId = (string) $this->_instagram->account_id;
$sessionId = (string) round($sessionId);
...
'u'                 => $accountId,
'mqtt_sid'          => $sessionId,
...
$result = strtr($result, [
    json_encode($accountId) => $accountId,
    json_encode($sessionId) => $sessionId,
]);
```

What do you think? This forces the values to be string, which means they get encoded as `"13123023"`. After that, it replaces the JSON-string encoded value (which must be inside of string quotes), with just the string without JSON-quotes. So something like `"13123023"` is replaced with `13123023`.

2. Randomizing intervals is a good idea. I guess just changing the "next interval" to be a random time instead of a fixed distance.

3. Damn you are totally right, the storage backends were written with heavy caching to avoid needless DB/disk writes. That of course makes parallel clients impossible since they would conflict with each other. Hmmmmmm...

valga:
1. I am worried about different thing. We run string replace against JSON that may contain data, provided by Instagram (such as `ig_mqtt_route` experiment, or something else in future). So, if Intagram gives us `%ACCOUNT_ID%` in there (very unlikely, but probability is not 0), we are caught. So, we must use something like ```'%ACCOUNT_ID_'.uniqid().'%'``` as placeholder.

3. The thing I am worried about is cookies. MQTT client uses `sessionid` cookie, which can be updated by another script, while RealtimeClient will never know about this update.

SteveJobzniak:
1. That's a good point, they could add our secret string and see us changing the value. Fixed: https://github.com/mgp25/Instagram-API/pull/1105/commits/d031c0c557fbe0bc4652990fa6b536af1529c497 (This func is just called once per connection so I used plain `.` string concatenation for code shortness).

2. It turns out I misunderstood experiments update. I see now that it happens during login-flow, and happens if it has been at least 7200 seconds since the last login-resume. I thought it repeatedly happened at regular intervals during MQTT connection. But since it only happens at `login()` call, then I don't think you have to randomize the "7200" thing, since users call login() at random intervals anyway. And besides, Instagram already detects that we're using the reversed API and not a real app client anyway. So it doesn't matter if they see us doing syncFeatures() at a certain interval. They already know. But they tolerate the private API users.

3. About the storage... Yes, if two clients are running in parallel, and one with outdated experiments has to write something else to disk, that write would also overwrite the newer experiments, or session, etc, stored by the other script. The storage handler wasn't made to share values with other clients, since we never had any need to do that before, since people used this either for very short-lived page serving scripts, or for long-running single "do a job" scripts, and we never really cached any settings that change between runs. This is the first time we've needed long-living concurrent scripts in a way where data may desync during concurrent uses... Damn...

So about number three... uhm... That's going to be a complex rewrite...

The current system only writes to disk/db whenever a new set() value differs from the cached value. So a sort of hellish hack would be these two steps:

1. Every time we are going to write, first read all old values from disk, then overwrite the value that was just changed (the one that triggered the write/save), and then write the whole blob to disk/db.

And 2. Uhhh.................. we would need to periodically read the file/database and refresh the in-memory cache *AND* **EVERY** variable that has ever done `get()` on those values.

Shit...

SteveJobzniak:
Oh and I just re-read:

> The thing I am worried about is cookies. MQTT client uses sessionid cookie, which can be updated by another script, while RealtimeClient will never know about this update.

Well, FileCookieJar solves that... because it reads/saves the cookies from disk on every request.

If the main concern is cookies, not the settings storage, does that mean it would be enough to change the Memcached and MySQL backends to always query the database for cookies before every request? But SHIT that's going to suck for performance. Every single library API call will need a MySQL query.

SteveJobzniak:
@valga Okay I finally finished reading through all changes. WHEW. Everything is clean and looks correct to me. And it worked when I ran it (my experiments put me on WebSocket, so I did not see MQTT but I assume it works too). This is definitely nuts. You've done an incredible job.

About the remaining issues we are discussing:

1. Experiment refresh randomization = Not necessary, for the reason I described. People already call login() at random personal intervals so it will usually be much more than 7200 seconds until the next time syncFeatures happens. And Instagram can already detect reversed API users vs real app users anyway. So don't worry about that.

2. Settings storage sharing between multiple scripts = Not possible. We can't keep track of every location in the code that has done a `get()` on some setting. Which is what we would have to do, to do settings-sharing. We would have to periodically read the latest settings from disk/db and update all of the code locations everywhere that has ever done a `get()` call (and many of those gets are saved in protected/private variables). It's just not worth it. Especially since most settings (token, experiments, etc), stays static and identical almost all the time. So multiple client executions simultaneously usually don't actually cause any disk write at all. And any outdated data would always self-correct itself, since all settings storage writes are atomic (both the db and file ones), which always guarantees that we will always have a "perfect, intact snapshot" of settings inside our settings file/db. Even if it may end up being from a more outdated concurrent PHP session. That's no big deal. If something is outdated, it'll just get updated next time another script runs again. So don't worry about settings-sharing between scripts.

3. Cookie sharing between multiple scripts = I was wrong above, I just checked Guzzle's FileCookieJar and it only loads cookies from disk *once* and then caches them in memory (exactly like the mysql cookie storage does), and it only writes the cookies to disk [when the object is destroyed](https://github.com/guzzle/guzzle/blob/master/src/Cookie/FileCookieJar.php).

So we couldn't fix this. Because we would have to create a new FileCookieJar / new CookieJar on *EVERY SINGLE* API call. Which means every call would: Read a file from disk/read the cookie string from a mysql database, json-decode it, build a cookie jar in memory, and attach it to Guzzle. Over and over again. Tons of disk IO and database reads/writes.

Hopefully the `sessionid` cookie is what I think, which means it stays the same all the time until someone calls logout() or their password changes.

And in that case, it's honestly up to the library users, to close down their realtime daemon, and restart it to get the latest cookies.

SteveJobzniak:
My review is complete. The code is very clean, follows project guidelines and works perfectly for me. And you've been testing it for a month or two, so I trust this code and I trust your skill. Merging.

So again, about the remaining issues:

1. Experiment refresh time randomization is not needed, so that's no issue.

2. Settings sharing between clients is not even possible (it would be a hell of `get()` variable syncing, so we'll not even go there, and usually all settings stay the same every time people login anyway), and any outdated disk/db cache which could perhaps get written to disk, would be updated the next time another client starts up and sees its outdated values and handles the upgrade again.

3. So the only issue is cookie sharing, and it's a very minor one. And it's not easily solvable. Because cookie jars are handled by Guzzle and it would kill performance to always build new objects in memory and attach new jars on every request. Especially since I don't think the session changes (since all clients always re-use the same token/session and just resume it). Furthermore, since each client keeps its own local in-memory cookie jar, every client *would* have its own "perfect", consistent internal state. It should be totally fine!

SteveJobzniak:
Congratulations @valga, it's finally merged! You've worked incredibly hard on this and I am very, very impressed. By putting it into the repo (it's dev-master, so no danger since we aren't putting anything in a stable release yet), people can start using it and contributing and we will get wider testing. I doubt there will be any issues since you've done a great, perfectionist job! But anything that comes up can be handled in a more chilled manner in other, small pull requests. You deserve a break now. ❤️

PR #1128, opened 2017-03-28
Update Dockerfile (add exif ext, ffmpeg, git)

daciwei:
Recently I try to use original Dockerfile in repo, but notice `docker build` don't working because lack of `ext-exif` and `git`.
I fixed it and also add `ffmpeg`. In original file there also no `ffmpeg` support.


SteveJobzniak:
@daciwei Thanks for the updated dependencies! Merged it and also cleaned up a bit:

https://github.com/mgp25/Instagram-API/blob/master/Dockerfile

Should still work but I haven't tested it.

PR #1142, opened 2017-04-02
ThreadItem Enhancements

valga:
1. List all available item types as constants.
2. Add `ActionLog` model, which is used for `action_log` items.

### DIFF
--- a/src/Response/Model/ActionLog.php
+++ b/src/Response/Model/ActionLog.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace InstagramAPI\Response\Model;
+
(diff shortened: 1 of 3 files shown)


mgp25:
@valga , @SteveJobzniak is gone, he will come back in 2-3 months. The PR seems okay, and you can merge it if you want. Is there any reason to list all types as constants? Are going to be used later?

Regards

valga:
@mgp25, I will miss him. 

Constants are just for reference (like those we have in `CarouselMedia`, `Item`, etc), so you can see all available message types and use them in subprojects, etc.

PR #1152, opened 2017-04-05
Fixed bugged login exceptions

aitte2:
Hello sir! :)

Recent commit 42beba3134ef9ef9c8b40e56406ba46bba2c94c4
added two factor login but broke the login exception system.

This fix reuses the same method of throwing as in Client.php
which makes proper exceptions thrown at login again, such as:

"Uncaught InstagramAPI\Exception\IncorrectPasswordException:
InstagramAPI\Response\LoginResponse: The password you entered is
incorrect. Please try again."

### DIFF
--- a/src/Exception/LoginException.php
+++ b/src/Exception/LoginException.php
@@ -1,7 +1,0 @@
-<?php
-
-namespace InstagramAPI\Exception;
-
-class LoginException extends RequestException
-{
-}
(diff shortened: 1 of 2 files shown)


aitte2:
When that is merged you can once again find bad passwords by simply doing:

```php
try {
    $i->login();
} catch (\InstagramAPI\Exception\IncorrectPasswordException $e) {
   // wrong password caught specifically (we can do others too
   // but you asked for password check)...
} catch (\InstagramAPI\Exception\InstagramException $e) {
   // all other login errors handled here...
}
```

mgp25:
@aitte2 perfect! 😄

PR #1182, opened 2017-04-16
login(): Fix and document return value

SteveJobzniak:
- Reverts 5d8be61e3c8b0e285a6a4ad2222dd879103548e9, since there
  was no way for us to always return a LoginResponse.

- Documents return value properly.


SteveJobzniak:
See comment here for extra information:

https://github.com/mgp25/Instagram-API/commit/5d8be61e3c8b0e285a6a4ad2222dd879103548e9#commitcomment-21779960

PR #1191, opened 2017-04-20
Instagram.php: Add getMediaLikersChrono

ranjbarhadi:
this new function does the same as getMediaLiker() but it returns a lighter and much simpler result


src/Instagram.php
```diff
@@ -2589,6 +2589,21 @@ public function getMediaLikers(
...
     /**
+     * Get a simplified version of users who liked a media item.
+     *
+     * @param string $mediaId The media ID in Instagram's internal format (ie "3482384834_43294").
+     *
+     * @throws \InstagramAPI\Exception\InstagramException
+     *
+     * @return \InstagramAPI\Response\MediaLikersResponse
+     */
+    public function getCronoMediaLikers(
```
matheuseduardo: seems like this method return likers in "chronological" sequence. _(right?)_
why not keep `getChronoMediaLikers` (using "Chrono").

just an idea.

ranjbarhadi: ahhh yes i did miss that...

azureatlas:
@ranjbarhadi where/when is this endpoint called by the official app (as opposed to the regular getMediaLikers() one)?

SteveJobzniak:
Hi @ranjbarhadi. Thanks for the contribution!

I renamed it from `getChronoMediaLikers` to `getMediaLikersChrono`. Because we try as much as possible to follow the official endpoint, which you can see is: `"media/.../likers_chrono/"` meaning the words in this case are meant to be in this order: "media -> likers -> chrono".

It also helps with understanding. Because "getChronoMedia" sounds like some kinda weird media item type, whereas "getMediaLikersChrono" is clear that it's about likers in "Chrono" fashion. I also tweaked the comment a bit to reflect that.

So now it's ready to merge, if it passes the next check. ;-) And those remaining questions are:

1. How did you find this endpoint?

2. Is it used by the Android Instagram client or only by the iOS client? To avoid easily getting banned by using iOS APIs from Android etc, we don't mix the two APIs (because they are different).

SteveJobzniak:
Pinging @ping since I see that you've got this chrono-endpoint in your Python library. Perhaps you can chime in here. :-)

ping:
On the mobile right now but if I remember correctly, the endpoint is in the apk but not actively used in the app afaik. At least not in a way that I can trigger. @mgp25 will need to confirm this one.

This reminds me that I need to tag this endpoint as experimental in my lib. 😅

SteveJobzniak:
@ping Thanks a lot for that great information! :) We'll see what @mgp25 discovers. Hmm, if it's true that it's not used officially then it's a super dangerous API.

@ranjbarhadi How did you find this API?

ranjbarhadi:
hi @SteveJobzniak . thanks for the help. this is my first contribution in github community and I'm trying to do my best.
i found this Endpoint by decompiling the latest Instagram APK. it was something like this:

```java
public static void m21723c(C7661t c7661t) {
        String str;
        C7661t.m21722b(c7661t);
        c7661t.f36329b.f13239d = false;
        if (c7661t.f36332e) {
            str = "media/%s/likers_chrono/";
        } else {
            str = "media/%s/likers/";
        }
        C2171e a = C7058i.m20087a(c7661t.f36334g, C3489l.m10785a(str, c7661t.f36331d), null, c7661t.f36329b.f13241f, null, false, false);
        a.f16485b = c7661t.f36335h;
        c7661t.schedule(a);
    }
```
i tried it and it worked.

ranjbarhadi:
@azureatlas i found this in the original APK. i don't have any information when Instagram prefers to use this endpoint. if you look at the code i provide for @SteveJobzniak in the above, Instagram is checking for something and if it returns true it uses this endpoint instead. what it is ? i still don't know. but I'm looking :)

mgp25:
It is an option (is time ordered? Yes/no). I don't know wether this is an unused endpoint or instagram is planning to add it soon. Some time ago, Instagram stopped showing feeds and likes in chronological order and started to do it randomly.

I don't know if they changed the algorithm so the endpoint we use displays media likers in chronological order, tests may be required. So far, i think its an user option

aitte2:
Safest to avoid it until we know if/when the app calls it, otherwise we are literally the only people in the world calling that endpoint, and is easy to ban us.

ranjbarhadi:
i' trying to add logs to smali code of Instagram's android client to pin point where this endpoint has been called. it a big messy code and i haven't yet found the appropriate place. any suggestions?

azureatlas:
@ranjbarhadi The "if" result might be based on an experiment that could be active or not.

SteveJobzniak:
@mgp25 My guess right now is that @azureatlas is right. The "if" is probably an experiment flag.

Good idea anyway to add a warning. I added an even more serious warning just to protect people, which means we can now merge it. People can make the decision if they have the courage to use this endpoint, but possibly losing their account. And I placed a "TODO" to research this function more in the future or to remove it if it wasn't safe.

So now it can sit there and people can make their own decision if they have the courage to use it. ;-)

@ranjbarhadi Thanks for the contribution.

Torq07:
@ranjbarhadi at comment from Apr 21 you telling that you found the option to paginate like Endpoint, can you share this info? I'm trying to paginate it but each time my attempt are failing. Do you add offset and limit params or just page number to paginate. 
I would be very appreciate to you if you share your findings with me

PR #1252, opened 2017-05-16
Add page name to user object.

matthijsotterloo:


SteveJobzniak:
Thanks! What's an example value for this? I guess this is for the new "business" feature which lets people name their business profile "page"?

matthijsotterloo:
@SteveJobzniak Yes correct. If you page is linked to a page called "Demo page" the field will return "Demo page"

PR #1259, opened 2017-05-19
Direct Messaging update

valga:
Direct methods have become a bit outdated, so this PR tries to fix that.

List of changes:
- lightweight (i.e. without file uploads) direct methods do not use multipart requests anymore;
- direct messaging methods now accept thread identifier as recipient via `['thread' => 123]`;
- `directLike()`, `directHashtag()`, `directLocation()` and `directProfile()` methods;
- `updateThreadTitle()` method;
- `deleteThreadItem()` method;
- `markDirectItemAsSeen()` method;
- `muteDirectThread()` and `unmuteDirectThread()` methods;
- `addUsersToThread()`, `leaveDirectThread()` and `hideDirectThread()` methods;
- `approvePendingThreads()`, `declinePendingThreads()` and `declineAllPendingThreads()` methods.

List of compatibility breaks:
- `directPhoto()` does not accept `$text` parameter anymore. Just because there is no way to do that in official application;
- `directThreadAction()` method is replaced with a set of approve/decline methods.


SteveJobzniak:
@valga Thank you _so much_ for taking on this direct sharing rewrite. I was going to get to it later, but I've been really slowed down by my health treatment for my severe burnout.

I love the backwards-compatible solution, so that we won't have to go to 3.x of the library. Using an associative array to signal that `['thread'=>'129319239']` to target threads is a _brilliant_ solution! I love it!

I'll be doing a full review of the RTC code this weekend. And when this one is done I'll help review it too. One thing already stood out to me at first glance: Various new functions such as directLike() aren't documenting the full `$recipients` possibilities, even though they all rely on _sendDirectItem internally (which can additionally accept a backwards-compatible string or array of user strings).

There are two ways to document functions that re-use a parameter from another function. Either just copy-paste the original text. Or, if that's a maintenance nightmare (in this case it's a lot of functions), it'd be fine to write something like `@param $recipients See _sendDirectItem() for possible values.`

ironbeasty:
@valga @SteveJobzniak I think I'm missing something, but to "approve" threads or direct messages, you need the thread_id, correct?

I added your pushes and I still can't find where to get the thread_id of pending messages.
I did var_dump($ig->getV2Inbox();) and CTRL+F and find no thread_id.

Help?

valga:
> Various new functions such as `directLike()` aren't documenting the full `$recipients` possibilities.

I made it on purpose: old recipients format is deprecated. While we still have backwards compatibility for it, it should be removed in future (major) releases.

So, docs for new methods mention only new, preferred, format.

SteveJobzniak:
@valga Ah I see. I agree with that! 👍

SteveJobzniak:
@valga I've just noticed that `directShare` has also changed behavior of the first parameter so I edited the top post to reflect that.

Direct sharing was actually the only thing I did not have time to do a complete rewrite of for 2.x. I knew they were only half-finished, but I didn't think that we'd need to change their parameters.

This makes me think... let's do it right. Several of these direct-functions are being changed enough that they break compatibility. So we're no longer going to be restricted by backwards compatibility.

This will therefore be part of 3.x, which will have all commits since 2.0 was released: Tons of new features, realtime client, improved responses, API v2 support, and rewritten direct sharing. And the only thing people have to note during that upgrade is simple: "The direct-functions have been rewritten". That's easy enough for people to do the upgrade easily.

So now that means your hands are completely untied. :-) Please do what you wished you could do but were stuck on when trying to be backwards-compatible before. That means... delete useless functions (like `directShare`? Since there is `directPhoto` already). Get rid of backwards-compatible `"user"` and `["user1", "user2"]` $recipients format and feel free to rename anything anywhere. There are no ties anymore. :-)

SteveJobzniak:
Btw it looks like mgp may have done an incorrect merge of "master" into this pull above. Just fyi. ^_^ I would revert that if I were you. He added some "if sharetype == share" block... which does not look ok.

valga:
Let's stop at this point. It is ready to merge.

SteveJobzniak:
@valga Okay thanks! I'll start reviewing now. There are a few problems. I'll cover them in review comments.

SteveJobzniak:
Okay I've finished the code review. There are basically some small documentation problems and some pretty big naming consistency problems that need to be discussed and fixed before it's 3.0-ready. If you're tired, don't worry about looking at them now, it will probably be at earliest a month or two before 3.0 is ready.

SteveJobzniak:
Update regarding the naming discussion:

https://help.instagram.com/400205900081854

That settles it. Instagram themselves calls it "Instagram Direct". So let's go for the "Direct"-only naming which I mentioned in the review above.

CashCode:
Good job! Thanks for update.
How about new messages check?
I mean method to see what threads has new messages, like v2inbox, but with field of new messages (1-2-3 and so on).

valga:
Actually I can't see a thing. May be I need to request a review from you again.

SteveJobzniak:
To summarize everything above, there are two kinds of problems; some slight documentation issues, but the main problem is the data naming inconsistencies. Those can be solved by following Instagram's own name for the feature: "Instagram Direct". So any class/function references to "DirectThread" needs to be just "Direct" instead. Putting all related funcs/classes all in the "Direct" terminology will solve it nicely and consistently and prevent any future problems.

valga:
Ok, I will review your review later.

SteveJobzniak:
@valga Thanks, there is no hurry. Take as many weeks as you want. :) We have a lot of time until 3.0 release.

src/Client.php
```diff
@@ -439,9 +439,14 @@ public function getMappedResponseObject(
         // means that the user can look at the full response details via the
         // exception itself.
         if (!$responseObject->isOk()) {
+            if ($responseObject instanceof \InstagramAPI\Response\SendItemResponse) {
+                $message = $responseObject->payload->message;
+            } else {
+                $message = $responseObject->getMessage();
```
valga: Sorry, but I can't use accessor methods without them being documented, because I simply don't have autocomplete/jump to declaration in that case.

And I can't use `$responseObject->message` either, because `getMessage()` method is not a simple getter.

valga:
I believe I fixed everything except for naming issues. I will provide a proposal for them later.

src/Client.php
SteveJobzniak: @valga Good point, I had forgotten that I rewrote getMessage to support object-based messages in some of Instagram's error situations, so we do need `getMessage()` at all times instead of just `->message`.

As for the reason to use object methods instead of the auto-mapped accessors.... ehm, we use the auto-mapped accessors everywhere. That's the code style we have everywhere else, that's why I reacted. Hmm... At least in this situation, in a file full of accessor methods, we should use the same way.

Using the direct property method in Realtime was fine since that's a brand new namespace. But mixing and mismatching isn't.

SteveJobzniak: In this case I'll move it to accessor method, since this code is done. As for moving the whole project to direct property access instead (for autocomplete purposes), that would be too risky because we sometimes write custom accessors (like you saw with getMessage). So for consistency, I'm fixing this.

valga:
Here is my proposal for method names:
```
getPendingInbox()          => getDirectPendingInbox()
getVisualInbox()           => getDirectVisualInbox()
getDirectShare()           => getDirectShareInbox()
getV2Inbox()               => getDirectInbox()

getRankedRecipients()      => getRankedDirectRecipients()
getRecentRecipients()      => getRecentDirectRecipients()

approvePendingThreads()    => approvePendingDirectThreads()
declineAllPendingThreads() => declineAllPendingDirectThreads()
declinePendingThreads()    => declinePendingDirectThreads()

directThread()             => getDirectThread()
updateThreadTitle()        => updateDirectThreadTitle()
muteDirectThread()         => muteDirectThread()
unmuteDirectThread()       => unmuteDirectThread()
addUsersToThread()         => addUsersToDirectThread()
leaveDirectThread()        => leaveDirectThread()
hideDirectThread()         => hideDirectThread()

directShare()              => sendMediaToDirect()
directMessage()            => sendTextToDirect()
directPhoto()              => sendPictureToDirect()
directLike()               => sendLikeToDirect()
directHashtag()            => sendHashtagToDirect()
directLocation()           => sendLocationToDirect()
directProfile()            => sendProfileToDirect()

deleteThreadItem()         => deleteDirectItem()
markDirectItemAsSeen()     => markDirectItemSeen()
```

src/Client.php
```diff
@@ -466,7 +466,8 @@ public function getMappedResponseObject(
             // API developer debugging? Throws error if class lacks properties.
             $mapper->bExceptionOnUndefinedProperty = true;
         }
-        /** @var Response $responseObject */
+        
+        /** @var Response|mixed $responseObject */
```
valga: Actually it can't be mixed, because you call some specific methods. So for being perfect here, we must introduce an interface with `setFullResponse()`, `isOk()`, `getMessage()` methods. Or we can totally ignore it.

src/Client.php
valga: Ok, but having a PHPDoc with `@method` for all responses would be great. But it is very time consuming, so I guess we will never have it.

SteveJobzniak:
@valga I just finished the re-review and did some slight grammar fixes etc. It all looks good.

So yes let's tackle the naming. I **love** the proposal a _lot_! With just 1 or 2 very tiny tweaks.


```
getPendingInbox()          => getDirectPendingInbox()
getVisualInbox()           => getDirectVisualInbox()
getDirectShare()           => getDirectShareInbox()
getV2Inbox()               => getDirectInbox()

getRankedRecipients()      => getRankedDirectRecipients()
getRecentRecipients()      => getRecentDirectRecipients()

approvePendingThreads()    => approvePendingDirectThreads()
declineAllPendingThreads() => declineAllPendingDirectThreads()
declinePendingThreads()    => declinePendingDirectThreads()

directThread()             => getDirectThread()
updateThreadTitle()        => updateDirectThreadTitle()
muteDirectThread()         => muteDirectThread()
unmuteDirectThread()       => unmuteDirectThread()
addUsersToThread()         => addUsersToDirectThread()
leaveDirectThread()        => leaveDirectThread()
hideDirectThread()         => hideDirectThread()

directShare()              => sendMediaToDirect()
directMessage()            => sendTextToDirect() --> The endpoint is called message, but I think this is fine anyway since this "messaging" function is just for text, and this naming matches $text params everywhere, so yeah! Go for it
directPhoto()              => sendPictureToDirect() --> sendPhotoToDirect(), to match endpoint & because nobody says "picture" these days
directLike()               => sendLikeToDirect()
directHashtag()            => sendHashtagToDirect()
directLocation()           => sendLocationToDirect()
directProfile()            => sendProfileToDirect()

deleteThreadItem()         => deleteDirectItem()
markDirectItemAsSeen()     => markDirectItemSeen()
```

Along with this awesome cleanup, the direct-related server responses will also have to be renamed to match (both old and new replies, since we have some older ones with the old naming; just look for `*direct*` and `*thread*` in Response and Response\Model).

A good naming format that we use most of the time is to name them after the function, so for example if "deleteDirectItem()" needs a response, name it "DeleteDirectItemResponse".

After this, the pull request is ready for sure! ❤️

src/Client.php
SteveJobzniak: It can't be mixed in our own code, because we always call `getResponse()` with a $baseClass value that is derived from Response.

But it can be mixed in other people's code, because `$baseClass` allows any class type. This is why we have the `Instagram->request()` function as public. So that people can make any request they want, and then do `getResponse()` with their own custom object.

valga:
Which one is better:
```
getPendingInbox()  => getPendingDirectInbox()
getVisualInbox()   => getVisualDirectInbox()
```
or
```
getPendingInbox()  => getDirectPendingInbox()
getVisualInbox()   => getDirectVisualInbox()
```

SteveJobzniak:
@valga The second format, because the more functions we have as "getDirectXX" the more consistently memorable it is.

Also the endpoint uses that format: `'direct_v2/visual_inbox'` -> `getDirectVisualInbox()` :)

SteveJobzniak:
It also matches all the other getDirect...Inbox functions (4 in total). Although... I think we need to think about this more...

We have two choices:

1. Either stick to `getDirectXX` format for memorability (people won't have to wonder if it's getDirectVisualInbox or getVisualDirectInbox if it's always getDirectXX format). In that case we should even consider stuff like `getDirectRankedRecipients()`.

2. Rename everything to "english sentence flow". Normally we do not do that. Hmm... thoughts?

SteveJobzniak:
I'll make tables of both methods, and we can see...

SteveJobzniak:
Sentence flow (less consistent, more natural language):

```
getPendingInbox()          => getPendingDirectInbox()
getVisualInbox()           => getVisualDirectInbox()
getDirectShare()           => getDirectShareInbox()
getV2Inbox()               => getDirectInbox()

getRankedRecipients()      => getRankedDirectRecipients()
getRecentRecipients()      => getRecentDirectRecipients()
```

Endpoint flow (more consistency, what we normally try to do but not always if it's too awkward):

```
getPendingInbox()          => getDirectPendingInbox()
getVisualInbox()           => getDirectVisualInbox()
getDirectShare()           => getDirectShareInbox()
getV2Inbox()               => getDirectInbox()

getRankedRecipients()      => getDirectRankedRecipients()
getRecentRecipients()      => getDirectRecentRecipients()
```

Hmm... I think my vote goes like this: Let's keep _slightly_ mixed low, because endpoint flow is too robotic. But make an exception to the inbox functions and use getDirectXInbox on all of those for consistency among those 4 functions.

Thoughts?

valga:
It is not an easy problem indeed. Since most users have (and use) an autocomplete feature, let's see from its point of view. Natural naming is faster to type, for example it is enough to type `getPen` to narrow a list down to 2 methods. But grouped naming gives an access to full set of methods, although this list is going to contain a lot of methods.

SteveJobzniak:
@valga That's true, but autocomplete is just `getDi<tab> = getDirect` and then `P<tab> = getDirectPending`.

I just had a shower and thought about this and maybe have a solution... it depends on if we can make the return value make sense (if all can have the same `getResponse()` object type).

```php
/**
* @param string $inboxType One of: "inbox", "pending", "visual", "share".
*/
public function getDirectInbox(
    $inboxType = 'inbox')
{
}
```

SteveJobzniak:
Same for:

```php
/**
* @param string $recipientType One of: "recent", "ranked".
*/
public function getDirectRecipients(
    $recipientType = 'recent')
{
}
```

So we get in total:

```
getPendingInbox()          => getDirectInbox($inboxType)
getVisualInbox()           => ...
getDirectShare()           => ...
getV2Inbox()               => ...

getRankedRecipients()      => getDirectRecipients($recipientType)
getRecentRecipients()      => ...

approvePendingThreads()    => approvePendingDirectThreads()
declineAllPendingThreads() => declineAllPendingDirectThreads()
declinePendingThreads()    => declinePendingDirectThreads()

directThread()             => getDirectThread()
updateThreadTitle()        => updateDirectThreadTitle()
muteDirectThread()         => muteDirectThread()
unmuteDirectThread()       => unmuteDirectThread()
addUsersToThread()         => addUsersToDirectThread()
leaveDirectThread()        => leaveDirectThread()
hideDirectThread()         => hideDirectThread()

directShare()              => sendMediaToDirect()
directMessage()            => sendTextToDirect()
directPhoto()              => sendPhotoToDirect() **
directLike()               => sendLikeToDirect()
directHashtag()            => sendHashtagToDirect()
directLocation()           => sendLocationToDirect()
directProfile()            => sendProfileToDirect()

deleteThreadItem()         => deleteDirectItem()
markDirectItemAsSeen()     => markDirectItemSeen()
```

valga:
> The main problem is that we should group related functions.
It is easily solvable with child classes and something like `$instagram->direct->getPendingInbox()`

And using one method for all inboxes (recipients) is not ok, because they can require different set of request params. And their response objects are not the same.

SteveJobzniak:
@valga Since we have to go 3.0 anyway, I actually really like the child classes idea, grouping all related functions by their "area".

Because this library has grown from small beginnings to now having over a hundred functions, and the auto-generated docs are _insanely_ messy, and we constantly have to scroll through the huge Instagram.php to edit it.

And people don't know what functions are related to each other, so they have to search for keywords like "story" and hope to find functions with that in the func-name, etc...

With child classes, something like this would be possible:

```
src/Request/Direct.php:

namespace Instagram\Request;

class Direct
{
    public function __construct($parent) { $this->_ig = $parent; }
    public function getPendingInbox() { return $this->_ig->request()->getResponse(); }
}

src/Instagram.php:

class Instagram
{
    public $direct;
    public function __construct($parent) { $this->direct = new Request\Direct($this); }
}

Then just:

$ig->direct->getPendingInbox();
$ig->direct->sendPhoto();
$ig->timeline->uploadPhoto();
$ig->timeline->getUserFeed();
$ig->story->uploadVideo();
```

It would really solve all of our documentation problems and all of our expansion/name clashing problems...

But I'm very sick and still doing health treatment so I don't have the energy for it, at least right now.

But we can START the process via this direct rewrite, and get the ball rolling!

```
getPendingInbox()          => direct->getPendingInbox()
getVisualInbox()           => direct->getVisualInbox()
getDirectShare()           => direct->getShareInbox()
getV2Inbox()               => direct->getInbox()

getRankedRecipients()      => direct->getRankedRecipients()
getRecentRecipients()      => direct->getRecentRecipients()

approvePendingThreads()    => direct->approvePendingThreads()
declineAllPendingThreads() => direct->declineAllPendingThreads()
declinePendingThreads()    => direct->declinePendingThreads()

directThread()             => direct->getThread()
updateThreadTitle()        => direct->updateThreadTitle()
muteDirectThread()         => direct->muteThread()
unmuteDirectThread()       => direct->unmuteThread()
addUsersToThread()         => direct->addUsersToThread()
leaveDirectThread()        => direct->leaveThread()
hideDirectThread()         => direct->hideThread()

directShare()              => direct->sendMedia()
directMessage()            => direct->sendText()
directPhoto()              => direct->sendPhoto() **
directLike()               => direct->sendLike()
directHashtag()            => direct->sendHashtag()
directLocation()           => direct->sendLocation()
directProfile()            => direct->sendProfile()

deleteThreadItem()         => direct->deleteItem()
markDirectItemAsSeen()     => direct->markItemSeen()
```

And for their responses... hmmm... that folder is a mess too due to so many responses. So even if we only get started with Direct, at least let's do that...

```
src/Response/Direct/SendItemResponse.php
src/Response/Direct/MuteThreadResponse.php
src/Response/Direct/PendingInboxResponse.php
src/Response/Direct/ThreadResponse.php
src/Response/Direct/UpdateThreadTitleResponse.php
```

For per-function responses it's basically the name of the function (minus useless parts like "get"). Some responses are used in many functions, so obviously generic names are fine like `SendItemResponse`.

All they would need at the top is `use Instagram\Response\Model` so we can still refer to "`@var Model\Whatever`".

What do you think?

SteveJobzniak:
This would really liberate our function naming. And make the API easier to understand (all related funcs in 1 file, for each area).

And people could just `$d = $ig->direct;` and then `$d->sendPhoto()` etc. Lots of nice benefits.

SteveJobzniak:
I'm going out now to treatment, so I'll be away. If the above idea sounds in any way ok to you, then proceed with it. :) And since all functions will be in a "direct" class you don't have to worry about asking about approved, consistent function names anymore.

This would really be a big improvement.

SteveJobzniak:
@valga I came back because I can't focus outdoors while something this big is unfinished. I need to merge this now so that we get the repo into a unified state so I can start ripping out functions and putting them in sub-classes.

valga:
> All they would need at the top is use Instagram\Response\Model so we can still refer to `@var Model\Whatever`.

As far as I remember, it doesn't work with JsonMapper. You need either fully qualified class name or class name which is related to current namespace.

> `src/Response/Direct/PendingInboxResponse.php`

You use Response 2 times. Perhaps we can throw last one.

SteveJobzniak:
@valga If you still have skype, please get online and add me (username: armindale). I've merged this now to start doing the work fast and furious this evening and we'll talk better in chat.

matthijsotterloo:
@valga What would be the way for deleting images in a thread? It always gives a `Wrong permissions` error.

PR #1260, opened 2017-05-19
Use GMP functions for shortcode conversions

valga:
Benefits:
- `toCode()` is about 8x faster now;
- `fromCode()` is about 4x faster now;
- no lookup tables.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,6 +37,7 @@
         "ext-curl": "*",
         "ext-mbstring": "*",
         "ext-gd": "*",
+        "ext-gmp": "*",
         "ext-exif": "*"
     }
 }
--- a/src/InstagramID.php
+++ b/src/InstagramID.php
@@ -47,6 +47,7 @@ class InstagramID
         do {
             $encoded = self::BASE64URL_CHARMAP[(int) ($id & 63)].$encoded;
         } while (($id >>= 6) > 0);
+
         return $encoded;
     }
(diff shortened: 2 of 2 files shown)


SteveJobzniak:
I would benchmark GMP, but there's a reason I didn't write this with GMP:

`Call to undefined function gmp_init()`

It's a very obscure PHP extension.

My 6 year old dual-core laptop can do:

Fifteen thousand (15,000) `toCode()` per second. Thirty-two thousand (32,000) `fromCode()` per second.

Most projects would probably convert a few hundred codes at most.

So I don't know if I agree with this change. Certainly not merging it without a discussion.

@mgp25 @valga Thoughts?

SteveJobzniak:
Okay I checked my hosted server (which for me, like for most people, is outside my control). It has no GMP library. I'll have to revert this.

SteveJobzniak:
@valga You are welcome to rewrite this proposal so that it uses a the universal code by default, but efficiently jumps to a different codepath with GMP *if* the system has it. Which I'd estimate is less than 1% of our library users, unfortunately.

In that case, I'd probably do a single check for the gmp_init function and cache that as a true/false somewhere. Then toCode/fromCode could have something like this at the top:

```php
if (self::$hasGMP) {
    // do GMP code and return instantly
}

// current code here... which runs if GMP didn't exist.
```

PR #1261, opened 2017-05-19
Fixes for multipart requests

valga:
This one is for @mgp25.

It seems, device id is no longer used as a multipart boundary. 2265154 replaces it with random string.

Also, it seems there was a bug that forced all uploaded filenames to be `pending_media_*.jpg`. 2f1884a fixes it.


mgp25:
@valga checking it in a while. Nice work! 👍

SteveJobzniak:
@valga Perfect. Thanks for bringing boundaries and upload filenames up-to-date!

Issue #1231, opened 2017-05-08
Subscriptions


mgp25:
For years I’ve given free support but there are so many questions that it takes up hours of my day every day. So I've decided to begin offering 2 types of subscriptions:

**Free user:**
- Bug reports
- New feature ideas
- Other stuff must be helped by each other/wiki; we'll close question-tickets with generic replies about wiki links, source code reading, and linking to the subscriber feature if you want to have personal help.

**Subscriber (20 eur/m):**
- Personal help with any API question

**Premium subscriber (100 eur/m):**
- Personal help with any API question
- Deep integration
- Latest Instagram keys/signature, which means better compatibility
- Granted the [PREMIUM license](https://github.com/mgp25/Instagram-API/blob/master/LICENSE_PREMIUM) for the duration of the subscription

**Note 1:** API will continue to be updated regardless of subscriptions. Keys will be updated only when they stop working.
**Note 2:** All information can be found in the source code, examples, wiki, FAQ and in already answered issues. Issues will be closed directly pointing to the issue tracker, source code and wiki.
**Note 3:** Collaborators, and contributors with substantial high-quality commits, are considered to be premium subscribers.

**Note 4:  API will continue to be updated regardless of subscriptions and will be free as always**. ALL info can be found in the repository. 

Read all the intro and notes, please.

Regards

**Contact or questions:** me [ @ ] mgp25.com

jpmoney5065:
Donations are cool.

But publicly setting up  a for-profit service is a bad idea. 

You are breaking the law, violating Facebook EULA, and using their IP. Also technically violating github ToS.

Be careful, this could get the whole repo shut down. And watch out for lawsuits!

Good luck!

jonatanfroes:
You're right. A lot of people make money with this API. You need to make too. I'll join the Subscriber plan.

Go ahead!

SteveJobzniak:
Hey everyone!

Mgp and I have spoken spoke about this and I fully support him in this decision. His reverse engineering of the API is the _whole reason_ that all of your fun, personal Instagram projects can be made! In fact, all other Instagram API projects for all other languages come from people copying this project's work! (For example, if there's been a small mistake here, they've all copied that to "their" projects, proving that the others were just copying this repo. 😄 )

[...]

PR #1273, opened 2017-05-24
Refactor requests

valga:
While doing direct messaging update and stuff, I faced with some annoying things:

- [x] For now, we must maintain request parameters order manually, and this is very annoying and error-prone. It can be done automatically using Java's `hashCode()` analogue.
- [x] CSRF token handling is very strange. While it is just a cookie value, we store it in separate place and update it just on `login()` call with some voodoo magic.
- [ ]  Speaking of automation, the `_csrftoken` is required for every `POST` request, `_uid` is required for every signed request, and `_uuid` is required for every `POST` request made by logged in user. For now, we add them manually.
- [x]  We have 4 methods that do step-by-step multipart requests creation. Why can't we have just one function for doing multipart requests?

Work in progress, but this may take a while...


SteveJobzniak:
Wow, sweet. Thanks for this! The request/signature code was only minimally touched for 2.0, and I know that it was pretty dirty.

I'll comment on the thoughts and questions.

1. About the parameter order: I've never noticed that the order matters. Does it really? Instagram receives our "GET" query parameters identically no matter what order we send them in, because the server hides that info from their scripts, so order should not matter for GET queries. And all POST-data is sent as encoded JSON body, so it shouldn't matter either since they just json_decode() that blob on their server and get an object where they can't see what order we sent the parameters in.
2. Yes, CSRF token handling is very, very strange. It was that way in v1, and I ported it that way to the v2 rewrite. I never liked that it wastes energy finding the token on every API request, but only actually saves it to a variable on a login() call. And the way the api request function returns the token as an array with [0] and [1] is very ugly.
3. Automating token adding for POST requests, and automating _uid adding for signed request, may be a good idea... But we must consider: What if an API _doesn't_ need those values? Adding it manually has been fine so far. Do you think we'll be safe automating it?
4. What part of the code you mean?

valga:
1. I don't care about Instagram. Exact parameters order is great for comparing our requests against requests made by application. And speaking about performance, it can calculate about 1M hashes and reorder about 100K arrays per second, so it is like nothing.
2. Actually I am going to scan cookies every time we _need_ CSRF token (like for `POST` request).
3. I will skip this point in case of at least one exception from mentioned rules. But it worth saying, that application does follow these rules.
4. `src/Client.php` has `uploadPhotoData()`, `requestVideoUploadURL()`, `changeProfilePicture()` and `directShare()` methods. We can add `addFile()` method to `Request` class which converts it to multipart one. And I don't know why `requestVideoUploadURL()` is multipart, because it doesn't upload any files.

SteveJobzniak:
1. Oh okay, faster than I assumed. And I can see the great valud of having the feature for easy comparison of JSON dumps, so I like the idea! :)
2. Oh you mean instead of storing token in settings + class instance, you will make it scan our cookie jar and extract it just for those requests? That sounds smart, I like it. Perhaps just add a function to Client.src called getToken() which scans the jar and returns it. And then call that as addParam("token", $this->client->getToken()).
3. For this, perhaps we CAN automate it (since most requests are signed and need token and it is annoying to repeat the code for sure). The Request class could then do $this->parent->client->getToken(). But just add functions to Request for disabling the auto-adding if we ever have to do it.
4. Ah you mean that the functions that upload files repeat building of multipart headers. Ehh, yeah, that was done so we have full control of how each endpoint is called. They differ in their multipart form data. And also in handling of what is done with the uploaded data afterwards. But if you want, make an addFile function in Request which builds the multipart form array (meaning boundary, upload id, and a list of form params to send). That would be a nice cleanup. Just be aware that lots of code calls functions like "uploadPhotoData". And also, those low-level file transfers should not be exposed to Instagram.php's public API. Keep those teo things in mind :)

SteveJobzniak:
@valga I just did some rewrites in Settings + Client related to cookie storing. We now only write cookies to database/disk once per session (just like with FileCookieJar): When the script exits, or when you `setUser()` to a _different_ user. So now it's much more efficient (instead of writing cookies to StorageHandler on EVERY request).

This also reminded me how the Client cookie loading works, which is good info for you:

When the library user does `setUser()`, we set the StorageHandler's active user to that user, and we tell the Client to `updateFromCurrentSettings()` which it uses to get various storage backend things such as the cookies and user agent of the current user.

What this means is that AFTER the `setUser()` call, we can be 100% sure that a CookieJar exists and has loaded all cookies that the user had (or created an empty jar if user had none).

Before that call there is of course no cookie jar.

Actually... I'm going to add a `getCookieValue($name)` and `getToken()` function to the Client now.

SteveJobzniak:
@valga Okay, so after adding `getCookie()` and `getToken()` above, I went ahead and refactored all code... Now we're finally liberated from the old v1.x way of getting the token! :-)

https://github.com/mgp25/Instagram-API/commit/c321297a3d891f6769c0fa5391f65a9b80a617e5

SteveJobzniak:
@valga Just remembered you said something about 'sessionid' in Mqtt... found it! ;)

https://github.com/mgp25/Instagram-API/commit/5d031d84e0a4a9b20ec3ef38750cca8757cdfbfd

valga:
@SteveJobzniak, I dislike an idea of using destructor for saving an important data. Destructor is great for closing handles or cleaning up temporary data, but it might never be called, so we can't rely on it for saving such precious data as cookies and settings.

So, we should ensure that settings and cookies are saved at least in `login()` and `logout()` methods.

SteveJobzniak:
@valga First off, Destructor always gets called. Only killing the PHP process via SIGTERM (instant exit) would prevent destructors from getting called.

Second off, saving data in destructors is totally valid, for example see Guzzle's FileCookieJar: https://github.com/guzzle/guzzle/blob/master/src/Cookie/FileCookieJar.php#L39

Third, **we** use FileCookieJar in Client when a storage backend replies `"file:<path>"` to its "get cookies" question. So this performance improvement finally brings all storage backends to identical behavior: Saving cookies on script exit (destruction) and on `setUser()`.

Settings are not handled in destructor. Settings are always written immediately when their value changes, because those change rarely and *their* values are very precious. Cookies are not precious. They can be re-created instantly by logging in again. But to do that we need settings (UUID, Phone ID,  Device ID) so that we can re-create the exact same cookies and won't look like a different device.

So settings are saved live. The cookie jar is saved at script exit and at `setUser()` (when switching to a different user).

But yes, it would be nice if we always save the jar on `login()` and `logout()` methods. Those are the truly important ones and it would be nice to write cookies instantly after that so that other processes can use them.

However, as the code is right now, we can't just add "save cookie jar" code to login. It would not work for Guzzle's FileCookieJar (which only saves cookies on destruct).

Here's what would need to be done:

Step 1: Move src/Settings/Storage/File.php's atomicWrite to Utils instead.
Step 2: Rewrite src/Client.php so that it no longer uses FileCookieJar. Use regular CookieJar even for file-based cookie backends, just like with the other backends.
Step 3: Rewrite src/Client.php's cookie loading feature to make it read the "cookie string" from disk, and rewrite its cookie saving feature to make it detect when the desired cookie storage type is "file" (we already have a variable in there with that type info in it; we'll need another variable, to keep the filename we used) and make it get the cookie string and write it to disk via atomicWrite (ensuring no partial overwrites by multiple processes).
Step 4: After that, and only after that, we will be able to add cookie saving to more places than just at script exit/switch user.

Hmm, but I like the benefit of being able to save cookies at login/logout. I'll take care of these steps now.

valga:
> Destructor always gets called. Only killing the PHP process via SIGTERM (instant exit) would prevent destructors from getting called.

Also destructor won't be called in case of fatal errors, bad coding, power failures, kernel panics, etc. So you can rely on it only if you want to close or cleanup something.

But I guess Guzzle's authors simply don't care about cookies. I will not discuss this issue anymore.

SteveJobzniak:
@valga Yes, that's why I agree about also force-saving cookies on login and logout. :-) I'm about 50% done rewriting so we no longer use any Guzzle FileCookieJar. After that we will be able to save anytime.

SteveJobzniak:
@valga Alright, now we're free! No more bad Guzzle FileCookieJar (which wasn't concurrency-capable, and could corrupt jars due to being non-atomic, and only wrote at end of script). The cookie jar is is now auto-saved at good intervals and after important API calls, and at script exit/setUser.

https://github.com/mgp25/Instagram-API/commit/07608c841fc443b158ebd01c7f27003fcad446aa

https://github.com/mgp25/Instagram-API/commit/bd4b26ac4d9f37bbeed16e658943e5a2391a1c99

https://github.com/mgp25/Instagram-API/commit/e291a7cf147be2c53c45135e300b605b1d5b4c18

valga:
> EDIT: Nevermind, yeah let's skip this one. Only 22 of our 100+ requests are "setSignedPost" so it's okay that we have to manually add the _csrftoken and _uid parameters there. There's no big benefit to automating it, and there's a risk we have more trouble later if we automated it.

Some stats. We have about 140 requests:
- about 80 of them are POST, so we need _csrftoken for them;
- about 75 of them are POST with logged in user, so _uuid is required for them;
- about 50 of them are signed with logged in user, so _uid is required.

SteveJobzniak:
@valga Hi! Since stepping back to focus on my health, I haven't read through the changes in detail but what I have read through looks great (the multipart builder). <3

Thanks so much for cleaning up that messy form building code and giving us such a beautiful method instead!

As for the automation, I wonder what you think about this idea: `autoAdd(['_csrftoken', '_uuid'])`. At least then we retain control over explicitly telling it to add those fields (so that our code is explicit and has no hidden behaviors), but we no longer need to repeat the long "`->addPost('_csrftoken', $this->ig->client->getToken())`" and "`->addPost('_uuid', $this->ig->uuid)`" and "`->addPost('_uid', $this->ig->account_id)`".

It's just a request for your comments and what you think about it. Any alternative ideas?

valga:
The main point of being fully automated is absence of any manual actions and following strict rules set. There are 3 simple rules I mentioned, so any semi-automated actions do not follow them, because one can forgot to add some param to a request. So we either make everything fully automated, or we skip this point completely. And I am being lazy these days, so I vote for skipping.

By the way, while testing refactored `requestVideoUploadURL()` method I found that:
1. Instagram uses new awesome video upload method (from Facebook I guess): it tries to upload whole file at once. And if something fails, it continues from last uploaded offset. So, no more chunks and no more full chunk retries.
2. Old uploads are still being used for direct videos (and probably something else), and "Video upload must be done in exactly 4 chunks" is not true. But this is not important I guess.

SteveJobzniak:
@valga Thank you so much for this awesome work!

And yes I am okay with skipping automated params, I never felt it was a big issue to just insert them manually. :)

Alright, as one of the last things I do before I put full focus on my health, I will do a full review of this new code later today when I get home from the health clinic. :)

SteveJobzniak:
@valga I got home but am too exhausted to think, so I will do the review when I wake up. It'll be merged tomorrow for sure!

SteveJobzniak:
@valga I am nearly done with the review. Just confused about one part:

https://github.com/mgp25/Instagram-API/pull/1273/files#diff-71322b46fb7df82529cd75bcd0556256R430

```php
+        $result = null;
 +        // Send the uploaded video to recipients.
 +        for ($attempt = 1; $attempt <= $maxAttempts; ++$attempt) {
 +            try {
 +                // Attempt to configure video parameters (which sends it to the thread).
 +                $result = $this->_sendDirectItem(
 +                    'video',
 +                    $recipients,
 +                    [
 +                        'upload_id'    => $internalMetadata['uploadId'],
 +                        'video_result' => $upload->getResult(),
 +                    ]
 +                );
 +                break; // Success. Exit loop.
 +            } catch (\InstagramAPI\Exception\InstagramException $e) {
 +                if ($attempt < $maxAttempts && strpos($e->getMessage(), 'Transcode timeout') !== false) {
 +                    // Do nothing, since we'll be retrying the failed configure...
 +                    sleep(1); // Just wait a little before the next retry.
 +                } else {
 +                    // Re-throw all unhandled exceptions.
 +                    throw $e;
 +                }
 +            }
 +        }
 +        if ($result === null) {
 +            throw new \RuntimeException('Failed to configure video for direct_v2.');
 +        }
```

What's this part for?

```php
 +        if ($result === null) {
 +            throw new \RuntimeException('Failed to configure video for direct_v2.');
 +        }
```

Because if the code above that fails all attempts then the code above throws an InstagramException with the server error message. So this code never runs.

Or have you found a scenario where the code above was needed?

SteveJobzniak:
@valga Alright I just finished reading very carefully through every line of code.

Here are my comments:

1. Holy shit this code is beautiful! The new way for attaching files to requests with a single function call, to make them into a Multipart POST, is **perfect**! I love it so much! :)
2. The question above about $result checking needs to be resolved before merging.
3. About your earlier notes about Instagram no longer using chunking: That's interesting! But we also use the chunking to keep the memory usage low (a 100 MB video = 4x25 MB RAM, instead of hitting RAM limits on the PHP instance), so any rewrite to non-chunked would need to use Guzzle's "multipartstream" or whatever it's called, to stream the file directly from disk.
4. About your note that `"Video upload must be done in exactly 4 chunks" is not true. But this is not important I guess.` I'm not surprised actually. I kept asking mgp because it didn't make sense to me that the server cared about that, and I wanted to make auto-generated amount of chunks of for example 500kb per chunk, but he insisted it needed to be exactly 4 chunks. I guess that since we don't need it to be 4 chunks, we could improve this (later, some other time) to split bigger files into more than 4 chunks, to keep RAM requirements even lower.
5. I noticed that the response type of `changeProfilePicture` has been changed from Response\Model\User, to Response\UserInfoResponse, but the endpoint is the same. I guess our old response type was incorrect? It was a very old API which we haven't touched in ages so they may have changed it. :) Thanks for fixing that in that case!

valga:
> Or have you found a scenario where the code above was needed?

It is just a safety measure against 'bad' behavior. For example, someone can call this method with `$maxAttempts = 0` and the loop will be skipped. Also this method relies on `_sendDirectItem()` which _may_ return null because of someone's fault.

SteveJobzniak:
@valga Oh wow nice catch, I should add the same catch to the other configure function. And I'll change it to UploadFailedException so that people can catch it as an InstagramException.

SteveJobzniak:
@valga Or... perhaps we should instead make sure no code can ever run with maxAttempts < 1? And throw invalidargumentexception if they break that rule.

SteveJobzniak:
@valga ALRIGHT! :-) This is ready to be merged. Thank you SO MUCH for this incredible refactoring. It's beautiful.

I'll separately take care of the maxAttempts validation. Nice discovery! :)

valga:
3. Actually you don't have to put chunks into memory, because Guzzle has streams. Also new upload method doesn't exist in 10.15.0 (which we are locked to because of #1231), so there is no need to implement it at all.
4. I didn't check large videos, but small ones were uploaded in 2 chunks, so it seems there are some rules to determine chunks count.

mgp25:
@valga version could be updated if any feature forces to it. Just tell me the version and ill get all the data

Regards

SteveJobzniak:
@mgp25 In this case it's not any data we need. We're talking about the upload of video data. And the potential of rewriting to use Guzzle streams instead of chunking. Although since I'm going to fully focus on treating my health, I won't touch that code.

@valga Thanks for the clarification. I suspect their app does something like "1 chunk per megabyte" or similar.

I wonder what happens in the new client when uploading a whole file "as 1 chunk", and if that whole file gets dropped by the server while their server is overloaded. Does the server still ask for byte ranges to be reuploaded or does it have to reupload the whole file? If it does the whole file then non-chunking is really bad for us.

And we don't need to fully follow their upload method.

1. Instagram already detects this API so we don't need to "hide" from it.

2. Chunking is a HTTP-layer feature of their web server. Their whole API is just served via HTTP. So chunking will always be supported. Their server will collect our chunks and assemble them and THEN call their API script with the full result as if *we* had sent them the whole thing in 1 big chunk. :) So it doesn't really matter how we do this uploading. And chunks let us retry small parts that were lost.

I think a big improvement would be to change the chunking code to something like `ceil(filesize % 500kb)`, to get a lot of smaller chunks, for faster reuploads when chunks get dropped. It's fine to have many small chunks: When trying to research a "normal" chunk size for HTTP byte ranges, I found that Microsoft IIS uses 256kb by default: https://docs.microsoft.com/en-us/iis/extensions/configuring-application-request-routing-arr/configure-byte-range-request-segment-size-in-application-request-routing

valga:
@mgp25 the question is: do we need a new video upload method? So far, we already have working one.

@SteveJobzniak, you can use Guzzle's streams even with our chunked uploads. And this might lower memory usage significantly.

> I wonder what happens in their new client when uploading a whole file "as 1 chunk", and if that whole file gets dropped by the server while their server is overloaded?

New upload method works like this:
```
do {
  $offset = getOffsetFromServer();
  $done = uploadFileFrom($offset);
} while (!$done);
```

So basically saying, new method always appends data on server. If something has failed (connection loss, etc), server tells client to upload from last saved offset. I guess if something has failed the hard way, server might tell to start from the beginning.

SteveJobzniak:
@valga Thanks for all of this information! I've used it to change the chunking to now aim at <= 512KB chunksize, which means we get low memory requirements: https://github.com/mgp25/Instagram-API/commit/c23708ef157a79950629c67c53580aa8a994bf86

Guzzle streams would be slightly better, but I'm not familiar with how to use them with byte-range uploads. But this code is pretty good and now only uses 0.5 MB of RAM.

As for the new method, ahh `getOffsetFromServer` explains it all... So the server tells us how much data it has... Interesting! I wonder if Instagram did this to fix their constant dropping of chunks during popular hours. Perhaps that method is better than chunking.

But eh, it's okay to keep our current ~512KB method until the day we're forced to change it. And I don't think they will ever force the new method. Because they must always support very old app versions too, so their servers should always have chunking support.

valga:
@SteveJobzniak, it seems old method is not that simple. It always starts with 200KB chunk, but then its size varies a lot: every chunk has its unique size. I guess application (or Instagram) detects upload speed and uses large chunks for fast connections.

SteveJobzniak:
@valga Yeah they probably measure the speed. I picked 512KB chunks because it's small enough to make retries fast, but still large enough to be great for medium & fast connections.

In our old code, a 25MB video was sent as 4x6.25 MB chunks, and `6.25MB > 200KB` and still worked, so we don't have to care what chunksize Instagram officially uses. ^_^

I'm not even sure their scripts can see what chunksize we used.

`data -> http chunked transfer -> Instagram's http server (apache) -> Apache collects our chunks & builds a final file -> then Apache runs their script with the final file`

valga:
> I wonder if Instagram did this to fix their constant dropping of chunks during popular hours. Perhaps that new method is more stable than chunking.

New method is being used by Facebook. It seems Facebook wants to use one infrastructure for everything. Same thing they are doing with WebSocket => MQTT migration.

SteveJobzniak:
@valga That makes sense. But since they don't even use this new method everywhere in their own app yet, we'll just have to wait and see how their own app develops. And Instagram cares extremely much about backwards compatibility with their old app clients. :) I think they'll keep the current method (even when we identify as a newer API) because all old official app versions need to be able to call it.

valga:
@SteveJobzniak, I had some free time and got some information.

1. Old method measures upload speed. The rule is: `chunk_size = previous_chunk_size / elapsed_time_in_msec * 5000`. First chunk size is always 200KB. Also chunk size can't be greater than 5MB.
2. New method is experiment called `ig_android_upload_reliability_universe`. So they are really concerned about uploads reliability. I made an implementation [here](https://github.com/valga/Instagram-API/commit/e4e5dc37d813bf317f24e06b85c0c862951b1752) so we can use it later. Or abandon it, if their experiment will fail.

SteveJobzniak:
@valga Wow, thanks for that great research!

1. Ah I see! That's a cool algorithm. Luckily their script won't know what chunksize we used since that's just between us an the server. So our simpler "use ~512KB per chunk" algorithm is safe.

2. Wow, yeah they really needed to do something about reliability. Their server keeps dropping chunks during high-stress hours and the upload bar kept going forwards and backwards inside the app due to dropped chunks.

Your rur implementation is looking good. A little bit messy and I see TODOs since it was a proof of concept, but that's acceptable and a great start, and it's worth re-visiting that code if their experiment goes well.

Actually, if you're itching for something that's a high-priority thing to look at, then read the offline messages I left on Skype on Saturday. It's about reviewing my committed work of sub-grouping functions. I left a bunch of info about the thoughts and maybe some things to look at. I feel like the groupings are "nearly perfect but not 100%", and that group concept is going to be in the upcoming 3.x release, so another set of your skilled eyes reviewing those would be awesome. They're pretty good already, so if you don't have energy then that's fine. :)

SteveJobzniak:
@valga Thanks, I saw your answer on Skype! I answered you there, just fyi. Anyway, I've been doing some final cleanups now before becoming inactive. The most important one for you is that Request's "addParams" is now called "addParam" (to match "addPost"). ;) All cleanups are now done and I'll only occasionally log into Github and see what's going on. Take care and see you around mate! :)

PR #1286, opened 2017-05-28
Fix pre-login flow

valga:
This one is for @mgp25.

Basically there is nothing important here: I just replaced deprecated endpoint with new one, and tweaked a param for login experiments call.

### DIFF
--- a/src/Instagram.php
+++ b/src/Instagram.php
@@ -381,22 +381,18 @@ class Instagram
     }
 
     /**
-     * Get signup challenge headers.
-     *
-     * Signup challenge is used to get _csrftoken in order to make a successful
-     * login or registration request.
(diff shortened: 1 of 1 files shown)


mgp25:
@valga its 100% correct!

mgp25:
@valga is there any reason you changed it from protected to public?

valga:
Strictly speaking, I did not change the visibility, but the diff looks like I did that. In reality I removed non-used protected method and added new public method.

I guess old method was protected, because it was related with sign up sequence (abused by spamers).

New method just registers an advertising ID, it has nothing to do with sign up. And as far as I remember, you can reset that ID on Android device, so I made it public.

SteveJobzniak:
@valga Sounds good to me!

Issue #1282, opened 2017-05-26
Search in followers & following api


sara89pr:
instagram added search function for follower/following list , is impleted this function in this package?

mgp25:
Nope

Regards

sara89pr:
Can you give me this endpoint to myself I correct?

PR #1292, opened 2017-05-29
Follower & Following Search

mangledbottles:
Commit for request #1282. Adds parameter if requested. Can search through users followers for username.


SteveJobzniak:
Thanks for contributing this feature!

Note that putting the new parameter in the middle (instead of at the end of the parameter list) breaks backwards compatibility. Anyone using this function will get incorrect function call results after this patch.

But I accept it because the next version will be 3.0 which has many compatibility breaks, and I like the query param being before the maxId. :)

mangledbottles:
No problem haha 👍 Trying to help out as I can
Thanks for your constant upkeep of this github :)

PR #1300, opened 2017-06-03
Direct/Realtime update

valga:
List of changes:
- add ability to send/receive disappearing media (both photo and video);
- add ability to send/receive links;
- add ability to send/receive/remove reactions to direct items;
- add validation for almost all user data;
- add ability to provide custom `client_context`;
- refactor `_sendDirectItem()` method;
- remove some duplicated classes.


src/Request/Direct.php
```diff
@@ -650,16 +662,137 @@ public function sendLocation(
...
+     *
+     * @return \InstagramAPI\Response\DirectSendItemResponse
+     */
+    protected function _handleReaction(
+        $threadId,
+        $threadItemId,
+        $reactionType,
+        $reactionStatus,
+        array $options = []
+    )
```
SteveJobzniak: Misformatted, the `)` goes at the end of the last parameter.

src/Request/Direct.php
```diff
@@ -650,16 +662,137 @@ public function sendLocation(
...
+     * @param string $reactionStatus One of: "created", "deleted".
+     * @param array  $options        An associative array of optional parameters, including:
+     *                               "client_context" - predefined GUID used to prevent double-posting.
+     *
+     * @throws \InvalidArgumentException
+     * @throws \InstagramAPI\Exception\InstagramException
+     *
+     * @return \InstagramAPI\Response\DirectSendItemResponse
+     */
+    protected function _handleReaction(
```
SteveJobzniak: This `_handleReaction` is placed before `sendReaction`.

I know that normal programming practice is to put functions before they're used. But these "request collection" class files are going to be read by normal people, so I've placed all internal, protected "utility" functions at the bottom of the file so they don't annoy users.

So please move this function down to the end of the file among the other protected functions.

src/Request/Direct.php
```diff
@@ -391,6 +391,8 @@ public function hideThread(
      *                           of numerical UserPK IDs. To use an existing thread
      *                           instead, provide "thread" with the thread ID.
      * @param string $text       Text message.
+     * @param array  $options    An associative array of optional parameters, including:
+     *                           "client_context" - predefined GUID used to prevent double-posting.
```
SteveJobzniak: I realize that there may be a good purpose, but what _is_ the purpose of making the "random post GUID" user-providable? ;) It's fine since it's optional, but still makes no sense. Is the idea that library users can re-send the same message with the same GUID multiple times so that Instagram auto-prevents the duplicates? I don't get it...

src/Request/Direct.php
```diff
@@ -399,27 +401,30 @@ public function hideThread(
...
-     * @param string      $mediaType  Either "photo" or "video".
+     * @param array  $recipients An array with "users" or "thread" keys.
+     *                           To start a new thread, provide "users" as an array
+     *                           of numerical UserPK IDs. To use an existing thread
+     *                           instead, provide "thread" with the thread ID.
+     * @param string $mediaId    The media ID in Instagram's internal format (ie "3482384834_43294").
+     * @param array  $options    An associative array of optional parameters, including:
+     *                           "client_context" - predefined GUID used to prevent double-posting;
+     *                           "text" - text message;
+     *                           "media_type" - either "photo" or "video".
```
SteveJobzniak: Making metadata configurable as an array is a good thing, it makes user-code clearer about the purpose of each parameter.

However, `media_type` isn't optional. It's always required. And I see that you've made it auto-select "photo" if no value or a non-"video" value was provided. But that just means that most people will miss this "optional" parameter and that the code will therefore submit their videos with media_type as the default "photo".

Something needs fixing here. Either something like:

```
     /**
       * Share an existing media item via direct message to a user's inbox.
       *
       * If you are sharing a video, you MUST provide the "media_type" option and set it to "video".
```

Or something like:

```
     * @param string $mediaId    The media ID in Instagram's internal format (ie "3482384834_43294").
                                 If you are sharing a video, you MUST provide the "media_type" = "video" option.
```

Or something else... perhaps re-phrasing the $options description to say that media_type is NOT optional if you are sharing a VIDEO.

src/Request/Direct.php
```diff
@@ -650,16 +662,137 @@ public function sendLocation(
...
+    }
+
+    /**
+     * Send a link to a user's inbox.
+     *
+     * @param array  $recipients An array with "users" or "thread" keys.
+     *                           To start a new thread, provide "users" as an array
+     *                           of numerical UserPK IDs. To use an existing thread
+     *                           instead, provide "thread" with the thread ID.
+     * @param string $link       An URL to send.
```
SteveJobzniak: Correct grammar is `A URL to send.` or better yet `The URL to send.` :)

src/Request/Direct.php
```diff
@@ -802,113 +937,134 @@ protected function _prepareRecipients(
...
-                $post['upload_id'] = $data['upload_id'];
-                if (isset($data['video_result'])) {
-                    $post['video_result'] = $data['video_result'];
+                $request->addPost('upload_id', $options['upload_id']);
+                // Set video_result if provided.
+                if (isset($options['video_result']) && strlen($options['video_result'])) {
+                    $request->addPost('video_result', $options['video_result']);
+                }
+                break;
+            case 'links':
```
SteveJobzniak: This feature can only send 1 URL at a time (there's only 1 `link_text` field), so perhaps rename this value to `"link"` to match the other singular-named types like "video" and "photo" etc?

src/Response/Model/DirectReactions.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace InstagramAPI\Response\Model;
+
+/**
+ * @method int getLikesCount()
+ * @method null|DirectReaction[] getLikes()
```
SteveJobzniak: Using this to define the available "magic methods" is fine, I guess... And if I make any edits I will be sure to update these too.

Just be aware that I can guarantee that other contributors will forget about updating these (either removing/deleting/renaming them, or updating their types). So they're an extra burden. I guess they're fine in these brand-new classes though. But we have them nowhere else. And they'd be a big burden if we added them everywhere. Hmm.

And you're not documenting the `is` and `set` methods. Not that those matter much though.

I just think this whole thing is a bit bad. I can see the "autocomplete in editor" argument though, so you can decide what to do here. But be aware that these will slowly drift out of sync with the contents of the actual classes because there's _no way_ others will remember to update these.

src/Response/Model/DirectReactions.php
```diff
@@ -0,0 +1,15 @@
...
+ */
+class DirectReactions extends \InstagramAPI\AutoPropertyHandler
+{
+    /** @var int */
+    public $likes_count;
+    /** @var null|DirectReaction[] */
```
SteveJobzniak: Please don't define properties like `@var null|...`. Remove the null everywhere.

We don't control Instagram's server API and they change and deprecate and rewrite things _all the time_. Just recently, they changed `comments` to `preview_comments` in all feeds, and if we had marked every fields as "null|whatever" or "whatever", then people would get a false impression that we're correct. But we can never guarantee that.

So while it's somewhat OK to assume strict types (like with `@var int`). It's not okay to assume which fields will be null.

Just go through all these new Response and Model classes and document every property with _only_ its normal value, not with null. Their API isn't reliable enough that we should be telling people what properties will/won't be null.

src/Request/Direct.php
valga: I can send a message via Realtime. In case of some hiccups there, I will retry with HTTP as a fallback. In both cases I will use the same predefined UUID to prevent double-posting.

src/Request/Direct.php
valga: Actually, it can send more than one URL, but only first one will have a nice display in inbox.

src/Request/Direct.php
valga: I just forgot to remove backwards compatibility. We had `media_type=photo` hardcoded for all medias earlier.

src/Response/Model/DirectReactions.php
valga: Will remove nulls.

src/Response/Model/DirectReactions.php
valga: Speaking of magic methods, why do we have them? Because lack of autocomplete just kills them.

src/Request/Direct.php
SteveJobzniak: @valga Oh okay I suspected it was something like that. I like it. 👍

src/Request/Direct.php
SteveJobzniak: Okay? But the point is that we should properly tell the users to provide `media_type=video` when they share a video, right?

src/Request/Direct.php
SteveJobzniak: Oh O_O, okay!

src/Response/Model/DirectReactions.php
SteveJobzniak: We have them because there are so many different properties, and writing (and remembering to write/edit/update/correctly name) their setters/getters/is-ers is very error prone and takes up a HUUUUUUUGE amount of code in each server response/model class.

For example, `src/Response/Model/Item.php` has 89 properties.

And we'd need this for each:

```php
public $aproperty;

public function getAProperty() {
    return $this->aproperty;
}

public function setAProperty($value) {
    $this->aproperty = $value;
}

public function isAProperty() {
    return (bool) $this->aproperty;
}
```

That's 12 extra lines of code per property. 89 properties x 12 = 1068 lines of extra code in `src/Response/Model/Item.php` to add all the getters/setters/is-ers.

Not everything in this world is about autocomplete. ;) I write using server-replies without autocomplete, and I use getters, is-ers, etc...

src/Response/Model/DirectReactions.php
SteveJobzniak: Thanks 👍  :)

src/Response/Model/DirectReactions.php
valga: It is not just autocomplete. There are code analysis tools, that help to prevent bugs. For example, someone has replaced `$property1` with `$property2` in some response. Code analysis will warn you about that. But it can't work with missing phpdocs for magic methods. It can't warn you about `getProperty1()` is being replaced with `getProperty2()`, because it doesn't know about both of them.

SteveJobzniak: We have a `devtools` folder for special tools. One possible solution to automate this would be to write a tool which recursively loops through everything in `src/Response` (and other folders with server replies in them), and uses the folder-path as the full path to the class, so `src/Response/Model/Item.php` would be interpreted as `\InstagramAPI\Response\Model\Item`, and then uses http://php.net/manual/en/book.reflection.php to instantiate that class.

Then loop through all methods and make a blacklist of hardcoded functions that already explicitly exist (ie `getMessage`).

Then loop through all properties, and use `getDocComment()` for each, to find the `/* @var ...` comment and use that to determine the return type of the property. If no doccomment is found, assume the type as `mixed`.

Next, loop through the list of properties and generate the getX, setX and isX function names from it (turning any `_` to camelcase, so `preview_comments` would be `getPreviewComments`, `setPreviewComments` and `isPreviewComments`).

Then skip the method if it was already in the blacklist (already defined in the class or one of its base-classes as a hardcoded function, ie like `getMessage`)

Then lastly take all this and auto-generate a class `@method` block with all of those auto-inspected functions. And make it open the `.php` file on disk, read all contents, remove the previous class docblock at the top of the file, and put in the new one.

Then we could periodically run that file to keep the magic function declarations up-to-date.

This almost seems like a semi-fun little project, but not for me, no time...

valga: Yeah, it is a good idea. For now, I will remove my phpdocs. But I will figure out something later.

By the way, all classes in Model folder are inherited from `\InstagramAPI\Response`. I think it is a mistake, because `\InstagramAPI\AutoPropertyHandler` is the right parent for them. What do you think?

SteveJobzniak: @valga Yeah, writing a reflector analyzer to auto-generate magic function docs would be a great thing. I think it may even be a lot of fun to code it. Kind of an unusual task. :)

As for the model classes: Yeah `AutoPropertyHandler` would almost always be a better parent, since model objects don't need `isOk()` etc... But the reason why it's parented as `Response` is because sometimes some API endpoints return an object directly + message and status fields. An example of that is:

```php
    public function getUserReelMediaFeed(
        $userId)
    {
        return $this->ig->request("feed/user/{$userId}/reel_media/")
            ->getResponse(new Response\Model\Reel());
    }
```

In the (very, very _few_) cases (maybe just this case? I can't remember) where Instagram responds with a top-level object, being derived from Response lets us use their object directly as a response instead of defining a response with manually copied fields from the object (which is error-prone).

Edit: I just checked...

```php
$ ag "@return.*Model"
Request/People.php
150:     * @return \InstagramAPI\Response\Model\FriendshipStatus

Request/Story.php
87:     * @return \InstagramAPI\Response\Model\Reel
```

Only two places in the code use this feature. But that's why we derived them from Response. :\ I wish Instagram didn't sometimes return raw model objects as the top-level object in their reply.

valga: So, is it safe to inherit all other models from `AutoPropertyHandler` or not?

SteveJobzniak: @valga It would be safe. I just did an alternative query and again I only see those two classes being used that way:

```php
+ $ ag "getResponse.*Model"
src/Request/People.php
155:        return $this->ig->request("friendships/show/{$userId}/")->getResponse(new Response\Model\FriendshipStatus());

src/Request/Story.php
95:            ->getResponse(new Response\Model\Reel());
```

I guess we can do it, since we would get rid of `getMessage()` and `getFullResponse()` etc on all normal model objects.

If we have to re-use an object in the future we'll just have to remember to add that part manually.

I think the safest would be to remind ourselves why we did this. Something like this:

```php
User.php:
class User extends \InstagramAPI\AutoPropertyHandler

Reel.php:
class Reel extends \InstagramAPI\Response // needs to be Response since some APIs use this object as their response
```

SteveJobzniak: @valga I've come up with a way to completely get rid of Response inheritance, but still be able to re-use Reel and FriendshipStatus objects.

```php
<?php

require __DIR__.'/vendor/autoload.php';

// Base autopropertyhandler...
class AutoPropertyHandler {
    public function __call($functionName, $arguments) {
        echo "Called: {$functionName}...\n";
    }
}

// Interface required by all Responses...
// This is necessary so that we can check if something is a response,
// since we cannot easily check if something has a certain trait.
interface ResponseInterface {
    public function getMessage();
}

// Reusable traits (properties and funcs) needed by a response...
// A trait is basically copy-pasted into the class where we "use" this trait.
trait ResponseTrait {
    /** @var string */
    public $message;

    public function getMessage() {
        return $this->message;
    }
}

// Normal response base class...
class Response extends AutoPropertyHandler implements ResponseInterface {
    use ResponseTrait;
}

// Normal model object...
class Reel extends AutoPropertyHandler {
    /** @var float */
    public $extrafield;
}

// This is a special Reel re-use class which extends Reel and manually
// adds the traits (functions) that a Response needs to have.
// Only downside: We cannot do `instanceof Response`... so we'll have
// to rewrite some other code in our library which checks that...
class ReelResponse extends Reel implements ResponseInterface {
    use ResponseTrait;
}

// This is a normal response class which directly extends Response...
class NormalResponse extends Response implements ResponseInterface {
}

$mapper = new \JsonMapper();
$mapper->bStrictNullTypes = false;

$data = json_decode('{"message":"hello world","extrafield":1.234}');

$baseobj = new NormalResponse();
$result = $mapper->map($data, $baseobj);
echo $result->getMessage() . PHP_EOL; // works since this derives from response class
$result->missingFunc();
var_dump($result);
echo "Is a ResponseInterface? ";
var_dump($result instanceof ResponseInterface);

echo "\n------\n\n";

$baseobj = new ReelResponse();
$result = $mapper->map($data, $baseobj);
echo $result->getMessage() . PHP_EOL; // works since this inherits response trait
$result->missingFunc();
var_dump($result);
echo "Is a ResponseInterface? ";
var_dump($result instanceof ResponseInterface);

echo "\n------\n\n";

$baseobj = new Reel();
$result = $mapper->map($data, $baseobj);
echo $result->getMessage() . PHP_EOL; // won't work since reel has no response trait
$result->missingFunc();
var_dump($result);
echo "Is a ResponseInterface? ";
var_dump($result instanceof ResponseInterface);
```

Result:

```
hello world
Called: missingFunc...
object(NormalResponse)#4 (1) {
  ["message"]=>
  string(11) "hello world"
}
Is a ResponseInterface? bool(true)

------

hello world
Called: missingFunc...
object(ReelResponse)#5 (2) {
  ["extrafield"]=>
  float(1.234)
  ["message"]=>
  string(11) "hello world"
}
Is a ResponseInterface? bool(true)

------

Called: getMessage...

Called: missingFunc...
object(Reel)#4 (1) {
  ["extrafield"]=>
  float(1.234)
}
Is a ResponseInterface? bool(false)
```

So with `src/ResponseTrait.php` and `src/ResponseInterface.php` and `src/Response.php` and always using `instanceof ResponseInterface` (instead of `instanceof Response`), we'll be able to re-use model objects as responses without littering them with Response class properties/methods when they're just used as normal objects.

So in this case, we'd have this setup:

```php
src/Response/ReelResponse.php:
class ReelResponse extends Model\Reel implements ResponseInterface {
    use \InstagramAPI\ResponseTrait;
}
```

That would give us "ReelResponse" which would be a fully implemented instance of ResponseInterface, with all ~80 or whatever public properties from the "Model/Reel" object.

And meanwhile, the normal "Model/Reel" object would never contain any Response class features.

If we do this we would have to carefully look at and rewrite all matches for `ag "instanceof.*Response" -C2`, to be 100% sure none of them are going to be affected by the restructuring. I am sure ServerMessageThrower would be affected. The rest seem fine, such as `$item instanceof Response\Model\Item`, because if we ever have some "ItemResponse extends Model\Item" then that would still be true, so that'd be FINE. Therefore it seems like only ServerMessageThrower would need a slight rewrite to accommodate this difference.

The `src/Response.php` file would also need a FAT warning comment somewhere in it, explaining to NEVER check `instanceof Response`, and to ALWAYS use `instanceof ResponseInterface` instead.

What do you think? We would get rid of all ugly Response properties from our model objects, and we would be able to fully inherit model object properties in our responses. And JsonMapper fully supports traits. This idea may be a big win.

valga: Actually I can't see any benefits of using traits. However, interface usage is great. I already told you about that.

So I vote for removing unneeded inheritance from all models (except for those 2) and using interface.

valga:
I believe I fixed everything you have mentioned.

Also, I will not add any new functionality into this PR. The only thing left is to verify media constraints.

SteveJobzniak:
@valga Excellent, I will review everything again.

With the new work I only see one problem. You're putting thread_ids etc in `$externalMetadata`

The purposes are like this:

`$externalMetadata` = things provided by the user, like `'caption'`
`$internalMetadata` = things generated by us internally, like `'thread_ids'`

So this new commit breaks that clean distinction by putting internal direct-share thread metadata in externalMetadata.

I know that it's because the functions in Internal.php mostly take `externalMetadata` and have no `internalMetadata` parameter. Because we had no previous need to generate internal metadata to send to those functions. Those functions all generate that internally.

But in this case of course we need internal metadata, like the thread ID...

So would you be ok with going through Internal.php and making sure that every function that takes metadata has both an internalMetadata and externalMetadata parameter?

A good example of a function that already has that is `configureSinglePhoto`:

```php
     * @param array  $internalMetadata Internal library-generated metadata key-value pairs.
     * @param array  $externalMetadata (optional) User-provided metadata key-value pairs.
...
        array $internalMetadata,
        array $externalMetadata = [])

```

src/Response/Model/DirectReactions.php
SteveJobzniak: @valga There's no point having an interface unless we have multiple classes that implement the interface. That's the whole point of an interface: Telling us what methods the multiple different classes based on it need to implement.

So if we only have a Response class, we have no need for ResponseInterface.

However, if we use _traits_, we'll have Response itself implement the ResponseInterface via the traits, and we'll have the re-usable model responses implement the ResponseInterface via the traits. Zero code duplication. And none of the model objects need to derive from Response anymore.

valga: Actually we are open for _any_ class. For example, someone can call `$instagram->request('/some/endpoint/')->getResponse(new \stdClass());` and we will fail hard. So `ResponseInterface` is kind of necessary thing to add.

As for traits, they will only help with those 2 (now 3) models used as raw responses.

valga:
Will do it later.

src/Response/Model/DirectReactions.php
SteveJobzniak: @valga But that kind of `getResponse` usage by users makes no sense. If they're going to implement an interface with their own buggy versions of getMessage etc, then they may as well just derive "extends Response" in their own custom class to get it auto-implemented, or if they really just want a stdClass they would do `getResponse(new Response())` and then `$response->getFullResponse()` to get the stdClass object. ;)

SteveJobzniak: To rewrite this in a clear and clean way, I guess this would be good:

1. Delete `src/Response.php`.
2. Paste all code (properties and functions) of Response.php to a new file called `src/ResponseTrait.php`
3. Create `src/ResponseInterface.php` and make a list of all functions that must be implemented (the ones from ResponseTrait.php).
4. Edit `Request::getResponse()` and `Client::getMappedResponseObject()` to document that their object param must be a ResponseInterface.
5. Edit `ServerMessageThrower` to make it look for a `ResponseInterface` instead of a `Response`.
6. Edit all classes that "extends Response", and make them do the `extends AutoPropertyHandler implements ResponseInterface { use \InstagramAPI\ResponseTrait; ...` thing.
7. Create new `src/Response/ReelResponse.php` (or whatever name matches the endpoint function call) and make it `extends Model\Reel implements ResponseInterface { use \InstagramAPI\ResponseTrait;`.

With that, no model objects contain any response code anymore, and we can cleanly turn anything into a response (and so can users, they can also use our trait to implement their own ResponseInterface classes)...

SteveJobzniak: @valga I was researching and found that my proposal is a standard practice:

**Symfony**:
- https://github.com/symfony/symfony/blob/582f4753a343f230fbe18b4e9a0747d48351ddfb/src/Symfony/Component/DependencyInjection/ContainerAwareInterface.php
- https://github.com/symfony/symfony/blob/582f4753a343f230fbe18b4e9a0747d48351ddfb/src/Symfony/Component/DependencyInjection/ContainerAwareTrait.php

**PHP-fig's Logger**:
- https://github.com/php-fig/log/blob/d8e60a5619fff77f9669da8997697443ef1a1d7e/Psr/Log/LoggerInterface.php
- https://github.com/php-fig/log/blob/d8e60a5619fff77f9669da8997697443ef1a1d7e/Psr/Log/LoggerTrait.php

**Phan**:
- https://github.com/etsy/phan/blob/f4dd2d9a0cd57be9f729454c3e3b38f6740fa8ac/src/Phan/Language/Element/ConstantInterface.php
- https://github.com/etsy/phan/blob/f4dd2d9a0cd57be9f729454c3e3b38f6740fa8ac/src/Phan/Language/Element/ConstantTrait.php
- https://github.com/etsy/phan/blob/f4dd2d9a0cd57be9f729454c3e3b38f6740fa8ac/src/Phan/Language/Element/FunctionInterface.php
- https://github.com/etsy/phan/blob/f4dd2d9a0cd57be9f729454c3e3b38f6740fa8ac/src/Phan/Language/Element/FunctionTrait.php

And they do it exactly like I proposed above:

- No `Response` class.
- Just a `ResponseInterface` and a `ResponseTrait`.

Example from symfony:

```php
final class LintCommand extends BaseLintCommand implements ContainerAwareInterface
{
    use ContainerAwareTrait;
    ...
```

I will be doing this rewrite today because I need to relax, and this is relaxing.

SteveJobzniak: @valga The rewrite is now complete. Now, the Model objects are completely pure (they only extend from AutoPropertyHandler). No more `getMessage()` and weird properties on them.

The definitions for model objects and the definitions for classes have been changed slightly. It's very easy to upgrade to the new format. Just look at the new src/Response and src/Response/Model folders.

SteveJobzniak:
@valga Okay the Response + Model refactoring is complete. Our Model objects no longer have a bunch of garbage "fullResponse", "message" etc. So now we get much better autocomplete.

Alright, so all that needs to be done in this pull request:

1. Merge the response & model class definition changes from master (easy).
2. Rewrite the few new responses/models to match the new response & model format used by master (easy).
3. Rewrite externalMetadata to correctly name it internalMetadata, since it doesn't contain external/user-provided data.
4. Check the media rules (you mentioned that above).
5. Merge this pull request! :)

valga:
I must say I dislike the misuse of traits. They are made for some kind of multiple inheritance. So, they are good when a class is a mix of response and model, like `UserReelMediaFeedResponse` one. But 95% classes are pure responses, so it is _natural_ for them to inherit from `GenericResponse`.

SteveJobzniak:
@valga Indeed. I just did that because I've seen the kind of damage to this repo normal contributions can do over the years, and I was just paranoid that if I kept some `src/Response.php` or `src/Response/BaseResponse.php` file, then future contributors would think all responses were based on that and would screw up some contribution by doing `if ($response instanceof BaseResponse)` to check if something is a response. Which would seem to work but would silently screw up the 3-4 responses that are instead based on model objects.

Even if we rename GenericResponse to BaseResponse (a better name when inheritance is involved), and put a FAT warning comment in BaseResponse.php telling people to NEVER check `instanceof BaseResponse` and to always use `instanceof ResponseInterface`. Then, people may _still_ miss it because they'll see that every response they look at is "extends BaseResponse" (98% of them would use that) so they'll just never know about ResponseInterface at all, and will just go "instanceof BaseResponse"...

Then, being a bit tired when merging such a contribution, and voila the bug is complete... and silent, and stays forever, and will possibly never be detected since only 2% of responses are derived from model objects instead.

So since I've seen the huge damage from random contributions in the past I valued code safety over code shortness. :\ I wish we didn't have to do that but it's unfortunately a real problem and was a huge part of why 1.x was so awful and messy and buggy, and that's why I am ensuring it doesn't happen again.

src/Request/Internal.php
```diff
@@ -212,16 +213,26 @@ public function configureSinglePhoto(
...
-                    ->addPost('client_timestamp', time())
+                    ->addPost('client_shared_at', (string) time())
+                    ->addPost('source_type', '3')
+                    ->addPost('configure_mode', '1')
+                    ->addPost('client_timestamp', (string) (time() - mt_rand(3, 10)))
+                    ->addPost('upload_id', $uploadId);
+                break;
+            case 'direct_story':
+                $request
+                    ->addPost('recipient_users', $externalMetadata['recipient_users'])
```
SteveJobzniak: external -> internal

src/Utils.php
```diff
@@ -332,7 +332,7 @@ public static function throwIfIllegalMediaResolution(
     /**
      * Verifies that a video's details follow Instagram's requirements.
      *
-     * @param string $targetFeed    Target feed for this media ("timeline", "story", "album" or "direct_v2").
+     * @param string $targetFeed    Target feed for this media ("timeline", "story", "direct_story", "album" or "direct_v2").
```
SteveJobzniak: Verifying photo/video constraints for direct_story?

valga: It seems that direct_story has the same constraints as ordinal story.

As for direct_v2, there are changes coming http://blog.instagram.com/post/161060792262/new-in-direct-support-for-landscape-and-portrait so I will verify it later.

valga:
> Do you have any solution?

Nope I don't have a solution against that bad contributions. So let's keep everything as it is now.

src/Request/Internal.php
valga: Solved.

SteveJobzniak:
@valga Hehe, you should see the kind of contributions happening in 1.x. Like already having a `src/Utils/SettingsAdapter.php` and `src/Utils/SettingsMysql.php`, and someone committed a brand new folder with `src/Utils/SettingsAdapter/SettingsInterface.php` and `src/Utils/SettingsAdapter/Memcached.php`, with the new "memcached" using the new interface. But zero rewrites of the old files. Not even moving the old ones into the new settingsadapter folder.

I don't expect that kind of people to understand what class to look for if they think all responses are always "BaseResponse". A name like that gets confusing and I would even forgive myself for believing all responses must be from BaseResponse.

It's just that I try very hard to avoid risks of 1.x re-appearing in the code. By making code very clear to avoid people adding bugs. As you've noticed with my review of comments.

And speaking of that: Great work on this pull request. It almost loos done! I just wonder about this: https://github.com/mgp25/Instagram-API/pull/1300/commits/6e4953ce29a241db5cacd3744960c21a5a34981d#diff-8196fda03bd99e941e07599792a645a4R356

On all other feeds there's a lower limit, videos cannot be less than 3 seconds. No such limit on direct story? Can we send a 0.5 second video?

valga:
> On all other feeds there's a lower limit, videos cannot be less than 3 seconds. Is there no such limit on direct story? Can we send a 0 second video?

We can send 0.1 second video to direct. And I fixed constraints to use it as lowest allowed duration.

So, this PR is ready to merge.

I will add support for direct portrait/landscape media in a separate PR when this functionality will come to Android.

SteveJobzniak:
@valga Thank you so much. Excellent work as always! And nice discovery about `sendPost`. The name `sendShare` was fine too (it's "sharing" media), but always nice to find the official name for it!

Merging now <3

PR #1301, opened 2017-06-03
Create autodoc script for auto-generation of PHPDoc for magic methods

valga:
Usage: `php devtools/autodoc.php`

An example of auto-generated docs:
```diff
--- a/src/Realtime/Action/Ack.php
+++ b/src/Realtime/Action/Ack.php
@@ -4,6 +4,14 @@ namespace InstagramAPI\Realtime\Action;

 use InstagramAPI\Realtime\Client;

+/**
+ * @method \InstagramAPI\Realtime\Action\Payload\Ack getPayload()
+ * @method mixed getStatusCode()
+ * @method bool isPayload()
+ * @method bool isStatusCode()
+ * @method setPayload(\InstagramAPI\Realtime\Action\Payload\Ack $value)
+ * @method setStatusCode(mixed $value)
+ */
 class Ack extends \InstagramAPI\Realtime\Action
 {
     public $status_code;
```


SteveJobzniak:
This is great. Looks like this idea required less code than I thought!

I noticed some problems not related to this class... but actually some cases that our magic setters/getters may not be handling:

In Model/Item:
public $iTunesItem;

This property is already camelcase... so the magic accessors currently don't even work for it.

In Response/TimelineFeedResponse:
public $_messages;

This has a leading underscore.

I'll investigate both of these problems in a moment.

SteveJobzniak:
Only one has leading underscore:

```php
+ $ ag "public.*\\\$_"
instagram-php/src/Response/TimelineFeedResponse.php
18:    public $_messages;
```

And for camelcase (skipping results with uppercase letters in trailing comments):

```php
+ $ ag "public.*[A-Z]"
Response/Model/AndroidLinks.php
7:    public $linkType;
8:    public $webUri;
9:    public $androidClass;
11:    public $deeplinkUri;
12:    public $callToActionTitle;

Response/Model/Item.php
180:    public $iTunesItem;

Realtime/Event/Patch/Op.php
26:    public $doublePublish;
```

valga:
Actually `_messages` property does not belong to `TimelineFeedResponse`. It is used in failed responses.

SteveJobzniak:
@valga Okay, I see. Well I've fixed AutoPropertyHandler now for camelcase and leading underscore scenarios. I'll push in a moment.

SteveJobzniak:
@valga Okay the fix is up:

https://github.com/mgp25/Instagram-API/commit/2b05fc972045e5fc9714df6b4a3e174fb33151e3

So as you can see, your auto-generated `getITunesItem` is already the correct format.

The only thing in the docs that's incorrectly generated is `@method Model\_Message[] getMessages()`

Modify the code like this (pseudo-code):

```php
if (strncmp($rawPropName, '_', 1) === 0) {
    // function should be [get/set/is].'_'.$camelCasePropName
}
```

So: `@method Model\_Message[] get_Messages()`

valga:
Done.

SteveJobzniak:
@valga Excellent. Interesting that you also added support for more than one leading underscore, by looping from the start through all underscores and adding one for each. Well, I guess that's okay. I only added support for a single leading underscore to the AutoPropertyHandler because that's the only scenario we've seen from Instagram

devtools/autodoc.php
```diff
@@ -0,0 +1,164 @@
...
+            $phpDoc = $property->getDocComment();
+            if ($phpDoc === false || !preg_match('#@var\s+([^\s]+)#i', $phpDoc, $matches)) {
+                $type = 'mixed';
+            } else {
+                $type = $matches[1];
+            }
+            $rawName = $property->getName();
+            // Convert property_name to PropertyName.
+            $name = str_replace(' ', '', ucwords(str_replace('_', ' ', $rawName)));
+            // Fix for __property_name.
```
SteveJobzniak: This comment could be a bit better for our future re-reading understanding. Perhaps:

`// Prepend any leading underscores to the PropertyName.`

SteveJobzniak:
@valga I've edited AutoPropertyHandler to understand multiple leading underscores, just like in your code. Just in case future properties have more than one leading underscore.

Oh and after the Response + Model refactoring (which has changed every response and model file), you'll need to merge master into this pull request and then re-generate the docs again.

SteveJobzniak:
Here's the easiest way to fix the 170 merge conflicts (haha): Do a git rebase interactively. Drop your "generate autodocs" commits (comment them out). Then, with the autodocs removed, just force-push this pull request branch so that the autodocs aren't being added. The result will be zero merge conflicts.

I would also in general prefer this solution. We can generate autodocs separately via a normal commit later. Right now this pull request is modifying 170 files which makes it hard to read the changes. ;D

SteveJobzniak:
I was testing it out and found a small bug:

```php
/** Hello world */
class ActivityNewsResponse extends AutoPropertyHandler implements ResponseInterface
{
...
```

Will lead to:

```php
/** Hello world */
 * @method mixed getAdsManager()
...
```

Possibly solvable by extracting all old text via `#/\*\*\s+(.+?)\s*\*/#s`, then re-writing the whole docblock...

valga:
I removed all auto-generated docs from this PR, so they can be added later by portions.

Also autodoc is able to normalize any existing PHPDoc and remove empty ones. One-liners is supported too.

And autodoc is able to handle ```__properties__with__unlimited__underscores__``` now.

SteveJobzniak:
@valga Fantastic, it now handles all test cases I throw at it, like single-line docblocks, or removing all properties, and putting extra blank lines in there, etc.

And the object references it uses are correct too since they're relative when necessary (ie `Model\Item`), or full paths when certain classes needed to use that.

I can't see any problems at all. It generates correct docs, overwrites correctly, merges correctly with existing docs (and skips old `@method` lines), etc.

It seems perfect. Is it ready to merge?

valga:
It is ready.

SteveJobzniak:
@valga <3

SteveJobzniak:
@mgp25 Okay this new tool is merged. Since you're working hard on adding responses/model items, just remember to save the new PHP class and then run "php devtools/generateAutoPropertyHandlerDocs.php" before committing each new model/response. It will automatically inject a comment into the PHP file, with the analyzed functions that the object will have. This is for people who write using IDEs. Thanks to this, their IDE will show stuff like `$item->get<press tab>` and show a list of `$item->getPk()`, etc. Autocompletion works thanks to this new Reflection tool we've come up with. :)

@valga Oh and I've now created a quick code style checker, to automate the work of finding functions and variable misnamings that sometimes slip through:

https://github.com/mgp25/Instagram-API/commit/d7bc6366206d66442e33d71649f6e52d148af379

mgp25:
@SteveJobzniak skype

PR #1322, opened 2017-06-12
Realtime: Fix namespace issue on some PHP versions

### LINKED ISSUE #1321: mgp25/instagram-php/src/Realtime.php on line 7
mgp25/instagram-api version dev-master
error in realtimeClient.php ->
Fatal error: Cannot use InstagramAPI\Realtime\Client as Client because the name is already in use in /Applications/XAMPP/xamppfiles/htdocs/mnmn/vendor/mgp25/instagram-php/src/Realtime.php on line 7

valga:
Fixes #1321.

### DIFF
--- a/src/Realtime.php
+++ b/src/Realtime.php
@@ -4,8 +4,7 @@ namespace InstagramAPI;
 
 use Evenement\EventEmitterInterface;
 use Evenement\EventEmitterTrait;
-use InstagramAPI\Realtime\Client;
-use InstagramAPI\Realtime\Event;
+use InstagramAPI\Realtime\Client as RealtimeClient;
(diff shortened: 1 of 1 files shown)


src/Realtime.php
```diff
@@ -4,8 +4,7 @@
 
 use Evenement\EventEmitterInterface;
 use Evenement\EventEmitterTrait;
-use InstagramAPI\Realtime\Client;
-use InstagramAPI\Realtime\Event;
```
mgp25: You have removed `use InstagramAPI\Realtime\Event;`

Is that okay?

valga: Yes, it is ok, because it was unused `use` statement.

PR #1350, opened 2017-06-22
Various improvements

valga:
1. bc97485 splits `syncFeatures()` in two methods and calls them both on experiments refresh.
2. 1547bee adds new methods used in pre-login flow (since 10.24.0). These methods are marked under research and not used for now.
3. 6e2371b updates our most used method `getTimelineFeed()`.

### DIFF
--- a/src/Instagram.php
+++ b/src/Instagram.php
@@ -449,6 +449,9 @@ class Instagram
             // Call log attribution API so a csrftoken is put in our cookie jar.
             $this->internal->logAttribution();
 
+            // Uncomment this call when IG_VERSION >= 10.24.0
+            //$this->internal->readMsisdnHeader();
+
             try {
                 $response = $this->request('accounts/login/')
                     ->setNeedsAuth(false)
--- a/src/Instagram.php
+++ b/src/Instagram.php
@@ -615,7 +615,7 @@ class Instagram
             // This also lets us detect if we're still logged in with a valid session.
             try {
                 $this->timeline->getTimelineFeed(null, [
-                    'is_pull_to_refresh' => true,
+                    'is_pull_to_refresh' => mt_rand(1, 3) < 3,
                 ]);
             } catch (\InstagramAPI\Exception\LoginRequiredException $e) {
                 // If our session cookies are expired, we were now told to login,
--- a/src/Instagram.php
+++ b/src/Instagram.php
@@ -444,7 +444,7 @@ class Instagram
 
         // Perform a full relogin if necessary.
         if (!$this->isLoggedIn || $forceLogin) {
-            $this->internal->syncFeatures(true);
+            $this->internal->syncDeviceFeatures(true);
 
             // Call log attribution API so a csrftoken is put in our cookie jar.
             $this->internal->logAttribution();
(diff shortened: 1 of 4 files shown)


mgp25:
@valga so far seems good.

1. Seems good, nice and clean
2. What version were you using? Im using 10.27.0 and i havent seen the `accounts/read_msisdn_header/` endpoint. Will check it when i have more time.
3. Seems okay. I haven't checked it but i have seen that header before so i guess its okay.

And since 2. `readMsisdnHeader` required msisdn, we can implement something like `setMsidn` or even `setPhoneNumber`. But we need to be cautious. Instagram isnt allowed to take that information if you are running the app over wifi. API is emulating wifi behaviour. I haven't checked that so far, but in theory we dont need to provide that info to them. I also saw in the description that is encoded, is encryption left in that function.

What are your thoughts @valga ?

valga:
I am using 10.26.0.

App calls `accounts/read_msisdn_header/` endpoint without `subno_key` param when on WiFi, so it is safe to uncomment that line with `readMsisdnHeader()` in pre-login flow in future.

When on mobile network, app also calls `accounts/msisdn_header_bootstrap/` endpoint. But its response doesn't contain any data (except for "ok" status), so I was unable to capture "normal" msisdn flow. It seems this feature is in development, but it already makes some calls.

And I will check what 10.27.0 does, but later.

SteveJobzniak:
Hi valga. Did a readthrough and code-quality wise these are all fine. :)

SteveJobzniak:
Actually the only part I reacted to was this:

`"latest_story_pk" The media ID in Instagram's internal format (ie "3482384834_43294").`

Is that accurate? Because "PK" (permakey) is the "3482384834" part, and "ID" is the whole "3482384834_43294" (mediaID_userID). So just checking. Sometimes Instagram uses the terms inconsistently so perhaps latest_story_pk indeed takes the ID as parameter.

valga:
> So perhaps latest_story_pk indeed takes the full media ID as parameter.

It really does that.

SteveJobzniak:
@valga As far as the technical protocol changes, there's a lot in getTimelineFeed but it seems fine.

However, I noticed `is_pull_to_refresh` is now set to true on the "resume session" login flow. "Pull to refresh" is the action of pulling down to force the timeline tab to refresh. Is that really what is sent to Instagram when the user switches away from the app and then returns to the app? Because our resume is meant to emulate what happens when the user goes to a different Android app, then switches back to Instagram. Otherwise we'll look like one long never-closed app session.

valga:
> Is that really what is sent to Instagram when the user switches away from the app and then returns to the app?

App never calls `feed/timeline/` when going back and forth: you need to manually refresh your feed. Actual behavior depends on last opened tab: almost every time app calls `direct_v2/inbox/` to init Realtime (because it runs in foreground only), and for timeline tab app calls `feed/reels_tray/` endpoint.

But app might call `feed/timeline` without `pull_to_refresh` flag, so I put some randomness into its value.

SteveJobzniak:
@valga Ahh I see. Thanks for checking and tweaking that behavior a bit, and for explaining the resume workflow. A bit of randomness looks great to me. :)

@mgp25 If you are also okay with all changes, we can merge this.

mgp25:
@SteveJobzniak seems good. You can merge it.

PR #1351, opened 2017-06-22
Improve ServerMessageThrower

valga:
Existing detection method that works on human-readable server messages is unreliable. In the same time, almost all server errors contain `error_type` field. So we should use it as additional source to check against.

Also this PR fixes detection for `InvalidUserException` and adds new `InvalidSmsCodeException`. And I used real responses for `exceptionTest` script. I don't have responses for `AccountDisabledException` and `ForcedPasswordResetException` though.

### DIFF
--- a/devtools/exceptionTest.php
+++ b/devtools/exceptionTest.php
@@ -8,28 +8,33 @@ require __DIR__.'/../vendor/autoload.php';
 use InstagramAPI\Exception\ServerMessageThrower;
 
 /*
- * Emulates various server message strings and verifies that they're mapped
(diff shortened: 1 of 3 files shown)


SteveJobzniak:
Thanks. I agree, the previous detection method would fail any time Instagram rewrites their error "message" field. Which sometimes happens. So thanks for adding "error_type" support!

I fixed the forgotten "break 2" -> "break 3" update (since there are now 3 loops) so that it properly exits as soon as it finds a matching exception.

Merging! 👍

PR #1367, opened 2017-06-28
Exceptions: Add NotFoundException and BadRequestException

valga:
Instagram API is inconsistent when it comes to errors. Sometimes it gives you JSON response, sometimes you receive nice HTML page. But almost every time you have a proper HTTP status code. However, it is available at low level only, so you can't check it, for example in `ServerMessageThrower`.

So, this PR pushes HTTP response to higher levels, and adds NotFoundException (404) and BadRequestException (400). These exceptions are thrown _after_ critical ones, so it doesn't break any important stuff.


SteveJobzniak:
Awesome! But not sure this is correct...

Instagram is indeed weird.

And one of those weirdnesses is this common scenario:

HTTP CODE: 404.
JSON: {"message":"No such user.", "status":"fail"}.

This is the reason why I didn't add any "notfound exception" throw on 404 etc. Actually, I added such an exception and then removed it again.

Instead we ignore the http code completely and just parse the JSON, which has the full information.

I am making you aware of this, and hopefully this scenario is not a problem in this implementation of HTTP status code handling?

SteveJobzniak:
```
 -        // NOTE: Detecting "404" errors was intended to help us detect when API
 -        // endpoints change. But it turns out that A) Instagram uses "valid" 404
 -        // status codes in actual API replies to indicate "user not found" and
 -        // similar states for various lookup functions. So we can't die on 404,
 -        // since "404" API calls actually succeeded in most cases. And B) Their
 -        // API doesn't 404 if you try an invalid endpoint URL. Instead, it just
 -        // redirects you to their official homepage. So catching 404 is both
 -        // pointless and harmful. This is a warning to future contributors!
 -        // ---
 -        // case 404: // "404 Not Found"
 -        //     die("The requested URL was not found (\"{$uri}\").");
 -        //     break;
```

I did a quick readthrough so far, and the above was what ai was referring to.

Have you found a scenario where 404 is actually useful? Invalid endpoint URLs don't cause 404. It's only used by valid endpoints, for valid replies, when they cannot find the user/media/post.

At least these new exceptions are derived from EndpointException, so they won't break the "catch EndpointException" workflow. :)

valga:
> Have you found a scenario where 404 is actually useful?

Can you give me a reliable way to check whether user/media/whatever exists? JSON is not an answer, because:
1. Sometimes you don't have JSON at all, because Instagram gives you nice HTML page (for example, call `$instagram->people->getInfoById(1);`. Existing code would throw `EmptyResponseException`. And you can't tell what happened from it, because the same exception would be thrown on 502 Bad Gateway errors.
2. Checking against error messages is not reliable, because they are subject to change.

Actually, it is a flaw of our response handling system, because we don't keep/provide HTTP response to user.

valga:
Also, take in mind that I don't check for 404 error right after receiving a response from Instagram. I let it pass to the response mapper, then check if it has `"status":"fail"`, then check it against critical errors (such as login/checkpoint/whatever required), and only at the end of this flow I throw `NotFoundException`.

SteveJobzniak:
@valga Yeah, Instagram's API is very dumb.

Problem 1: They misused 404 "Not Found", which is meant for URLs that cannot be found at all. They were supposed to use 204 "No Content", which means: Request was OK, but no content exists for your request.

Problem 2: If you try an invalid API URL, they redirect you to instagram.com with a real HTML reply, instead of giving you a good 404 error. Idiotic. So our API thinks it's "Empty Response"... since JSON decoding fails.

Problem 3: The reason we used "json decoding" to detect "empty response" was literally because Instagram's HTTP status codes are a total mess. Sometimes 200. Sometimes 404. Sometimes god knows what!

So no, I don't know any reliable way to handle Instagram's mess of status codes and JSON mix. And yeah I was aware of getInfo for an invalid user ID giving a instagram.com page with 404... it's all so dumb.

Well, as long as your new code lets all regular JSON decoding and handling happen, and only uses 404 handling as a fallback after all other exceptions are handled (if no other exception was found), and still derives the 404 exceptions from EndpointException (since it's an endpoint error), then this should be fine. I will review this in more detail in a moment. Anything I should think about when I review?

valga:
I don't remember if I did a proper testing of video uploads, since they use raw (non-JSON) responses. So please double-verify that scenario.

SteveJobzniak:
@valga Okay will do! I'll start checking now.

SteveJobzniak:
@valga Thoughts so far:

1. I like the api() rewrite to make it do the decoding to the target object instead of having Request handle the decoding. This change should have been done as a separate commit, but that's ok this time. :)

2. The flow of the code looks correct when I read through it carefully.

3. Many comments need updating, because the code under them was changed to do something else. But I'll take care of all of those in separate commits after merging.

I've got some more reading to do and then some tests.

SteveJobzniak:
@valga Okay I got sidetracked by an email that came in and had to do a task.

Okay so I have finished the review. All code changes are good, and I like that these new exceptions are just a fallback, and that they are based on EndpointException, and that they are triggered with the real server JSON "message" via autoThrow AND via getMappedResponseObject whenever we faile ot decode the server response. This is pretty much perfect.

As for the video upload scenario: Yeah, you don't have access to Guzzle's HTTP response, so in that case we're still using "getMappedResponseObject" which will be unable to look at the server HTTP code. But ehh that's fine, the errors "NotFound" and "BadRequest" don't happen on the video upload endpoint, and we still check JSON for validity if an error happens during video upload, so it's totally fine.

Thanks for this great work! Time to merge and just fix a few comments, and perhaps a few variable names for clarity. But the code itself and the structure is already great! :)

SteveJobzniak:
@valga Actually, I noticed that _apiRequest returns Guzzle's response too, as ['response']. So I'm adding that tweak to the video uploader's getMappedResponseObject too (which is done when the final chunk is done and we get an API response), so that we always check the HTTP status code in all of our internal usage of the API. Checking the code during video upload should work fine, since we always get 200 OK from those endpoints.

SteveJobzniak:
@valga All good now, merged and tweaked:

https://github.com/mgp25/Instagram-API/commit/60cd407661012914a4a4326b77284f5d6be285dc

Great job. And I tried the `$instagram->people->getInfoById(1);` case again. It's much nicer now, getting a `NotFoundException` (`EndpointException`) with "Requested resource does not exist." as the message, instead of an `EmptyResponseException`.

Now people can try{} catch(EndpointException $e){} as usual and even catch it when users are not found.

Thanks a lot for a nice solution to a long problem! Your implementation, with the highest priority towards first doing JSON-decoding + looking for other exceptions is great and solves all problems. :-)

SteveJobzniak:
@valga Btw, since we talked about "_messages" in the past and you thought it should be in ResponseInterface instead and that it was used "in case of errors/failed endpoint calls", actually it's not. It's already correct the way it is and was indeed only used in "feed/timeline/". I've now seen it in a response. This is just FYI so that you also know this:

```
GET:  feed/timeline/?rank_token=<censored>&ranked_content=true
> 200 	 228B
RESPONSE: {"num_results": 0, "more_available": false, "auto_load_more_enabled": true, "feed_items": [], "is_direct_v2_enabled": true, "next_max_id": null, "status": "ok", "_messages": [{"time": 1498737314.27, "key": "update_push_token"}]}
```

valga:
Yeah, I was wrong about its purpose. But it really can be in every response. Also, here is a list of supported messages: fb_needs_reauth, vkontakte_needs_reauth, twitter_needs_reauth, ameba_needs_reauth, update_push_token.

SteveJobzniak:
@valga Oh, hmm. So it's an optional field in every response. That's annoying... (adding all that extra code to the base ResponseInterface/ResponseTrait = ugh). Maybe we'll fix that later, ugh. ;-)

SteveJobzniak:
@valga Okay I have moved _messages to the core: https://github.com/mgp25/Instagram-API/commit/ec3be752b071e5df46799154bc29d70cf4e42a3a

Annoying to pollute the core with such a rare field, but it was inevitable I guess. Thanks for reporting the issue. :)

PR #1369, opened 2017-06-28
Add SASL authentication support to Memcached

ddzobov:


SteveJobzniak:
Thanks a lot for contributing, I will check this tomorrow. :)

SteveJobzniak:
Okay so... this code is not valid. I guess you haven't tried it before submitting.

In Memcached.php you're adding a `$locationConfig['sasl']` check. But $locationConfig can never contain 'sasl', because you didn't edit the Factory.php file which is where $locationConfig is created.

I'll research what SASL is and implement this properly in a manual commit instead. Thanks for the idea, whatever it is. :-) I guess people need SASL.

ddzobov:
Yes, i tried this with custom class and forgot about it...)

Thank you, i'll wait for changes.

SteveJobzniak:
Oh ok, hehe :) Well I have now finished it:

https://github.com/mgp25/Instagram-API/commit/920ebb178f7ab90597b6e5078141d7a8a71da51b

Please try if this works. I don't have any SASL memcached server.

The value is provided via 'sasl_username' and 'sasl_password' in the storageconfig. Both must be provided.

So: [ 'storage' => 'memcached', 'sasl_username' => ......

PR #1385, opened 2017-07-06
Minor Change in ImageAutoResizer.php

vorujov:
FYI: `sys_get_temp_dir()` might return unwritable directory. Since this library is being used the InstagramAPI internally, having a option to set `$_tmpPath` might be good. 

I added public static variable `$defaultTempPath`. Which can be set as `\InstagramAPI\ImageAutoResizer::$defaultTempPath = '/path/to/temp/dir/';` 

Default value of `$defaultTempPath` is null, which means that use `sys_get_temp_dir()`


SteveJobzniak:
Thanks! This is a good idea but I'll have to tweak it a bit to document it better so people understand it. Tomorrow I'll merge. :)

SteveJobzniak:
Thanks, included in core now as https://github.com/mgp25/Instagram-API/commit/5a737049332668cfeb246b1ea769396202cb5eb3

I made a new commit because there was a lot to tweak so it was easier for me that way.

Thanks for the great idea! :)

SteveJobzniak:
You will also want to set this one:

https://github.com/mgp25/Instagram-API/commit/a1fa3fadc7ac5d222a1a5bc350d3278e93919c78

This will fix all places in the code that uses sys_get_temp_dir(), for you.

Have fun! :)

Issue #1353, opened 2017-06-23
Failed to init Realtime: Can not subscribe to inbox


rpw98:
I was trying to test realtimeClient example but I got this error:

Failed to init Realtime: Can not subscribe to inbox

For whatever reason, maybe a new Instagram change, recent "direct_v2/inbox" calls returns no subscription information. Maybe that is the cause?

Thank you!

dish615:
i'm getting this as well! is there another way to get the subscription info?

dish615:
i'm using Charles to sniff and i noticed a new uri endpoint - /api/v1/push/register/. this might be how instagram switched it's sub method? i don't know though...

valga:
Well, Instagram is deprecating WebSocket client (Telegraph) in favor of MQTT one since 10.26.0. It seems that someone has failed there, because I have `"is_telegraph_enabled":"1"` in my experiments, but `direct_v2/inbox` doesn't have `subscription`. Or they made it intentionally to force kill Telegraph, I don't know.

mgp25:
@valga although PR was okay, error persists. I have reopened this issue so you can check it later

valga:
@mgp25 hmm, the only scenario to have mentioned error now is:
1) receive valid subscription model in inbox response;
2) start websocket client;
3) receive event with must_refresh flag;
4) receive null for subscription model in inbox response.

So, either you have _another_ error or Instagram is targeting our implementation. Post your RTC debug log, and I will verify that.

mgp25:
@valga i tested `realtimeClient.php`.

**Debug log**
(Some data and responses are omitted)

```
POST:  feed/timeline/
RESPONSE: OK

POST:  feed/timeline/
RESPONSE: OK

[13:33:36] [rtc] starting websocket client

GET:  direct_v2/inbox/?persistentBadging=true
RESPONSE: OK

[!!!] Failed to init Realtime: Can not subscribe to inbox.
```

valga:
Your log is from unfixed version, because it starts WebSocket client before checking `direct_v2/inbox/` response.

mgp25:
@valga you are right, totally my fault. I thought i synced my code with master, it wasnt. 

Everything running flawlessly.

Regards 😄

rpw98:
Thank you. But how to test it? When I run the realtimeClient example, it stopped here:

[...]

PR #1387, opened 2017-07-06
Realtime update

valga:
There is nothing new here: it is just a maintenance. See commits for changelog.


SteveJobzniak:
Thank you so much, this is freaking awesome! I wonder about the 10.29.0 commit which removes websocket. Is it safe to merge this while our non-subscriber constants are at 10.22.0?

SteveJobzniak:
Ehh, I'll merge it! If we find some problem with constants we can deal with it somehow.

valga:
See #1353. Instagram doesn't have WebSocket server anymore.

SteveJobzniak:
Ahh perfect! So it's safe. :)

By the way, you said you are waiting for some constant bump for a feature... which version is it you need?

We'll have to consider bumping soon, because Instagram Live Replays exist in all story feeds for newer clients, so we'll have to see what version to bump to.

SteveJobzniak:
10.26 added Replays: https://hypebeast.com/2017/6/instagram-stories-replay-live-video

I think that was the exact version you were waiting for?

valga:
Yeah, I mentioned 10.26.0 as a good candidate to bump to. And I confirm that it is needed to get `post_live` fields (aka live replays) in responses.

SteveJobzniak:
@valga Okay so it's settled, I'll bump to 10.26.0 in a moment. Have you got any username I can test with to look for the post_live field and implement its model object? I can't find anybody who has saved a replay, but it sounds like you have someone.

valga:
@SteveJobzniak jamieleesix. Also we need to update (login) experiments.

mgp25:
@valga will update experiments and so tonight.

PR #1388, opened 2017-07-06
Update models to return live replays

zuexo:
current user id with replays: 2054256522


SteveJobzniak:
Perfect. Thank you! I was just working on this, and this looks right. Merging! :)

SteveJobzniak:
@zuexo Verified. This is all correct. Thanks again! ❤️

PR #1390, opened 2017-07-09
Refactor requests

valga:
I decided to refactor requests before going on new video uploader. So, `Client` now is liberated from last file upload method, and `Request` takes full responsibility on building HTTP request. I made a separate PR to discuss these changes.

Also, I tweaked video chunks uploader a bit: from now, it uses HTTP status code to decide what happened and it uses `Range` header instead of body response (because app does that).


src/Client.php
```diff
@@ -647,26 +623,22 @@ protected function _buildGuzzleOptions(
...
      */
     protected function _guzzleRequest(
-        $method,
-        $uri,
-        array $guzzleOptions = [])
+        HttpRequestInterface $request)
     {
         // Add critically important options for authenticating the request.
-        $guzzleOptions = $this->_buildGuzzleOptions($guzzleOptions);
+        $guzzleOptions = $this->_buildGuzzleOptions();
```
SteveJobzniak: What is this? It looks like you've made it impossible to override Guzzle request options per-request?

src/Client.php
```diff
@@ -675,12 +647,12 @@ protected function _guzzleRequest(
...
-            break;
-        // WARNING: Do NOT detect 404 and other higher-level HTTP errors here,
-        // since we catch those later during steps like getMappedResponseObject
-        // and autoThrow. This is a warning to future contributors!
+            case 429: // "429 Too Many Requests"
+                throw new \InstagramAPI\Exception\ThrottledException('Throttled by Instagram because of too many API requests.');
+                break;
+            // WARNING: Do NOT detect 404 and other higher-level HTTP errors here,
+            // since we catch those later during steps like getMappedResponseObject
+            // and autoThrow. This is a warning to future contributors!
```
SteveJobzniak: This change is just weird. Switch and its cases being on the same level is a common pattern since it avoids overly indented code. Same as an if/elseif/else being on the same level.

src/Client.php
```diff
@@ -712,431 +684,92 @@ protected function _guzzleRequest(
...
-     *                                                    class decode failed.
-     * @throws \InvalidArgumentException                  If no object provided.
-     *
-     * @return array An array with the Guzzle "response" object, and the raw
-     *               non-decoded HTTP "body" of the request, and the "object" if
-     *               the "decodeToObject" library option was used.
+     *
+     * @return HttpResponseInterface
      */
     protected function _apiRequest(
```
SteveJobzniak: I see that you moved a lot of the responsibilities of this func to other functions. A lot of changes, so I'll just assume this is fine. ;)

src/Request.php
```diff
@@ -259,39 +483,107 @@ protected function _getRequestBody()
...
+        // Prevent request from sending multiple times.
+        if ($this->_httpResponse === null) {
+            if ($this->_needsAuth) {
+                // Throw if this requires authentication and we're not logged in.
+                $this->_throwIfNotLoggedIn();
+            }
+
+            $this->_resetHandles();
+            try {
+                $this->_httpResponse = $this->_parent->client->api($this->_buildHttpRequest());
```
SteveJobzniak: This and some other places use `$this->_parent->client->...`. I can't remember why we coded it that way. Perhaps because we moved code from somewhere else that needed to go to "parent->client", and missed the rewrite when placing them in Client.php. My gut question right now is "Isn't `$this->_parent->client` the same as `$this`?".

src/Request.php
```diff
@@ -259,39 +483,107 @@ protected function _getRequestBody()
...
+    /**
+     * Perform the request and map its response data to provided object.
+     *
+     * @param ResponseInterface $baseClass An instance of a class object whose properties to fill with the response.
+     *
+     * @throws \InstagramAPI\Exception\InstagramException
+     *
+     * @return ResponseInterface An instance of baseClass.
+     */
+    public function getResponse(
```
SteveJobzniak: Nice. I like the getResponse for ResponseInterface object and getRawResponse for raw object/array. Very nice idea!

SteveJobzniak:
This is fantastic. I was confused at first but now I see that you've made Client into basically a Guzzle/server communicator, and made Request into the request-builder.

There are just a few minor questions and comments above, and perhaps some tiny tweaks. It all looks great. I have not tested it but I know how thorough you are!

src/Client.php
valga: Actually, we had no public method to set Guzzle options per request, because we used removed per-request options only for storing headers and body.

As we take full HTTP request object as param, we no longer need to use per-request options.

src/Client.php
valga: This change was made by accident with moving this code back in forth, will revert it. Although PSR-2 says:
> The case statement MUST be indented once from switch

src/Request.php
```diff
@@ -259,39 +483,107 @@ protected function _getRequestBody()
...
-                .(strpos($this->_url, '?') === false ? '?' : '&')
+            $endpoint = $endpoint
+                .(strpos($endpoint, '?') === false ? '?' : '&')
                 .http_build_query(Utils::reorderByHashCode($this->_params));
-        } else {
-            $endpoint = $this->_url;
         }
-
+        // Prepare headers.
+        if (!count($this->_headers)) {
```
SteveJobzniak: This seems dangerous. Means that anytime a header is added, all default headers will be ignored unless we specifically remember to call addDefaultHeaders, like you did manually in the timeline feed endpoint.

src/Request/Internal.php
```diff
@@ -1058,4 +1061,248 @@ public function sendClientEventLogs(
...
+                    }
+
+                    // Prepare request.
+                    $request = new Request($this->ig, $uploadParams['uploadUrl']);
+                    $request
+                        ->addHeader('Content-Type', 'application/octet-stream')
+                        ->addHeader('Session-ID', $uploadParams['uploadId'])
+                        ->addHeader('Content-Disposition', 'attachment; filename="video.mov"')
+                        ->addHeader('Content-Range', 'bytes '.$chunk['rangeStart'].'-'.$chunk['rangeEnd'].'/'.$videoSize)
+                        ->addHeader('job', $uploadParams['job'])
```
SteveJobzniak: This is missing many headers of the original code. Is that okay? I guess the app has changed and we no longer need them.

We used to have all of these:

```php

--
-                    $headers = [
  |   | -                        'User-Agent'          => $this->_userAgent,
  |   | -                        'Connection'          => 'keep-alive',
  |   | -                        'Accept'              => '*/*',
  |   | -                        'Cookie2'             => '$Version=1',
  |   | -                        'Accept-Encoding'     => 'gzip, deflate',
  |   | -                        'Content-Type'        => 'application/octet-stream',
  |   | -                        'Session-ID'          => $uploadParams['uploadId'],
  |   | -                        'Accept-Language'     => Constants::ACCEPT_LANGUAGE,
  |   | -                        'Content-Disposition' => "attachment; filename=\"video.{$videoExt}\"",
  |   | -                        'Content-Range'       => 'bytes '.$chunk['rangeStart'].'-'.$chunk['rangeEnd'].'/'.$videoSize,
  |   | -                        'job'                 => $uploadParams['job'],
  |   | -                    ];


```

src/Request/Internal.php
```diff
@@ -1058,4 +1061,248 @@ public function sendClientEventLogs(
...
+                "Upload of \"%s\" failed. Instagram's server returned an unexpected reply and is probably overloaded.",
+                $videoFilename
+            ));
+        }
+
+        // Manually decode the final API response and check for successful chunked upload.
+        /** @var Response\UploadVideoResponse $upload */
+        $upload = $this->ig->client->getMappedResponseObject(
+            new Response\UploadVideoResponse(),
+            Client::api_body_decode((string) $httpResponse->getBody()), // Important: Special JSON decoder.
```
SteveJobzniak: This code is a bug. The old code used `$response['body']` which was a cache of the getBody() contents. As I explained earlier, Guzzle's getBody() is a stream which is only readable once, after which it is an empty string on subsequent calls.

In this new code you are doing httpResponse->getBody() twice.

src/Client.php
SteveJobzniak: Aha okay. So you basically kept the parameter around so we can re-add the feature in buildGuzzleOptions later? Right now it looks a bit weird to have the unused param. What's your thought on that?

src/Request.php
valga: > Guzzle only lets you read the Body stream ONCE.

I can't test it right now, but it is not 100% correct. Guzzle return body as `Stream` object, and `Stream` has 2 methods to get its contents: `getContents()` reads the remaining buffer, and `__toString()` seeks stream to the beginning and calls `getContents()`. So, basically, you can rewind the stream and read it again. 

And I used `(string)` typecase on response body everywhere.

src/Client.php
SteveJobzniak: Ah okay. We don't follow PSR coding standards though (which is generally good but has some bad aspects too). We only follow their PSR-4 autoloading (1 class per file, namespace hierarchy folders).

src/Request.php
valga: Yeah, I don't like either. Do you have any thoughts on this? For you info, we can't add default headers to every request, because video uploaders don't use them.

src/Request/Internal.php
valga: `Client` sets `User-Agent`, `Connection`, `Accept` and `Accept-Encoding` headers on every request, because they depend on client's implementation. So, I removed only one header: `Cookie2`.

But after double-checking I found that I lost `Accept-Language` header while deciding where to put it. Will fix it.

src/Request/Internal.php
valga: As I explained above, `Stream` is seekable, so it is not a bug. Actually, I tested video uploads, and they worked fine. Anyway, will double check it.

src/Client.php
valga: > What's your thought on that?

I don't know. We can either remove unused param, or add public method to set per-request options. Or ignore it and leave it for future use.

src/Client.php
valga: Mine IDE auto-formats code on copy-paste, and I totally missed indent change, because I follow PSR-2 in general.

src/Request.php
SteveJobzniak: Ohh, that's it. I was using getBody-getContents which means the stream was never rewound. You can use string casting as much as you want: https://stackoverflow.com/questions/30549226/guzzlehttp-how-get-the-body-of-a-response-from-guzzle-6

src/Request.php
SteveJobzniak: Hmm, how about a function called `setAddDefaultHeaders(false)` (default to true)?

src/Request/Internal.php
SteveJobzniak: Ah so Cookie2 is gone from the app?

And nice, good catch about Accept-Language!

src/Request/Internal.php
SteveJobzniak: You were right. String-casting the body can be done repeatedly.

src/Client.php
```diff
@@ -712,431 +684,92 @@ protected function _guzzleRequest(
...
-     *                                           otherwise we decode to standard object.
-     *                                           Note that this parameter should be FALSE
-     *                                           if you are providing the baseClass param!
-     * @param ResponseInterface|null $baseClass  An instance of a class object whose
-     *                                           properties to fill with the response,
-     *                                           or NULL to get a standard object/array.
+     * @param HttpRequestInterface   $request   HTTP request to send.
+     * @param ResponseInterface|null $baseClass An instance of a class object whose
+     *                                          properties to fill with the response,
+     *                                          or NULL to get a standard object/array.
```
SteveJobzniak: This is wrong. Baseclass param is not taken.

src/Client.php
SteveJobzniak: Perhaps change the public api() call-chain to take both the $request and $guzzleOptions, and give Request.php a "setGuzzleOptions(array)" public function?

Speaking of which, just a code tip: I usually do "array $foo = null" as default parameter, because PHP before 7.0 sucks at empty arrays. It allocates a lot of memory buckets assuming the array is going to used. In PHP7 they never pre-allocate buckets for empty arrays anymore.

src/Request.php
```diff
@@ -259,39 +483,107 @@ protected function _getRequestBody()
...
+            $this->addDefaultHeaders();
+        }
         /** @var string|null $postData The POST body contents; is NULL if GET request instead. */
```
SteveJobzniak: This is now a StreamInterface|null. ;)

SteveJobzniak:
Okay I have now done a 3rd readthrough. It all looks very good apart from the few things mentioned in the review. Awesome. :)

valga:
Ok, I believe I fixed everything except for that Guzzle's per-request options. I leave it up to you.

And don't forget to squash this PR.

SteveJobzniak:
@valga Thanks a lot, everything looks great now! <3 I'll add the per-request guzzle options feature next after merging.

SteveJobzniak:
This was a really great cleanup by the way. Loving these changes!

SteveJobzniak:
@valga Okay the options feature is done:

https://github.com/mgp25/Instagram-API/commit/92a1cae2ae23cb0fb8d1b3f943427bd6e0a3861b

You can now use `->setGuzzleOptions(['debug' => true])` next time you're doing raw HTTP stuff. :)

PR #1391, opened 2017-07-09
Refactor video uploads

valga:
As we (or I) are going to implement a new video uploader, we need to have just one entry point for all video uploads. So this is the main goal of this PR.

But I also made a couple of changes:
1. I got rid of `$maxAttempts` params, because they are kind of pointless (and dangerous because of that). We use consts now.
2. I rewrote existing uploader to work like app: it changes chunk size based on network/Instagram speed, it tries all provided upload URLs in case of errors, it handles errors properly.
3. Internal metadata is now an object.


src/Request/Direct.php
```diff
@@ -636,16 +614,16 @@ public function sendVideo(
...
         // Send the uploaded video to recipients.
         $result = null;
-        for ($attempt = 1; $attempt <= $maxAttempts; ++$attempt) {
+        for ($attempt = 1; $attempt <= Internal::MAX_CONFIGURE_RETRIES; ++$attempt) {
             try {
                 // Attempt to configure video parameters (which sends it to the thread).
                 $result = $this->_sendDirectItem('video', $recipients, array_merge($options, [
                     'upload_id'    => $internalMetadata['uploadId'],
-                    'video_result' => $upload->getResult(),
+                    'video_result' => $internalMetadata['videoResult'],
```
SteveJobzniak: Is this key right? It says uploadResult in the actual Internal::uploadVideo function.

src/Request/Internal.php
```diff
@@ -1118,192 +1165,151 @@ public function uploadVideoChunks(
...
 
-            // Update the range's Start for the next iteration.
-            // NOTE: It's the End-byte of the previous range, plus one.
-            $rangeStart = $rangeEnd + 1;
-        }
-
-        // Read and upload each individual chunk, doing retries when necessary.
-        $handle = fopen($videoFilename, 'r');
-        $httpResponse = null;
         try {
```
SteveJobzniak: This try should perhaps be moved a bit higher up, very close to when fopen() was called, to protect againt handle leaks if the Stream() constructor fails.

src/Request/Internal.php
```diff
@@ -1118,192 +1165,151 @@ public function uploadVideoChunks(
...
-
-            // Add the current chunk's parameters to the list.
-            $videoChunks[] = [
-                'fileOffset' => $rangeStart, // fseek offsets are 0-indexed too!
-                'chunkSize'  => $chunkSize, // Size (in bytes) of this chunk.
-                'rangeStart' => $rangeStart, // Start offset for the HTTP chunk.
-                'rangeEnd'   => $rangeEnd, // End offset for the HTTP chunk.
-            ];
+        // Init state.
+        $uploadUrl = array_shift($uploadUrls);
```
SteveJobzniak: This seems weird, since the loop below instantly sets uploadUrl to null on each iteration.

SteveJobzniak: Nevermind, I misread. It switches URL if all attempts failed on a certain server. Gotcha.

SteveJobzniak:
Hi valga, I did a quick read before bed. It looks cool so far. Will check more tomorrow.

src/Request/Metadata/Internal.php
```diff
@@ -0,0 +1,231 @@
...
+        $targetFeed,
+        $photoFilename)
+    {
+        // Figure out the photo file details.
+        // NOTE: We do this first, since it validates whether the photo file is
+        // valid and lets us avoid wasting time uploading totally invalid files!
+        $this->_photoDetails = new PhotoDetails($photoFilename, Utils::getPhotoFileDetails($photoFilename));
+
+        // Validate the photo details and throw if Instagram won't allow it.
+        Utils::throwIfIllegalPhotoDetails($targetFeed, $this->_photoDetails);
```
SteveJobzniak: I can see that this has been a big rewrite so unfortunately you've missed the throw-chains that are now happening.

For example, this location is throwing an InvalidArgumentException but setPhotoDetails has no documented throws at all.

Every rewrite must maintain throws-chains so that functions can be safely used without unexpected exceptions.

src/Request/Metadata/MediaDetails.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        return $this->_filename;
+    }
+
+    /**
+     * Constructor.
+     * @param string $filename
+     * @param array  $details
+     */
+    public function __construct($filename, array $details)
```
SteveJobzniak: This and many other functions will be detected if you run `devtools/checkStyle.php`. Good idea to run that periodically to check for this accident.

src/Request/Timeline.php
```diff
@@ -31,7 +32,7 @@ public function uploadPhoto(
         $photoFilename,
         array $externalMetadata = [])
     {
-        return $this->ig->internal->uploadSinglePhoto('timeline', $photoFilename, [], $externalMetadata);
+        return $this->ig->internal->uploadSinglePhoto('timeline', $photoFilename, $externalMetadata);
```
SteveJobzniak: This seems like a regression. Why have you removed the ability to provide internal metadata? It was placed here so that we could do it anytime we need the ability. Because we're not going to put metadata in $externalMetadata.

You could have made that param `InternalMetadata $internalMetadata = null` and make these calls send a null for now.

src/Request/Timeline.php
```diff
@@ -111,54 +111,43 @@ public function uploadAlbum(
...
+                $internalMetadata->setPhotoUploadResponse($this->ig->internal->uploadPhotoData('album', $internalMetadata));
             }
+
+            $media[$key]['internalMetadata'] = $internalMetadata;
         }
 
         // Configure the uploaded album and attach it to our timeline.
-        $internalMetadata = []; // NOTE: NO INTERNAL DATA IS NEEDED HERE YET.
-        $configure = $this->ig->internal->configureTimelineAlbumWithRetries($media, $internalMetadata, $externalMetadata);
+        // NOTE: NO INTERNAL DATA IS NEEDED HERE YET.
```
SteveJobzniak: Keyword: "Yet".

src/Request/Internal.php
SteveJobzniak: Good. Fixed. 👍

SteveJobzniak:
The use of objects for internal metadata is a great change, ensuring safety and consistency. Fantastic!

A few small problems remain in the current state of this pull request:

1. The throw-chains aren't properly documented anymore.
2. Internalmetadata was removed. We are most likely going to need it sometime for something, which is why I ensured we had the parameter (and it is always placed BEFORE externalMetadata in the param order, for consistency; so please re-add it as `($internalMetadata, $externalMetadata)`). Or give some great reason why we shouldn't have it. "We don't need it yet" isn't a reason to remove it. The InternalMetadata parameter is the only way of shuffling arbitrary extra data to internal functions.
3. Some small style issues findable via `devtools/checkStyle.php`.

I've read through a bit and these seem to be the only issues.

src/Request/Timeline.php
valga: I did not remove it. I just changed the order of params: `internalMetadata` is the last one now.

valga:
I found a great reason to use internal metadata in albums. It now locks `client_sidecar_id` to be the same between all configure calls.

SteveJobzniak:
@valga That's a good idea! 👍

SteveJobzniak:
Nice, the new commits seem to address all issues.

 What are the remaining issues and todos on this pull request?

src/Request/Direct.php
```diff
@@ -683,12 +667,12 @@ public function sendVideo(
...
-        array $externalMetadata = [],
-        $maxAttempts = 10)
+        array $externalMetadata = [])
     {
-        $internalMetadata = $this->_addRecipientsToMetadata($recipients, []);
+        $internalMetadata = new InternalMetadata();
+        $internalMetadata->setDirectRecipients($this->_prepareRecipients($recipients, true));
 
-        return $this->ig->internal->uploadSingleVideo('direct_story', $videoFilename, $internalMetadata, $externalMetadata, $maxAttempts);
+        return $this->ig->internal->uploadSingleVideo('direct_story', $videoFilename, $externalMetadata, $internalMetadata);
```
SteveJobzniak: The issue with param order remains here.

src/Request/Internal.php
```diff
@@ -278,16 +272,51 @@ public function configureSinglePhoto(
...
+        }
+        if ($internalMetadata->getVideoDetails() === null) {
+            $internalMetadata->setVideoDetails($targetFeed, $videoFilename);
+        }
+
+        // Request parameters for uploading a new video.
+        $internalMetadata->setVideoUploadUrls($this->_requestVideoUploadURL($targetFeed, $internalMetadata));
+
+        // Attempt to upload the video data.
+        $internalMetadata->setVideoUploadResponse($this->ig->internal->_uploadVideoChunks($targetFeed, $internalMetadata));
```
SteveJobzniak: This callpath is weird. We are already inside Internal.php, no need to call $this->ig->internal. ;) There may be others like this, I only noticed this one.

src/Request/Internal.php
```diff
@@ -278,16 +272,51 @@ public function configureSinglePhoto(
...
-     *                                 but NOT "album", they are handled elsewhere).
-     * @param string $videoFilename    The video filename.
-     * @param array  $internalMetadata (optional) Internal library-generated metadata key-value pairs.
-     * @param array  $externalMetadata (optional) User-provided metadata key-value pairs.
-     * @param int    $maxAttempts      (optional) Total attempts to upload all chunks before throwing.
+     * @param string                $targetFeed       Target feed for this media ("timeline", "story", "direct_story",
+     *                                                but NOT "album", they are handled elsewhere).
+     * @param string                $videoFilename    The video filename.
+     * @param array                 $externalMetadata (optional) User-provided metadata key-value pairs.
+     * @param InternalMetadata|null $internalMetadata (optional) Internal library-generated metadata key-value pairs.
```
SteveJobzniak: Order. :)

src/Request/Internal.php
```diff
@@ -344,22 +350,20 @@ public function uploadSingleVideo(
     /**
      * Asks Instagram for parameters for uploading a new video.
      *
-     * @param string $targetFeed       Target feed for this media ("timeline", "story", "album" or "direct_v2").
-     * @param array  $internalMetadata (optional) Internal library-generated metadata key-value pairs.
+     * @param string           $targetFeed       Target feed for this media ("timeline", "story", "album" or "direct_v2").
+     * @param InternalMetadata $internalMetadata (optional) Internal library-generated metadata key-value pairs.
```
SteveJobzniak: Seems this is no longer optional so (optional) cna be removed.

src/Request/Internal.php
```diff
@@ -1071,32 +1081,25 @@ public function sendClientEventLogs(
...
-        if ($maxAttempts < 1) {
-            throw new \InvalidArgumentException('The maxAttempts parameter must be 1 or higher.');
-        }
-
-        // Verify that the file exists locally.
-        if (!is_file($videoFilename)) {
-            throw new \InvalidArgumentException(sprintf(
-                'The video file "%s" does not exist on disk.',
+        $videoFilename = $internalMetadata->getVideoDetails()->getFilename();
+        // Video uploads should be chunked to save RAM; determine chunk size!
```
SteveJobzniak: This comment seems incorrect now. :) The code for determining chunksize was removed.

src/Utils.php
```diff
@@ -350,53 +405,78 @@ public static function throwIfIllegalMediaResolution(
...
+     *
+     * @throws \InvalidArgumentException If Instagram won't allow this photo.
+     */
+    public static function throwIfIllegalPhotoDetails(
+        $targetFeed,
+        PhotoDetails $photoDetails)
+    {
+        $photoFilename = $photoDetails->getFilename();
+        // Validate image type.
+        // NOTE: Instagram has no disk size limit, but this length validation
```
SteveJobzniak: Hehe this NOTE is about video length validation. Remove it from here (photos) ;)

valga:
I have a couple of things related to uploads, but I will address them later in new PRs.

I have only one thing that bothers me: we have only 2 attempts to upload video chunk. App does the same, but it can retry upload cycles like forever. But as we handle more than one server now, I guess it should be fine. Or not...

SteveJobzniak:
Perfect. I have read through the whole pull request again and can't see any more errors. And you have fixed every note above.

As for the retries: I guess the app cycles between the servers forever until the user aborts. Hmm. Well, we have 3 servers and 2 retries per chunk per server. This will be fine until the new video upload method. Perhaps raise it to 3 retries per chunk, though?

valga:
We have 4 servers. And I can't really say whether 3 retries per chunk are ok or not.

SteveJobzniak:
I think 2 is just very low since I remember testing and needing 5 retries to succeed with a chunk when a server was overloaded. So what is preferable? Try more times on the same server or let it fail and switch to a different server?

Btw: https://github.com/mgp25/Instagram-API/blob/master/examples/uploadVideo.php this file has a comment about raising maxAttempts. ^_^

valga:
> I remember testing and needing 5 retries to succeed with a chunk when a server was overloaded.

Was it "exactly 4 chunks per file" time?

Anyway, I changed per-chunk retries from 2 to 5.

SteveJobzniak:
I can fully agree with bumping it to 5. Worst case scenario if all 4 servers fail a chunk 5 times, we are only using 20 maxAttempts. Most uploads only fail a little bit. So this is good.

And no, it was not always 4 retries. Sometimes 1, sometimes 8, but usually in the area of 2-3 failures before success. So 5 is a good number which we should be okay with.

SteveJobzniak:
I misunderstood the question. Yes, that was in the days when we used 4 chunks per file. :)

SteveJobzniak:
I forgot to ask if it's ready for merging? It sounded that way but I better make sure. <3

valga:
Yeah, it's ready. And don't forget to squash it.

SteveJobzniak:
Thank you, this has been a beautiful refactoring of chunk uploading and internal metadata! :)

PR #1403, opened 2017-07-13
Resumable video uploader

valga:
In short, it tries to upload whole file at once, and continues from last uploaded offset in case of errors.

Technical note: while we upload whole file at once, we don't store large files in memory, because we use Guzzle's streams. Refer to [implementation](https://github.com/guzzle/guzzle/blob/56bf6c1cc3e1b24691f10a09321516d4790cecbd/src/Handler/CurlFactory.php#L252) for more details.


SteveJobzniak:
@valga Brilliant and beautiful as always! And thanks for the link that confirms that Guzzle streams are memory-efficient. Awesome!

I have read theough everything carefully and only have a few small notes:

1. I think we should use a new Internal.php const called something like `MAX_RESUMABLE_RETRIES`, so that we control it separately from the chunk-based uploader. I cna do that change afte rmerging, though, so don't worry about it.
2. Did you manage to test the attempt-failure loop? It treats the server response as Ok if the Offset field exists and is 0+. Which means it never throws any error and retries? Or how are errors detected? Perhaps the offset field is missing, it falls back to _isOk which finds no "status" field and therefore returns false? Is that how it works?

By the way, I discovered to my shock yesterday that we have made a big licensing accident. We made this library GPLv3 to ensure that people who use the code have to contribute back to help improve the library, otherwise we do all this work and get nothing for it (not even code). But it turned out that GPLv3 does not protect web libraries. That is what something called AGPLv3 is for!

We therefore need to fix this by changing all code to a modified version of AGPLv3, with an extra rule that anyone using this library for websites _must_ open source their own site code too. Along with this main AGPLv3 license we'll also have a second license, which exempts people of premium status (which includes very high quality collaborators such as yourself for _life_), so that premiums do not need to open source their sites. This is intended to stop the loophole which people have abused for so long.

I will type this up properly later, in a new LICENSE.md document and in README.md. I hope you are okay with fixing the license to what was originally intended, and that you will have premium status for life (which you highly deserve!).

:-)

mgp25:
@SteveJobzniak i think ill create a link that will show all premium subscribers and put it in the readme, that way everyone could check it, and we dont need to update any license or readme everytime there is a premium subscription.

valga:
> I have read through everything carefully and only have a few small notes:

1. I added a new const for it.
2. I changed `isOk()` method to handle everything by itself without going to "standard" status checker.

As for license matter, I don't think that any license can stop someone from illegal activities. Also, I don't like that legal stuff, but strictly speaking I hold copyright only on `ImageAutoResizer` class, which is not a substantial part of the library. So you can do anything without asking me for approval.

SteveJobzniak:
@valga About the license: Thank you. I finally found a license that was made to ensure code sharing/contributions. It's called the RPL-1.5 and the license has now been applied. It's similar to GPL but is very viral and requires code-sharing. The only exception to it is for premium code contributors/subscribers. Of course, bad companies will just ignore it anyway, but we still wanted to be sure we have the correct license because this library is constant abused. Now that situation has finally been fixed to the extent that we can fix it.

Re 2: Thanks, that looks much clearer and safer! And nice idea about adding an exception message if it's missing. Looks ready to merge! :)

PR #1405, opened 2017-07-14
Use single method for all retry-configurators

valga:
Single method removes code repetition in configurators and also:
1. It handles `media_needs_reupload` error properly. Instagram treats such cases as non-fail (example: `{"message": "media_needs_reupload", "error_title": "staged_position_not_found extra:{}", "status": "ok"}`), so we did not catch it earlier.
2. It uses `cooldown_time_in_seconds` field from response as a configure delay.


src/Client.php
```diff
@@ -509,11 +509,11 @@ protected function _printDebug(
...
-     * @param ResponseInterface          $baseClass      An instance of a class object whose
-     *                                                   properties to fill with the response.
-     * @param mixed                      $serverResponse A decoded JSON response from
-     *                                                   Instagram's server.
-     * @param HttpResponseInterface|null $httpResponse   HTTP response object (if available).
+     * @param ResponseInterface     $baseClass      An instance of a class object whose
+     *                                              properties to fill with the response.
+     * @param mixed                 $serverResponse A decoded JSON response from
+     *                                              Instagram's server.
+     * @param HttpResponseInterface $httpResponse   HTTP response object (if available).
```
SteveJobzniak: Since this is no longer optional, "(if available)" is no longer making sense. :P

src/Request/Direct.php
```diff
@@ -619,29 +619,18 @@ public function sendVideo(
...
-                if ($attempt < Internal::MAX_CONFIGURE_RETRIES && strpos($e->getMessage(), 'Transcode timeout') !== false) {
-                    // Do nothing, since we'll be retrying the failed configure...
-                    sleep(1); // Just wait a little before the next retry.
-                } else {
-                    // Re-throw all unhandled exceptions.
-                    throw $e;
-                }
             }
-        }
-        if ($result === null) { // Safeguard since _sendDirectItem() may return null in some cases.
```
SteveJobzniak: You said that result could be null from the function. I guess not? It actually didn't look like null was possible last time I checked.

src/Request/Internal.php
```diff
@@ -1073,6 +993,76 @@ public function sendClientEventLogs(
...
+            } catch (ThrottledException $e) {
+                throw $e;
+            } catch (LoginRequiredException $e) {
+                throw $e;
+            } catch (FeedbackRequiredException $e) {
+                throw $e;
+            } catch (CheckpointRequiredException $e) {
+                throw $e;
+            } catch (\Exception $e) {
+                // Ignore everything else.
```
SteveJobzniak: This is the first time our "Ignore everything else." comment has something else after it (`continue;`)... So it should say something like `Ignore everything else, and retry`.

src/Request/Internal.php
```diff
@@ -1073,6 +993,76 @@ public function sendClientEventLogs(
...
+                continue;
+            }
+
+            $httpResponse = $result->getHttpResponse();
+            $fullResponse = $result->getFullResponse();
+            switch ($httpResponse->getStatusCode()) {
+                case 200:
+                    if (strtolower($result->getMessage()) === 'media_needs_reupload') {
+                        throw new \InstagramAPI\Exception\UploadFailedException(sprintf(
+                            'Configuration of "%s" failed: you need to reupload the media (%s).',
```
SteveJobzniak: This is ultra-minor, but stylistically a capital letter looks nicer, perhaps? `'Configuration of "%s" failed: You need to reupload the media (%s).'`

src/Request/Internal.php
```diff
@@ -1073,6 +993,76 @@ public function sendClientEventLogs(
...
+                            $entity,
+                            (isset($fullResponse->error_title) ? $fullResponse->error_title : 'unknown error')
+                        ));
+                    }
+
+                    return $result;
+                case 202:
+                    sleep(isset($fullResponse->cooldown_time_in_seconds) ? $fullResponse->cooldown_time_in_seconds : 1);
```
SteveJobzniak: I guess this is fine (reading it from the raw fullresponse object) code-wise but doesn't this mean that a field should be added to the ConfigureResponse class?

src/ResponseInterface.php
```diff
@@ -111,4 +113,19 @@ public function getFullResponse();
...
+     */
+    public function setHttpResponse(
+        HttpResponseInterface $response);
+
+    /**
+     * Gets the HTTP response.
+     *
+     * @return HttpResponseInterface
+     */
+    public function getHttpResponse();
```
SteveJobzniak: To match AutoResponsePropertySetter (and not letting it auto-handle missing properties and become documented as isHttpResponse() in every class by the auto-documenter) we need to have explicit `isHttpResponse()` too. Check the other funcs in ResponseInterface and ResponseTrait for the format of that func.

src/ResponseInterface.php
```diff
@@ -111,4 +113,19 @@ public function getFullResponse();
      * @return bool
      */
     public function isFullResponse();
+
+    /**
+     * Sets the HTTP response.
+     *
+     * @param mixed $response
```
SteveJobzniak: It's not mixed. It's `@param HttpResponseInterface $response`, here and in the Trait too. :)

src/Request/Internal.php
SteveJobzniak: By the way, the reason we ignored their "cooldown time" is because they never tell the truth.

They say "4.5 seconds" (at least when I worked on this), but a configure is possible within maybe 0.8 seconds or so. So by ignoring their time we were faster. Not sure if that is still the case.

valga: App no longer uses `configure_delay_ms`. In fact, new resumable uploader doesn't even send it.

src/ResponseInterface.php
valga: Fixed.

src/Request/Direct.php
valga: I don't remember saying that thing about null. That was a catch-all safeguard to ensure that we have a valid result object. Just in case that someone has failed in main loop, because it was not so clear.

src/Request/Internal.php
valga: As for taking `cooldown_time_in_seconds` from full response, this field is used by app for failed responses (along with `feedback_message`, `checkpoint_url` and so on). So we can't place it into `ConfigureResponse` for sure.

src/Request/Direct.php
SteveJobzniak: https://github.com/mgp25/Instagram-API/pull/1273#issuecomment-304678944

Quote: `this method relies on _sendDirectItem() which _may_ return null because of someone's fault.`

Hehe that's where you said it. But I never understood it then either because the func only returns a value or throws. Never returns null.

src/Request/Internal.php
SteveJobzniak: Aha okay, so it's only there on failure from that endpoint... I see. Then it's fine.

Just one more question/thought: Should we be using "InstagramException $e; if $e->hasResponse(): $result = $e->getResponse()" in the try-catch chains earlier? Otherwise, if status:fail, we'll never get any $result value and never process it?

SteveJobzniak:
Okay so I re-checked and all is great, just one question before merging:

Should we be using "InstagramException $e; if $e->hasResponse(): $result = $e->getResponse()" in the try-catch chains in configureWithRetries()? That way we can extract the whole response/result object even when the try{$result=call_user_func()} fails. Otherwise, if status:fail, the call_user_func will throw and we'll never get any $result value and never process this cooldown value? Hmm?

SteveJobzniak:
Okay so I re-checked and all is great, just one question before merging:

Should we be using "InstagramException $e; if $e->hasResponse(): $result = $e->getResponse()" in the try-catch chains in configureWithRetries()? That way we can extract the whole response/result object even when the try{$result=call_user_func()} fails. Otherwise, if status:fail, the call_user_func will throw and will go immediately to "catch Exception $e: continue" and we'll never get any $result value and never process this cooldown value and just instantly loop back and retry again with no delay? Hmm?

valga:
Hmm, I was not able to catch a response with `cooldown_time_in_seconds`, and I can't say whether it is `ok` or `fail`. So I added a check for failed responses.

SteveJobzniak:
@valga Ahhhh, I see. Well, the new code is the totally correct way of getting the response even when an exception happens, and you're doing the isOk check manually, so everything looks safe this way!

By the way, I was just about to merge but realized there's still no sleep if status is 200 but response is not marked Ok. So should we manually add a sleep(1) here?

```
                    } elseif (!$result->isOk()) {
                        // Continue to the next attempt.
                        sleep(1); // <-- this may be a good idea?
                        break;
```

valga:
I rewrote configurator to sleep at least 1 sec between retries.

SteveJobzniak:
Excellent, every code path looks safe and reliable. Thanks a lot. Time to merge! :)

PR #1406, opened 2017-07-14
Various small fixes

valga:
See commits for changelog.


SteveJobzniak:
Thanks, these are all great discoveries and logical fixes. I'll take a break to eat and then investigate the `Storage/File` change. That's the only one that isn't totally obvious.

valga:
> That's the only one that isn't totally obvious.

You used that property in [exception message](https://github.com/valga/Instagram-API/blob/e660819a141d19aa18c421e42d89e98d4694a15d/src/Settings/Storage/File.php#L305) without defining/assigning it.

SteveJobzniak:
Ahhh, now I remember. Because the File storage class concerns disk-paths to Cookies and Settings files, it had no reason to keep the _username cached. But since we're using it in an exception message, we of course need it. Thanks for that discovery! I'm going to the lunch break now and will check and merge after that. :) And also taking a look at the other pull requests again. They'll all be merged during the day. :)

SteveJobzniak:
Thank you, everything is great 👍  :)

PR #1408, opened 2017-07-14
ImageAutoResizer: Use dynamic output quality

valga:
I don't expect this to be merged, because the reason "app does that" is not that strong, but it is worth discussing at least.

Nevertheless, app really does that.

### DIFF
--- a/src/ImageAutoResizer.php
+++ b/src/ImageAutoResizer.php
@@ -56,6 +56,15 @@ class ImageAutoResizer
     const MAX_WIDTH = 1080;
 
     /**
+     * Minimum allowed image width.
+     *
+     * Used only for quality calculation.
+     *
+     * @var int
+     */
+    const MIN_WIDTH = 200;
+
+    /**
      * Maximum allowed image height.
      *
      * This is derived from 1080 / 0.8 (tallest portrait aspect allowed).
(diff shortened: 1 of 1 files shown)


SteveJobzniak:
Haha, this is cute - but why do it? :)

Read page 5 of the JPEG spec: https://www.w3.org/Graphics/JPEG/jfif3.pdf

The headers of a JPEG contain:

- length
- identifier ("JFIF")
- version
- units (for pixel density), such as if it's in pixels or DPI etc
- Xdensity and Ydensity: how many units horizontally and vertically
- Xthumbnail and Ythumbnail: unit size of thumbnail
- thumbnail pixels
- data payload for the rest of the image

There's nothing in a JPEG saying what quality it was saved with. So we don't need to reduce quality. Why would we want to do that? We'd just lose sharpness.

SteveJobzniak:
Furthermore, I just found out that "JPEG Quality" is implementation-specific. There is no common definition for it. So what our PHP JPEG creator creates at quality 95 is not the same as what the app creates at quality 95, for example:

> Jpeg compression algorithm has some parameters which influence on the quality of the result image.
> 
> One of such parameters are quantization tables which defines how many bits will be used on each coefficient. Different programs use different quatization tables.
> 
> Some programs allow user to set quality level 0-100. But there is no common defenition of this number. The image made with Photoshop with 60% quality takes 46 KB, while the image made with GIMP takes only 26 KB.
> 
> Quantization tables are also different.
> 
> There are other parameters such subsampling, dct method and etc.
> 
> So you can't describe all of them by single quality level number and you can't compare quality of jpeg images by single number. But you can create such number like photoshop or gimp which will describe compromiss between size on quality.
> 
> More information: http://patrakov.blogspot.com/2008/12/jpeg-quality-is-meaningless-number.html
> 
> Common practice is that you resize the image to appropriate size and apply jpeg after that. In this case huge and middle images will have the same size and quality.

SteveJobzniak:
In summary, if you look at that page: The JPEG quality number is totally meaningless and Instagram can't even read it from the file. So definitely skipping this. ;-P

PR #1409, opened 2017-07-14
ImageAutoResizer: ability to expand the image with white margins instead of cropping

safhome:
This patch adds ability to expand the image with white margins (top/bottom or left/right) instead of cropping while adjust an aspect ratio.
Sometimes cropping an image is unacceptable, because it can contain any graph or other information that the user wants to convey to the recipient as a whole. In this case, the more successful solution is to add white fields to the image - in the interface of the Instagram they'll merge with the background and will look quite acceptable.


SteveJobzniak:
Thanks, this is worth thinking about when the other pull requests are done. 👍  Saving this (open) for later.

SteveJobzniak:
Okay so all other pull requests are finally reviewed and taken care of. But I'm exhausted after a long series of recent events, and I critically need to take care of my health treatment.

This pull request contains lots of code changes, algorithm changes, and conflicts with the current state of the file. It's too much for me to read through. And it needs to be updated to match the latest sources.

Furthermore, from what I saw at a quick review, it does things like `imagecopyresampled($output, $source, $src_x / $scale, $src_y / $scale` which seems like it would produce floating point coordinates.

In general, I would support the idea of adding 'expand'-support, but since I don't have any energy left to review the code, there's only two options: 1. Hmm, @valga, can you review it (or implement this idea cleanly in a way better fitting your original class structure, and perhaps with support for choosing border color (white or black)). Alternatively, 2. We'll have to decline this. Because I literally have zero energy left to think about this right now.

safhome:
I've updated my PR to conform actual master.

I've used already defined color for borders instead of define additional parameter because:
1) one PR - one feature;
2) white color has been already used in ImageAutoResizer for background, so change this - additional feature;
3) so many parameters have to go to the transfer of parameters through the array - another additional feature;
4) in a real life white borders cover 95% of the needs, if not 100%.

valga:
I reviewed it and here is a list of suggestions:
1. I can't imagine yet another operation other than crop and extend, so we can just use bool flag for operation.2
2. `$x2` and `$y2` variables are useless, so you can safely remove them and use `$width` and `$height` instead.
3. `$x1` and `$y1` should be renamed to `$shiftX` and `$shiftY`. When shift is positive, it need to be applied to src image, otherwise it will define a position on dst image.
4. There is a lot of copy-paste in calculations, so you should refactor them.
5. Do not use `$scale` param. I might get wrong, but all you need is to apply it on `$shift` variables with `round()`. Because we already have old and new dimensions ready.

valga:
> one PR - one feature

There is no such rule. You can send multiple features in one PR if they are related to each other. For example, background color is related to image extending feature, and both of them are related to refactoring of `ImageResizer` class to accept its options as array instead of dozens of params. And all these 3 features are in just one class, so you kind of _must_ send them in one PR.

safhome:
Ok, I get it, and I'll work on it.

SteveJobzniak:
The merging wasn't proper either. The comments are in the wrong places and you didn't match the new code style.

I'll reject this one. It's too sloppy and messes up the class.

SteveJobzniak:
The idea is worth discussing further, but done properly. Here are some ideas @valga. I may even do them but not right now... or perhaps you'd be up for it?

1. Constants: `const CROP = 1; const EXPAND = 2;`
2. Constructor: `$this->_operation = $operation === self::EXPAND ? self::EXPAND : self::CROP`
3. Change constructor to take `array $options` instead.
4. Add some `$_expandBgColor = ...;` property, which defaults to white but can be set via constructor `$options` array.
5. As for the algorithm: It would simply need to calculate the target (expanded) dimensions. Nothing else (shift/offset/etc) matters when doing expansion. The final "imagecopyresampled" can simply do the job of figuring out the centered position (and scale) within the new canvas, with a single algorithm. Things like `$cropFocus` are not needed when expanding.

SteveJobzniak:
The idea is worth discussing further, but done properly. Here are some ideas @valga. I may even do them, but not right now because I still need to rest... or perhaps you'd be up for it? If so, please let me know so we don't do the same work in parallel. Otherwise I'll keep this as a to-do for another day.

1. Change constructor to just take `($inputFile, array $options)` instead of lots of params.
2. Constants (int): `const CROP = 1; const EXPAND = 2;`
3. Constructor param: `$this->_operation = isset($options['operation'] && $options['operation'] === self::EXPAND ? self::EXPAND : self::CROP` 
4. Add a `protected $_bgColor;` property, which defaults to white but can be set to something else via constructor `$options['bgColor']` array. This value is used in both expansion, and when scaling/converting a transparent PNG. Hence the generic "bgColor" property name (since it's not just for expansion).
6. As for the algorithm: It would simply need to calculate the target (expanded) dimensions (and of course ensuring that the target canvas width/size fits within the max width Instagram allows). None of the other variables (shift/offset/cropfocus/x/y/etc) matters _at all_ when doing expansion. The final "imagecopyresampled" function can simply do the job of figuring out the centered position (and scale) of the original image within the new, larger canvas, with a single, simple algorithm which auto-determines the scaling (something like `min(targetWidth/originalWidth, targetHeight/originalHeight)` which would give the necessary downscaling change), and then calculates the downscaled dimensions of the original image (we probably need to use `floor()` for that), and then calculates the centered offset of the downscaled original image's dimensions vs the new canvas (via `(targetHeight-downscaledHeight)/2` and `(targetWidth-downscaledWidth)/2`). And features like `$cropFocus` makes no sense when expanding, so all "focusing"-related code can be ignored. In other words, the algorithm for expansion which I just described is _very_ simple.

SteveJobzniak:
Heck, since I've just done the planning work, there's not much left to do. I'll get started on coding this today.

SteveJobzniak:
@valga Alright, the algorithm is complete! It's very clean and efficient. Read the commit messages. The only thing I thought about doing, but didn't do, was to move the cropping-algorithm's x/y calculations to the exact same code location as the expand-algorithm. But there's no real point... They are just some extra vars that are very quickly calculated and then simply ignored by the expand algorithm, so we don't really have to move them, and it would possibly just complicate things if we tried moving them. So I'm happy with this final code. It is clean and modular, as intended. And doesn't break the class (unlike the original PR ;-)).

ImageAutoResizer: Use array for constructor params
https://github.com/mgp25/Instagram-API/commit/a1f85ef91f9477e1c34cc6d9249f702575b01e64

ImageAutoResizer: Ability to change bgcolor
https://github.com/mgp25/Instagram-API/commit/b934ce1e21d1778bd0cd4251197d362c0ab7a3c0

ImageAutoResizer: Add new "operation" parameter
https://github.com/mgp25/Instagram-API/commit/9be94a68f2ece17635f320ea62b8435cb04be89b

ImageAutoResizer: Expand-algorithm implementation
https://github.com/mgp25/Instagram-API/commit/a1dc9a9f6899aa1fc8d062f006d73b730b6bf42d

@safhome Enjoy! :-) http://i.imgur.com/Nk471sf.jpg The final commit shows how to use the expand-mode. Have fun!

Triden:
Sorry with the card does not work translate to you, for some reason.)

SteveJobzniak:
@Triden Hehehe, it's easy with this link: https://www.paypal.me/Armindale/0usd, but perhaps PayPal is blocked for you. :-O

Issue #1410, opened 2017-07-14
Backup Bug


rudemex:
Hi guys, how are you?

I was testing the backup () function and it returns an error when it is video.
The console returns this to me.

`PHP Fatal error:  Call to a member function getUrl() on null in /Users/mex/..../deploy/vendor/mgp25/instagram-php/src/Instagram.php on line 2901`

If I find the fix to the bug, I'll let them know.

Regards!

rudemex:
Ok, I found what the problem, in the function only define 2 types of files, photos or video, but they are not defining if it is carrousel.

I'm going to dig a little to see how to solve it.

Regards!

mgp25:
@rudemex you are using an outdated code

rudemex:
@mgp25 That strange, I made a composer update, delete the dependency of the json, and I reinstalled, and igue giving me the same error, this you tell me, is in l version dev? Or in the stable?


Thanks

mgp25:
**README:** `composer require mgp25/instagram-php dev-master`

rudemex:
sorry, but not work :/

Change the stable version 2.0 to the dev-master, but it returns an error

PHP Fatal error:  Cannot use isset() on the result of an expression (you can use "null !== expression" instead) in /Users/mex/.../deploy/vendor/mgp25/instagram-php/src/Request.php on line 166

Look for what that line says, and it's because of the set-up constant of the version. If I force a version, I get the login, but not the backup of my account.

Regrads!

SteveJobzniak:
@rudemex I remember this kind of error. It means you are using PHP 5.5 (or was it 5.4?) or older. They did not support isset() and empty() etc on expressions.

This library requires PHP 5.6 or newer. Even if you are running "composer require ..." as a PHP 5.6+ process, your actual PHP web process is outdated.

This is an issue with your web host configuration. Invalid.

SteveJobzniak:
@rudemex Turns out this was not related to the old `empty()` bug. A recent commit added an `isset()` around a constant which is apparently only supported in PHP7+. It has been fixed now in #1411.

rudemex:
hi @SteveJobzniak 

I have php 5.6, I did a test on the update they did on #1411 , I managed to login, but it does not seem to work all the rest, upload image, get info from a user, etc.

SteveJobzniak:
[...]

PR #1411, opened 2017-07-15
Do not use isset() on consts

valga:
It fixes a bug introduced in #1390: although PHP7 allows it, we must use `array_key_exists()` on const arrays.

Thanks to @rudemex for pointing it in #1410.

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -163,7 +163,7 @@ class Request
     public function setVersion(
         $apiVersion)
     {
-        if (!isset(Constants::API_URLS[$apiVersion])) {
+        if (!array_key_exists($apiVersion, Constants::API_URLS)) {
             throw new \InvalidArgumentException(sprintf('"%d" is not a supported API version.', $apiVersion));
         }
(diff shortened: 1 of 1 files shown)


mgp25:
@valga Nice! Seems everything correct and it has compatibility with previous php versions (PHP 4 >= 4.0.7, PHP 5, PHP 7).

Since the array is a contant there is no problem using it, it has fixed values (won't be null). Tested and working fine.

Thanks again @valga 

References:
- http://php.net/manual/es/function.array-key-exists.php
- http://php.net/manual/es/function.isset.php
- https://stackoverflow.com/questions/3210935/difference-between-isset-and-array-key-exists

SteveJobzniak:
Hmm, ohh, so that thing was introduced in the recent refactoring.

I thought it was the same as the earlier issue where `empty(expr)` only works in PHP5.6+.

I see now that PHP 5.6 does not support `isset(expr)`. Verified with:

```
<?php
const FOO=['a'=>1];
var_dump(isset(FOO['a']));
```

Issue #1422, opened 2017-07-19
Getting locked out for spam, but it's the VERY first action.


GarlandM:
Hi,

It seems that I'm getting the dreaded "Sorry, this feature isn't available right now" issue. 

However, I can assure you, I am not spamming. This happens almost immediately, upon my second or third like, the account will be like blocked via this API.

However, I'm still able to manually like via an actual device, such as my galaxy or iPhone.

Thoughts?

mgp25:
@GarlandM If you are not spamming, you are using a flagged connection then, like a proxy or similar.

Not API issue, dont use flagged networks

Regards

GarlandM:
@mgp25 

I can 100% guarantee you this is NOT a flagged proxy or connection. As stated in my original message, if I use the same account over wifi (thus the same external IP) as testing this script locally (again, same external IP) it works fine.

It's only when I use this script with the account is it blocked.

valga:
@mgp25 apart from some new params to like requests, Instagram API has a new feature: some params can get out from signed body to POST data.

mgp25:
@valga but only in rare ocassions, otherwise its vulnerable to be modified in a mitm. Can you point which ones so i can double check them?

valga:
@mgp25 I checked it more thoroughly and found that it is actually an old feature: they have been using it to skip files from signed bodies. And there are only 3 such params:
1) `profile_pic` file at account creation;
2) `profile_pic` file at account modification;
3) `d` flag at like/unlike.

So you don't have to check anything. As for issue, it seems that Instagram fights with spam and blocks accounts with invalid request params.

mgp25:
@GarlandM post debug log

@valga yup, they use an AI to determine if a user is legit or not, and not sending some params could lead to an account block.

SteveJobzniak:
@valga Will we need to emulate unsigned POST?

valga:
@SteveJobzniak, see recent PR.

@mgp25, I didn't fix params in like methods, I just added the ability to send params in a different way. But it seems that we have kind of broken params, and fixing them could lead to spam, so it's up to you to decide.

SteveJobzniak:
The pull request is now slightly tweaked and merged.

As for whether to fix the functions... hmm... Liking is pretty innocent. "X liked your photo" is gonna happen thousands of times anyway even without spammers alive, so anyone using Instagram usually turns off notifications about photo-likes.

[...]

PR #1427, opened 2017-07-21
Add ability to exclude params from signed body

valga:
As mentioned in #1422, sometimes we need to exclude certain params from signed body and send them as POST vars.

Usage:
```php
$request->setSignedPost(true, ['d'])->addPost('d', '0');
```

Result:
```
signed_body=...&ig_sig_key_version=4&d=0
```


SteveJobzniak:
@valga Excellent! Which syntax do you prefer?

```php
$request
    ->setSignedPost(true, ['d'])
    ->addPost('d', '0');

// or...

$request
    ->addPost('d', '0', false);
```

I think I prefer the first one even though it is more verbose, because it becomes very obvious that something weird is happening via that setSignedPost call.

Although... how about...

```php
$request
    ->addUnsignedPost('d', '0');
```

And addUnsignedPost can be a simple wrapper: `$this->addPost($param, $value); $this->_unsignedPosts[] = $param;`

mgp25:
@SteveJobzniak first one looks clean, i like it

SteveJobzniak:
@mgp25 The 3rd one may be even better though, because the 1st method relies on us typing the lists perfectly and keeping 2 separate lists of param-names in sync.

So the 3rd method is (quoted from above):

```php
$request
    ->addUnsignedPost('d', '0');
```

And addUnsignedPost can be a simple wrapper: `$this->addPost($param, $value); $this->_excludeSigned[] = $param;`

SteveJobzniak:
Okay so I added a commit to this PR which changes the method to:

```php
$request
    ->addUnsignedPost('d', '0');
```

It's easier and less error-prone since we don't need to manually maintain the separate array of unsigned keys this way.

SteveJobzniak:
@mgp25 So apparently, Media::like() and unlike() needs an update with a few Unsigned POST values. We can now use this `$request->addPost->('foo', 'bar')->addUnsignedPost('d', '0');` to solve that.

andrewostroumov:
@valga what is d with value 0 param? Is this need to add to like function as unsigned param?

SteveJobzniak:
@andrewostroumov Nope. Mgp25 has updated to the latest parameters (check commits). `d` was not used there.

valga:
@mgp25 just forgot about it. `d` is double tap flag.

SteveJobzniak:
@valga Ahh so `d` is used when the user double-taps the image to like it (instead of pressing the heart icon manually)? Then we don't need `d`.

valga:
`d=1` when like is done via double tap and `d=0` when like is done via heart icon. So it is always required.

SteveJobzniak:
@valga Well, mgp used the latest APK to check the parameters and he said it doesn't have that param. Hmm.

@mgp25 Want to comment? :)

mgp25:
I haven't tested liking with double tap, i dont think its required if we have already this one with the heart icon. Nevertheless, we can do another function called `doubleTapLike()` but i think it isnt required.

Thoughts? @SteveJobzniak @valga

valga:
@mgp25 Please post a captured POST body for like call made with your APK. You can strip signed body.

mgp25:
@valga will check tomorrow, maybe i missed it.

mgp25:
@valga yup, i can confirm what you have said. Nevertheless, i want to discuss something in private. If you want send me an email or we can create a private chat in gitter.

Regards

valga:
@mgp25, don't forget to add that unsigned param to like/unlike methods.

mgp25:
@valga yup, im checking and automatizing a couple of things and ill add them later tonight :p

PR #1434, opened 2017-07-27
Allow non-square direct videos

valga:
It turned out that resumable uploader already had it, so I just tweaked upload params for segmented uploader. With a tiny bit of refactoring.


SteveJobzniak:
@valga Brilliant! I will review this today but at a quick read literally everything seems great!

I only wonder why `public function useResumableUploader` instead of `protected function _useResumableUploader`? I'm just trying to think imagine if there's a point of letting external functions (outside of Internal.php) check if the resumable uploader will be used or not?

SteveJobzniak:
Okya the review is complete and everything looks perfect (I love the new way of generating upload params). The only downside is getVideoUploadParams and useResumableUploader. Their purpose is only within Internal.php, so let's make them protected, unless you have some specific reason not to do that?

valga:
What do you think about making all protected upload methods in `Internal` public? It's great for power users, because they can, for example, serialize internal metadata and run separate uploads steps manually.

SteveJobzniak:
Do you mean that they can manually call all the various things like "5x Upload Video, then after those are done run all the configure calls manually" etc?

Yeah... I suppose it doesn't hurt to make it all public if someone wants to mess around like that.

In this PR or a new one?

valga:
New one, tomorrow.

SteveJobzniak:
Okay that sounds good. But before merging, let's quickly think about the two new public functions: Will getVideoUploadParams and useResumableUploader be needed for power users? It seems like at least `_useResumableUploader` has no external purpose even for power users and should be protected? Or do you want to make every function public?

SteveJobzniak:
Alright. Here's the list that needs to be decided (what should be public or protected):

```
public function getVideoUploadParams
public function useResumableUploader
protected function _requestVideoUploadURL
protected function _saveExperiments
protected function _getFirstMissingRange
protected function _uploadVideoChunks
protected function _uploadResumableVideo
```

SteveJobzniak:
I could commit the function visibility changes after merging, if we can just define what should be public or protected out of the list above.

valga:
> So it looks like we could keep everything protected and make _getVideoUploadParams and _useResumableUploader protected...

Ok, you can do it.

SteveJobzniak:
Okay done https://github.com/mgp25/Instagram-API/commit/6c8cd70e934b1de4e0d5dd0b45b7166cc52bb3d2

PR #1435, opened 2017-07-27
Unified inbox is not an experiment anymore

valga:
So we must always send `use_unified_inbox` param.


SteveJobzniak:
Nice, very clear commit here. I just wonder why `hasUnifiedInbox()` is not removed? The function is not called anywhere anymore and the experiment it checks for is not stored after this change. So the function should be deleted too, right?

valga:
I just forgot about it. Removed. Don't forget to squash this PR.

mgp25:
@valga version?

valga:
@mgp25, since 10.24.0.

mgp25:
@valga perfect, everything seems okay and ready to merge.

SteveJobzniak:
👍

PR #1436, opened 2017-07-27
Use PHP-CS-Fixer

valga:
To get rid of "StyleCI" commits.

Usage: `composer codestyle`.


mgp25:
@valga checking 👍

SteveJobzniak:
@valga Very cool. I haven't seen this package before. Nice to have a built-in "StyleCI". We just have to be sure this one doesn't enforce a very different style than what we used StyleCI for.

valga:
@SteveJobzniak I picked up rules that don't change anything important in existing code. In fact, you will just reindent a couple of lines in `examples` script after running that command at the moment.

However, this ruleset fixes most annoying things, like:
```php
    /**
     * @param int $aa Test
     * @param callable $b Again
     * @throws \InvalidArgumentException Test
     * @throws \RuntimeException Test2
     * @return int
     */
    public function test(
        $aa,
        callable $b)
    {
        $b([
            't' => $aa,
            "tt" => $aa+1,
        ]);
        return 1;
    }
```
will be changed to:
```php
    /**
     * @param int      $aa Test
     * @param callable $b  Again
     *
     * @throws \InvalidArgumentException Test
     * @throws \RuntimeException         Test2
     *
     * @return int
     */
    public function test(
        $aa,
        callable $b)
    {
        $b([
            't'  => $aa,
            'tt' => $aa + 1,
        ]);

        return 1;
    }
```

mgp25:
Seems okay so far. @SteveJobzniak ?

SteveJobzniak:
That is fantastic. The most annoying thing is when we miss like 1 space before a PHPdoc param name or somethibg, and have to StyleCI that later...

I am on iPad right now, but was reading through some of the PHPCSFixer docs and saw a bunch of interesting ones. This is a list to research for possible conclusion. I also read about how to create a fixer and it seems like it may be pretty easy to do one that looks at function definition arguments and ensures they are indented (something like: if token == a function definition argument, check if previous token is indentation). But their coding system is not entirely clear so I won't open that can of worms by trying to write that fixer yet.

Herr are some interesting fixers I found (I bet some are already in this PR):

phpdoc_add_missing_param_annotation

single_line_after_imports

function_declaration

no_spaces_inside_parenthesis

no_unneeded_control_parentheses

method_argument_space

mgp25:
@SteveJobzniak [Off-Topic]: API seems ready to go for release 3. I'll wait for these PR that are open. Im going to check minor stuff, but seems okay. Do you want to write a resume of the changes, i'll do a brief one, but i know you like to write a lot xD

SteveJobzniak:
@mgp25 Cool. So there's no more functions that need to be added?

In that case, here's our todo list before V3:
- Merge these pull requests.
- Review the dimension and aspect ratio changes in the ticket (we're not done with that discussion yet, I was gonna get to that after these pull requests).
- Possibly (probably) edit ImageAutoResizer to add support for story-cropping (since right now it only supports timeline-cropping).
- Write a summary. Yeah I can do it when the day comes. :)

valga:
> `phpdoc_add_missing_param_annotation`

I didn't find it useful, because it will not write a description for param and it will not guess type (in case of mixed).

> `no_unneeded_control_parentheses`

I don't remember anyone writing something like `return ($a);`, but it's safe to include this rule.

Other rules are already included into `@PSR2` metarule.

SteveJobzniak:
@valga Ahh, yeah I thought maybe that was what "@PSR2" meant in their readme...

Okay so about the `no_unneeded_control_parentheses`: Yeah let's add it. We don't write like that. And we want to be sure it stays that way. It also protects against `echo()` and `print()` (which isn't the same as `printf()`). :-)

And about `phpdoc_add_missing_param_annotation`: Hmm... well, it looks useful as a protection against missed params, as long as we set `only_untyped=false` to make it document all missing params. So if a param is not documented, it apparently does something like `@param mixed $foo` and `@param someclass $bar`? If so then that's better than no param documentation at all, and when it adds the param text we will notice the lack of description text and write about that param manually.

SteveJobzniak:
This is a great pull request by the way, I look forward to playing with it. I am going to start using this codestyle checking method in my own code too. I had no idea PHP-CS-Fixer existed. 👍

valga:
Ok, I tweaked the ruleset to inherit it from `@Symfony` metarule, so `no_unneeded_control_parentheses` is included. Also I included `phpdoc_add_missing_param_annotation` and fixed existing PHPDoc to prevent false-positives on that rule.

SteveJobzniak:
@valga Oh that's a great change. Symfony includes more cleanliness rules and made our custom list shorter. Nice!

About the existing PHPdoc changes: Those were not false positives. ;-) Those were clear bugs in the PHPdoc. Bad param variable names, accidental `@var` instead of `@param` etc. They were caught thanks to the new rule, hehe. Awesome.

I noted above that they have a config option though:

```
phpdoc_add_missing_param_annotation

Phpdoc should contain @param for all params.

Configuration options:

only_untyped (bool): whether to add missing @param annotations for untyped parameters only; defaults to true
```

It sounds useful. Otherwise it will only fix "func($a)" but not "func(SomeClass $a)"... So `'phpdoc_add_missing_param_annotation' => true,` to `'phpdoc_add_missing_param_annotation' => ['only_untyped' => false],`.

And lastly let's move to PHP-CS-Fixer's recommended name (it's in their docs):
- Rename `.php_cs` to `.php_cs.dist`. Having ".dist" for the project defaults is standard practice for composer packages.
- Add `.php_cs` to `.gitignore`. That's for local repo user-overrides of the rules.

@valga I could merge this and then do these 3 minor changes. Or you can do them?

Also, about PHPUnit: Yeah I use it all the time! But so far I haven't seen much need for it in this project since 99% of the functions perform API calls and we can't validate those using local "server response" mockup data since the server may not act that way and then we'd have outdated local mockup data. And as for unit testing stuff like Utils functions, yeah that's nice I guess... that's one area where we could use it.

SteveJobzniak:
Yeah... https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/Phpdoc/PhpdocAddMissingParamAnnotationFixer.php

They're using `only_untyped` to decide whether it will add `@param` for missing params. If the param has no typename, it will get documented. If it has a typename it will not get documented (even though it's missing from PHPdoc) unless `only_untyped => false`.

SteveJobzniak:
@valga I'll merge in half an hour or so, so let me know if you want to quickly commit those fixes as part of the pull request instead of me adding separate commits.

valga:
Feel free to do all things you mentioned.

As for PHPUnit, we can cover just critical things, such as requests builder and errors handler.

SteveJobzniak:
@valga Alright it's merged, updated config a little bit, and ran it. It found some mistakes. This is awesome. I'm definitely going to copy this great tool into my personal projects too. Incredibly good stuff!

As for PHPUnit, I don't have enough health/energy to help writing tests. I'm exhausted. :\

@mgp25 From now on remember to always run `composer codestyle` before saving commits.

I recommend first staging your own code, then run "composer codestyle" and look at what it changes compared to your staged work, and then stage those changes too.

Why? Because: It may add something like `@param mixed $someParamForgottenToDocument`... and we need to manually realize that mistake and write proper docs for those params (proper type and description).

So by staging your own work first, and then running a fix and seeing what it changes, you'll see if it does any codestyle changes that will need your manual correction.

Anyway, wow this code checker tool is awesome! So much fun!

---

Here's the updated todo list for 3.0: https://github.com/mgp25/Instagram-API/pull/1436#issuecomment-318430572

PR #1447, opened 2017-07-31
Direct: Remove sendLink() in favor of sendText()

valga:
Realtime doesn't support sending links, so it is left untouched.


SteveJobzniak:
@valga Excellent. I caught an `array $url` type which won't work in PHP5.6 but fixed it directly here.

By the way, as I mentioned in the ticket: This will send stuff like "Check out google.com" as `links: 'google.com'`. If we want to only send links with `protocol://` prefix, we'd simply have to skip any URLs with empty `['protocol']` array value. Should we do that, or is it safe to send `google.com` as a links value? I merged anyway since I assume that you saw the question in the ticket and that it's therefore safe. :)

valga:
@SteveJobzniak hmm, `array` typehint is from PHP5.1. Am I missing something?

> This will send stuff like "Check out google.com" as links: 'google.com'. Is that legal?

Yes, app does the same.

SteveJobzniak:
@valga No, you're not missing something. I had just woken up and had a brainfart. I constantly use array typehint but somehow in the moment after waking up I forgot. I'll re-add it.

About the links: PERFECT! :) Glad to hear it.

mgp25:
@SteveJobzniak gitter

PR #1464, opened 2017-08-08
Request/Live: Post live viewers list and fixes for responses

solomax2014:


mgp25:
@solomax2014 is `getPostLiveViewerList()` a not signed request? no extra post param?

solomax2014:
@mgp25, I don't know about signing, I think if these two methods don't require it, so this doesn't require it also. The request works fine. Yes, there is a ```max_id``` param, that I forgot, I'm going to fix it now.

solomax2014:
@mgp25, ok, I've added ```max_id``` param. Unfortunately, I don't have so many users to test it, but I think it works like in other functions.

mgp25:
@solomax2014 im going to merge it since its okay, but let me know where in the app should i go, to double check if everything is fine.

Regards

solomax2014:
@mgp25, it is the eye button (bottom left) (if you are the owner of this post live). Or try to slide up the screen to see it.

PR #1468, opened 2017-08-09
Check resources before closing them

valga:
It turned out that Guzzle's streams close supplied handles in `__destruct()`, so there is a chance that handle is already has been closed when we are trying to close it. And that leads to `supplied resource is not a valid stream resource` errors.


mgp25:
@valga i have checked guzzle repo and you are right.

Thanks!

PR #1469, opened 2017-08-09
Added example for live videos.

matthijsotterloo:


aitte2:
This example lacks how to stream the video, and how to stop the stream, so it's an incomplete example...?

matthijsotterloo:
If we’d want to make an example on how to stream there’s way more code and libraries needed.

aitte2:
Then why this file?

mgp25:
This example lacks of information, nevertheless i'll push one example with more details and ways to use it. Actually, there are many ways to use it and no extra code is required.

I'll add it later, i need to review few things before.

Regards

aitte2:
@mgp25 nice! I also think you should change the parameter order of create():

```php
create('1184', '720', $streamMessage);
```

It's an annoying order. If I would want to use the default screen size but just change message, I have to provide a screen size too.

Current order: `$height, $width, $message`
Better order: `$message, $width, $height`

maybe? :)

solomax2014:
@aitte2, the better order is ```$width, $height, $message```, as some people don't use the broadcast message.

aitte2:
@solomax2014 Okay, I agree with that!

solomax2014:
@aitte2, as I know, 720 is maximum width for video stories, so I think it can be applied to lives also. I've seen that usual dash_playback_url gives 396x704 resolution, now I'll check the dash_abr_playback_url resolutions, because I've seen there many of video streams.

solomax2014:
@aitte2, video representations from dash_abr_playback_url manifest:
- live-md-v: 324x576 (FBQualityLabel="324p")
- live-hd-v: 396x704 (FBQualityLabel="396p") (it's the only video representation in dash_playback_url)
- live-ld-v: 216x384 (FBQualityLabel="240p")
- live-hd1-v: 504x896 (FBQualityLabel="480p")

bookin:
Can someone say me how I can stop the stream (after )? Thanks.

aitte2:
@bookin search open tickets, dude

PR #1493, opened 2017-08-17
Fix sprintf() args

### LINKED ISSUE #1487: Direct sendPost bug
hi,
media ID in Instagram's internal format has been changed , and regex in `sendPost` method is wrong and always throw an invalid mediaId exception.

and in `Direct.php` on `line 521` and `527` there is missing parameter for sprintf method and the parameter passed for InvalidArgumentException in a wrong way , and raise an error.

```
 Too few arguments in /vendor/mgp25/instagram-php/src/Request/Direct.php on line 521.
 Too few arguments in /vendor/mgp25/instagram-php/src/Request/Direct.php on line 527.
```

valga:
Fixes #1487.


mgp25:
@valga sure, perfect.

Do you mind running `composer codestyle` with your PR? Since i forgot in the previous commit.

valga:
@mgp25 Sure. But you really need to set up a pre-commit hook.

Ok, I believe I found all issues with `sprintf()` so it is ready to be merged.

PR #1498, opened 2017-08-19
Request/Live: New function deletePostLive()

itsMoji:
Request/Live: New function deletePostLive()
add PhpStorm temporary files to .gitignore


mgp25:
@itsMoji if `deletePostLive` only has `{"status": "ok"}` as response, you should not create a new response and use `GenericResponse` instead. 

See `like()` as an example.

Once you have done these changes, it is ready to be merged.

Regards

itsMoji:
@mgp25 yes its return only `{"status": "ok"}`
I used addToPostLive() as example and addToPostLive returns `{"status": "ok"}` too
so addToLiveResponse is not necessary so i delete it

mgp25:
@itsMoji now it's perfect! 😄 

Thanks for contributing

Regards

PR #1523, opened 2017-08-30
Refactor Realtime/Push

valga:
This PR minimizes coupling between Realtime/Push clients and Instagram instance. Existing code has even `PATCH` op that depends on Instagram instance. After merging this PR one will be able to run RT/Push clients in their own projects even without instantiating Instagram class.

Also this PR addresses some fixes and improvements:
1. User-Agent string and Client Stack value have been fixed for Push client.
2. Direct experiments key is updated to match current app version.
3. Realtime events do not process event data if target has no listeners for this event. This will save some resources.
4. Realtime doesn't refresh user session anymore, because it is not its job.

BC breaks:
1. Realtime doesn't have `init()` method anymore.


valga:
Ok, it is ready to be merged. Don't forget to squash the commits.

I have some other stuff in queue, but it depends on PRs filed to another repos, so I will do it later (maybe).

mgp25:
@valga everything seems good. Awesome work as always! 😄 👌

SteveJobzniak:
❤️

Issue #1199, opened 2017-04-21
How to Reed feedback title?


lonlyman:
While i am testing the api , i sent the same comment to a Media twice by mistake , but i note something interesting , instagrams respond with a nice message .

`"feedback_title": "Duplicate Comment"`

how can i get this message without all other stuff within the responde ?
is there a getter for Feedback Title?

With Regards.

mgp25:
@lonlyman have you read the documentation? `getFeedbackTitle()`

lonlyman:
Dear @mgp25 , i am an educated man, and yes i did search and read before creating this issues.
here is the $RESPONSE:


` {"message": "feedback_required", "feedback_title": "Duplicate Comment", "feedback_message": "It looks like you already left this comment. Try posting a new one.", "feedback_url": "repute/report_problem/instagram_comment/", "feedback_appeal_label": "Report problem", "feedback_ignore_label": "OK", "feedback_action": "report_problem", "status": "fail"}`

try{
post comment 
}catch (\Exception $e){
```
$e->getFeedbackTitle();
$e->getFeedbackMessage();
```
}

None of them worked , but an error message returned that there is no methods called getFeedbackTitle .

mgp25:
@lonlyman because it is an exception but i'll add more exceptions when i have some free time. I'll ping you when its done, nevertheless if i forget it, ping me

Regards

lonlyman:
@mgp25 So it is my turn to tell you " have you read the documentation?"
LOL just joking ...

Thank you in advanced

SteveJobzniak:
@lonlyman Hi. Since we keep getting requests from people who want to get extra server reply details in some exceptions, I thought it was finally time to do something about it.

So I've extended the exception system. Every exception in this library is derived from InstagramException, which in turn now has brand-new `hasResponse()` and `getResponse()` functions!

With this, people can finally get complete server response details when they need it.

I've created an example in https://github.com/mgp25/Instagram-API/blob/master/examples/exceptionDetailsExample.php

In your case the code would be like:

```php
try{
     post comment 
} catch (\InstagramAPI\Exception\FeedbackRequiredException $e) {
    if ($e->hasResponse()) {
        var_dump($e->getResponse()->getFeedbackTitle());
    }
}
```

There will most likely not be any more exceptions. The set we have is perfect. It covers all critical CORE errors (such as "login_required", "feedback_required", "sentry_block", etc), and it also covers all per-endpoint problems via EndpointException.

[...]

PR #1534, opened 2017-09-03
Do not use JsonMapper

valga:
@SteveJobzniak Hi, I want to hear your opinion on this. 

I ported a generator from one of my projects, so there was not so much work on this (in case of rejection). I don't like JsonMapper, because it lacks lazy loading, also it has to process reflection data at least once (or every time, if you run queues to send requests).

Basically this PR replaces JsonMapper with auto-generated setters and getters. I included a couple of samples for generated code. It certainly could be simplified, for example we could skip caching for scalar values.

Also it has a big BC: we must declare all properties protected (or even private), because there must be only one method for obtaining or setting a data.

### DIFF
--- a/devtools/generateAutoPropertyHandlerMethods.php
+++ b/devtools/generateAutoPropertyHandlerMethods.php
@@ -1,0 +1,521 @@
+<?php
+
+set_time_limit(0);
+date_default_timezone_set('UTC');
+
+require __DIR__.'/../vendor/autoload.php';
+
+$mode = isset($argv[1]) ? $argv[1] : AutoMethods::MODE_APPEND;
+
+$updater = new AutoMethods(__DIR__.'/../src/', $mode);
+$processedFiles = $updater->run();
+exit((int) ($mode !== AutoMethods::MODE_VALIDATE || !count($processedFiles)));
+
+class AutoMethods
+{
+    const MODE_APPEND = 'append';
+    const MODE_REWRITE = 'rewrite';
+    const MODE_VALIDATE = 'validate';
+
+    const PADDING = "XXXPADDINGXXX\n";
+
+    /**
+     * @var string
+     */
+    private $_dir;
+
+    /**
+     * @var string
+     */
+    private $_mode;
+
+    /**
+     * @return array
+     */
+    public function getAvailableModes()
+    {
+        return [
+            self::MODE_APPEND,
+            self::MODE_REWRITE,
+            self::MODE_VALIDATE,
+        ];
+    }
+
+    /**
+     * Constructor.
+     *
+     * @param string $dir  Directory to process.
+     * @param string $mode Updated mode
+     */
+    public function __construct(
+        $dir,
+        $mode)
+    {
+        $this->_dir = realpath($dir);
+        if ($this->_dir === false) {
+            throw new InvalidArgumentException(sprintf('"%s" is not a valid path.', $dir));
+        }
+        if (!in_array($mode, $this->getAvailableModes())) {
+            throw new InvalidArgumentException(sprintf('"%s" is not a valid mode.', $mode));
+        }
+        $this->_mode = $mode;
+    }
+
+    /**
+     * Convert file path to class name.
+     *
+     * @param string $filePath
+     *
+     * @return string
+     */
+    private function _extractClassName(
+        $filePath)
+    {
+        return '\InstagramAPI'.str_replace('/', '\\', substr($filePath, strlen($this->_dir), -4));
+    }
+
+    /**
+     * Extract property type from its PHPDoc.
+     *
+     * @param ReflectionProperty $property
+     *
+     * @return string
+     */
+    private function _getType(
+        ReflectionProperty $property)
+    {
+        $phpDoc = $property->getDocComment();
+        if ($phpDoc === false || !preg_match('#@var\s+([^\s]+)#i', $phpDoc, $matches)) {
+            $type = 'mixed';
+        } else {
+            $type = $matches[1];
+        }
+
+        return $type;
+    }
+
+    /**
+     * Converts underscores to camel cases.
+     *
+     * @param string $property
+     *
+     * @return string
+     */
+    private function _camelCase(
+        $property)
+    {
+        // Trim any leading underscores and save their count, because it's a special case.
+        $result = ltrim($property, '_');
+        $leadingUnderscores = strlen($property) - strlen($result);
+        if (strlen($result)) {
+            // Convert all chars prefixed with underscore to upper case.
+            $result = preg_replace_callback('#_([^_])#', function ($matches) {
+                return strtoupper($matches[1]);
+            }, $result);
+            // Convert fist char to upper case.
+            $result[0] = strtoupper($result[0]);
+        }
+        // Restore leading underscores (if any).
+        if ($leadingUnderscores) {
+            $result = str_pad($result, strlen($result) + $leadingUnderscores, '_', STR_PAD_LEFT);
+        }
+
+        return $result;
+    }
+
+    /**
+     * Generate a list of methods signatures based on properties list.
+     *
+     * @param ReflectionClass $reflection
+     *
+     * @return array Map of method name => [source, type]
+     */
+    private function _generateMethodsSignatures(
+        ReflectionClass $reflection)
+    {
+        $result = [];
+        $properties = $reflection->getProperties();
+        $parent = $reflection->getParentClass();
+        foreach ($properties as $property) {
+            if (strpos($property->getDocComment(), '@internal') !== false) {
+                continue;
+            }
+            $propertyName = $property->getName();
+            // Determine property type.
+            $type = $this->_getType($property);
+            // Skip properties available in parent class.
+            if ($parent->hasProperty($propertyName)) {
+                $parentType = $this->_getType($parent->getProperty($propertyName));
+                if ($parentType === $type) {
+                    continue;
+                }
+            }
+
+            // Normalize property name.
+            $normalizedName = $this->_camelCase($propertyName);
+            $signature = [$propertyName, $type];
+            // getPropertyName() method.
+            $getter = 'get'.$normalizedName;
+            $result[$getter] = $signature;
+
+            // setPropertyName() method.
+            $setter = 'set'.$normalizedName;
+            $result[$setter] = $signature;
+
+            // isPropertyName() method.
+            $iser = 'is'.$normalizedName;
+            $result[$iser] = $signature;
+        }
+
+        return $result;
+    }
+
+    /**
+     * @param ReflectionClass $reflection
+     *
+     * @return ReflectionMethod[]
+     */
+    private function _getExistingMethods(
+        ReflectionClass $reflection)
+    {
+        $result = [];
+        foreach ($reflection->getMethods() as $method) {
+            $result[$method->getName()] = $method;
+        }
+
+        return $result;
+    }
+
+    /**
+     * @param ReflectionMethod $method
+     * @param array            $lines
+     *
+     * @return array
+     */
+    private function _removeExistingMethod(
+        ReflectionMethod $method,
+        array $lines)
+    {
+        $result = $lines;
+        $from = $method->getStartLine();
+        $phpDoc = $method->getDocComment();
+        if (!empty($phpDoc)) {
+            $from -= substr_count($method->getDocComment(), "\n") + 1;
+        }
+        $to = $method->getEndLine();
+
+        for ($i = $from - 1; $i < $to; ++$i) {
+            $result[$i] = self::PADDING;
+        }
+
+        return $result;
+    }
+
+    /**
+     * @param array $lines
+     *
+     * @return array
+     */
+    private function _filterEmptyLines(
+        array $lines)
+    {
+        $result = array_filter($lines, function ($line) {
+            return $line !== self::PADDING;
+        });
+        $prevLine = '';
+        $result = array_filter($result, function ($line) use (&$prevLine) {
+            $line = trim($line);
+            if ($prevLine === '' && $line === '') {
+                return false;
+            } else {
+                $prevLine = $line;
+
+                return true;
+            }
+        });
+
+        return $result;
+    }
+
+    /**
+     * @param string $type
+     *
+     * @return bool
+     */
+    private function _isSimple(
+        $type)
+    {
+        return in_array($type, ['string', 'bool', 'int', 'float', 'double', 'array']);
+    }
+
+    /**
+     * @param string $type
+     *
+     * @return bool
+     */
+    private function _isArray(
+        $type)
+    {
+        return substr($type, -2) === '[]' && !$this->_isMixed($type);
+    }
+
+    /**
+     * @param $type
+     *
+     * @return bool
+     */
+    private function _isMixed(
+        $type)
+    {
+        return $type === 'mixed' || strpos($type, '|') !== false;
+    }
+
+    /**
+     * @param $source
+     * @param $type
+     *
+     * @return array
+     */
+    private function _generateArrayGetter(
+        $source,
+        $type)
+    {
+        $result = [];
+        $result[] = "            if (is_array(\$this->_jsonData->{$source})) {\n";
+        $result[] = "                \$this->{$source} = [];\n";
+        $result[] = "                foreach (\$this->_jsonData->{$source} as \$idx => \$value) {\n";
+        if ($this->_isMixed($type) || $this->_isSimple($type)) {
+            $result[] = "                    \$this->{$source}[\$idx] = \$value;\n";
+        } elseif ($this->_isArray($type)) {
+            throw new \RuntimeException('No support for inner arrays.');
+        } else {
+            $result[] = "                    \$this->{$source}[\$idx] = new {$type}(\$value);\n";
+        }
+        $result[] = "                }\n";
+        $result[] = "            } else {\n";
+        $result[] = "                \$this->{$source} = null;\n";
+        $result[] = "            }\n";
+
+        return $result;
+    }
+
+    /**
+     * @param $methodName
+     * @param $source
+     * @param $type
+     *
+     * @return array
+     */
+    private function _generateGetter(
+        $methodName,
+        $source,
+        $type)
+    {
+        $result = [];
+        $result[] = "\n";
+        $result[] = "    /**\n";
+        $result[] = "     * @return {$type}\n";
+        $result[] = "     */\n";
+        $result[] = "    public function {$methodName}()\n";
+        $result[] = "    {\n";
+        $result[] = "        if (\$this->{$source} !== null) {\n";
+        $result[] = "            return \$this->{$source};\n";
+        $result[] = "        } elseif (isset(\$this->_jsonData->{$source})) {\n";
+        if ($this->_isMixed($type) || $this->_isSimple($type)) {
+            $result[] = "            \$this->{$source} = \$this->_jsonData->{$source};\n";
+        } elseif ($this->_isArray($type)) {
+            $result = array_merge($result, $this->_generateArrayGetter($source, substr($type, 0, -2)));
+        } else {
+            $result[] = "            \$this->{$source} = new {$type}(\$this->_jsonData->{$source});\n";
+        }
+        $result[] = "\n";
+        $result[] = "            return \$this->{$source};\n";
+        $result[] = "        } else {\n";
+        $result[] = "            return null;\n";
+        $result[] = "        }\n";
+        $result[] = "    }\n";
+        $reuslt[] = "\n";
+
+        return $result;
+    }
+
+    /**
+     * @param $methodName
+     * @param $source
+     * @param $type
+     *
+     * @return array
+     */
+    private function _generateSetter(
+        $methodName,
+        $source,
+        $type)
+    {
+        if ($this->_isArray($type) || $type === 'array') {
+            $valueType = 'array ';
+        } elseif ($this->_isMixed($type) || $this->_isSimple($type)) {
+            $valueType = '';
+        } else {
+            $valueType = $type.' ';
+        }
+        $result = [];
+        $result[] = "\n";
+        $result[] = "    /**\n";
+        $result[] = "     * @param {$type} \$value\n";
+        $result[] = "     *\n";
+        $result[] = "     * @return static\n";
+        $result[] = "     */\n";
+        $result[] = "    public function {$methodName}(\n";
+        $result[] = "        {$valueType}\$value)\n";
+        $result[] = "    {\n";
+        $result[] = "        \$this->{$source} = \$value;\n";
+        $result[] = "        \$this->_jsonData->{$source} = \$value;\n";
+        $result[] = "\n";
+        $result[] = "        return \$this;\n";
+        $result[] = "    }\n";
+        $reuslt[] = "\n";
+
+        return $result;
+    }
+
+    /**
+     * @param $methodName
+     * @param $source
+     *
+     * @return array
+     */
+    private function _generateIser(
+        $methodName,
+        $source)
+    {
+        $result = [];
+        $result[] = "\n";
+        $result[] = "    /**\n";
+        $result[] = "     * @return bool\n";
+        $result[] = "     */\n";
+        $result[] = "    public function {$methodName}()\n";
+        $result[] = "    {\n";
+        $result[] = "        return isset(\$this->_jsonData->{$source}) && \$this->_jsonData->{$source};\n";
+        $result[] = "    }\n";
+        $reuslt[] = "\n";
+
+        return $result;
+    }
+
+    /**
+     * @param string $methodName
+     * @param array  $signature
+     * @param array  $lines
+     *
+     * @return array
+     */
+    private function _generateMethodCode(
+        $methodName,
+        array $signature,
+        array $lines)
+    {
+        list($source, $type) = $signature;
+        if (!substr_compare($methodName, 'get', 0, 3)) {
+            $method = $this->_generateGetter($methodName, $source, $type);
+        } elseif (!substr_compare($methodName, 'set', 0, 3)) {
+            $method = $this->_generateSetter($methodName, $source, $type);
+        } elseif (!substr_compare($methodName, 'is', 0, 2)) {
+            $method = $this->_generateIser($methodName, $source);
+        } else {
+            $method = [];
+        }
+
+        return $method;
+    }
+
+    /**
+     * @param string             $filePath
+     * @param array              $newMethods
+     * @param ReflectionMethod[] $existingMethods
+     * @param int                $insertAt
+     *
+     * @return bool
+     */
+    private function _generateMethods(
+        $filePath,
+        array $newMethods,
+        array $existingMethods,
+        $insertAt)
+    {
+        $lines = file($filePath);
+
+        foreach ($newMethods as $methodName => $signature) {
+            if (isset($existingMethods[$methodName])) {
+                $lines = $this->_removeExistingMethod($existingMethods[$methodName], $lines);
+            }
+            $method = $this->_generateMethodCode($methodName, $signature, $lines);
+            array_splice($lines, $insertAt - 1, 0, $method);
+            $insertAt += count($method);
+        }
+
+        $lines = $this->_filterEmptyLines($lines);
+        file_put_contents($filePath, implode('', $lines));
+
+        return true;
+    }
+
+    /**
+     * Process single file.
+     *
+     * @param string          $filePath
+     * @param ReflectionClass $reflection
+     *
+     * @return bool
+     */
+    private function _processFile(
+        $filePath,
+        ReflectionClass $reflection)
+    {
+        $allMethods = $this->_generateMethodsSignatures($reflection);
+        $existingMethods = $this->_getExistingMethods($reflection);
+        $diff = array_diff_key($allMethods, $existingMethods);
+        switch ($this->_mode) {
+            case self::MODE_APPEND:
+                $result = count($diff)
+                    ? $this->_generateMethods($filePath, $diff, $existingMethods, $reflection->getEndLine())
+                    : true;
+                break;
+            case self::MODE_REWRITE:
+                $result = count($allMethods)
+                    ? $this->_generateMethods($filePath, $allMethods, $existingMethods, $reflection->getEndLine())
+                    : true;
+                break;
+            case self::MODE_VALIDATE:
+                $result = !count($diff);
+                break;
+            default:
+                throw new RuntimeException(sprintf('Unknown mode "%s".', $this->_mode));
+        }
+
+        return $result;
+    }
+
+    /**
+     * Process all *.php files in given path.
+     *
+     * @return array
+     */
+    public function run()
+    {
+        $directoryIterator = new RecursiveDirectoryIterator($this->_dir);
+        $recursiveIterator = new RecursiveIteratorIterator($directoryIterator);
+        $phpIterator = new RegexIterator($recursiveIterator, '/^.+\.php$/i', RecursiveRegexIterator::GET_MATCH);
+
+        $result = [];
+        foreach ($phpIterator as $filePath => $dummy) {
+            $reflection = new ReflectionClass($this->_extractClassName($filePath));
+            if ($reflection->isSubclassOf(\InstagramAPI\AutoPropertyHandler::class)) {
+                if ($this->_processFile($filePath, $reflection)) {
+                    $result[] = $filePath;
+                }
+            }
+        }
+
+        return $result;
+    }
+}
--- a/src/AutoPropertyHandler.php
+++ b/src/AutoPropertyHandler.php
@@ -30,119 +30,18 @@ namespace InstagramAPI;
 class AutoPropertyHandler
 {
     /**
-     * __CALL is invoked when attempting to access missing functions.
-     *
-     * This handler auto-maps setters and getters for object properties.
-     *
-     * NOTE: Don't worry about performance. This whole function takes about
-     * 0.025 MILLIseconds per property, which actually compares very favorably
-     * to native accessor functions (which take about 0.002 MILLIseconds per
-     * call).
-     *
-     * @param string $functionName Name of the method being called.
-     * @param array  $arguments    Array of arguments passed to the method.
-     *
-     * @throws \Exception If the function type or property name is invalid.
-     *
-     * @return mixed
-     *
-     * @see http://php.net/manual/en/language.oop5.magic.php
+     * @var object|null
      */
-    public function __call(
-        $functionName,
-        $arguments)
-    {
-        // Extract the components of the function they tried to call.
-        $chunks = self::explodeCamelCase($functionName);
-        if ($chunks === false || count($chunks) < 2) {
-            throw new \Exception("Unknown function {$functionName}.");
-        }
-
-        // Determine the type (such as "get") and the property (ie "is_valid").
-        $functionType = array_shift($chunks);
-        $propertyName = implode('_', $chunks); // "is_valid"
-
-        // Some objects have rare camelcase properties, so instead of naming it
-        // "i_tunes_item" they have "iTunesItem" as the property. We'll have to
-        // check for the existence of a camelcase property and use it if found.
-        if (($len = count($chunks)) >= 2) {
-            // Make word 2 and higher start with uppercase ("i,Tunes,Item").
-            // NOTE: Instagram's rule so far is that the first word is always
-            // lowercase when they use camelcase.
-            for ($i = 1; $i < $len; ++$i) {
-                $chunks[$i] = ucfirst($chunks[$i]);
-            }
-            $camelPropertyName = implode('', $chunks); // "iTunesItem"
-            if (property_exists($this, $camelPropertyName)) {
-                $propertyName = $camelPropertyName; // Use this name instead.
-            }
-        }
-
-        // Make sure the requested function has a corresponding object property.
-        if (!property_exists($this, $propertyName)) {
-            throw new \Exception("Unknown function {$functionName}.");
-        }
-
-        // Return the kind of response expected by their desired function.
-        switch ($functionType) {
-        case 'get':
-            return $this->{$propertyName};
-            break;
-        case 'set':
-            $this->{$propertyName} = $arguments[0];
-            break;
-        case 'is':
-            return (bool) $this->{$propertyName};
-            break;
-        default:
-            // Unknown function type prefix...
-            throw new \Exception("Unknown function {$functionName}.");
-        }
-    }
+    protected $_jsonData;
 
     /**
-     * Explodes a string on camelcase boundaries.
+     * Constructor.
      *
-     * Examples:
-     * - "getSome0XThing" => "get", "some0", "x", "thing".
-     * - "getSome0xThing" => "get", "some0x", "thing".
-     *
-     * @param string $inputString
-     *
-     * @return string[]|bool Array with parts if successful, otherwise FALSE.
+     * @param object|null $jsonData
      */
-    public static function explodeCamelCase(
-        $inputString)
+    public function __construct(
+        $jsonData = null)
     {
-        // Split the input into chunks on all camelcase boundaries.
-        // NOTE: The input must be 2+ characters AND have at least one uppercase.
-        $chunks = preg_split('/(?=[A-Z])/', $inputString, -1, PREG_SPLIT_NO_EMPTY);
-        if ($chunks === false) {
-            return false;
-        }
-
-        // Process all individual chunks and make them all completely lowercase.
-        // NOTE: Since all chunks are split on camelcase boundaries above, it
-        // means that each chunk ONLY holds a SINGLE fragment which can ONLY
-        // contain at most a SINGLE capital letter (the chunk's first letter).
-        foreach ($chunks as &$chunk) {
-            $chunk = lcfirst($chunk); // Only first letter may be uppercase.
-        }
-
-        // If there are 2+ chunks and the first (the function type) ends with
-        // trailing underscores, it means that they wanted to access a property
-        // beginning with underscores, so move those to the start of the 2nd
-        // ("property name") chunk instead.
-        if (count($chunks) >= 2) {
-            $oldLen = strlen($chunks[0]);
-            $chunks[0] = rtrim($chunks[0], '_'); // "get_" => "get".
-            $lenDiff = $oldLen - strlen($chunks[0]);
-            if ($lenDiff > 0) {
-                // Move all underscores to prop: "messages" => "_messages":
-                $chunks[1] = str_repeat('_', $lenDiff).$chunks[1];
-            }
-        }
-
-        return $chunks;
+        $this->_jsonData = $jsonData;
     }
 }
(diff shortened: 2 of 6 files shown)


mgp25:
@valga this seems interesting. So we keep the responses as always, set the properties as we have being doing, and you did and auto setter/getter for each property. I haven't seen JsonMapper effiency. 

This PR is for code effiency, simplify or both? And the only thing to change in all response is `public` to `protected` as i have understood right?

Your opinion on this @SteveJobzniak ?

SteveJobzniak:
Hmm, this is very fascinating and I think I like it. The way it checks if a value is set already, and if so returns it. And if not, it loops through/processes/generates the value once and caches it. And the lazy loading is nice (nothing is processed until wanted).

But it has a few huge downsides.

- How does it allow us to override getters? Like getMessages() in the generic response trait. Important to be able to make custom functions.
- How does it handle mixing our custom functions with its autogenerated functions? When our response + model files are going to be filled with get/set/is-ers, it will be very hard to see our custom ones like `Item::isAd()` etc... Perhaps make it generate a section at the bottom of the class, under some `///////// AUTOMATIC blablabla:` header.
- It's bloated as hell right now. Every function is a copy-paste of itself. One way to solve that would be to use generic methods of generating the property name and the target variable type, such as `$this->{$propName}` and `new $className()` in a single generic internal `_getter()` etc function instead f every function repeating the code (with classes with 100 properties that'll be utter hell to read through).
- Lastly, I think perhaps a better idea is to just put all this code in the AutoPropertyHandler. Everything derives from that. Sure, it means we have to yet again regex-split and process every get/set/is-call but the performance of that is totally negligible. And it means we have a single piece of code which cleanly manages all data processing. For example `getSomething` would run AutoPropertyHandler's get-handling which in turn looks for the cached value, and if so returns it, if not reads it from the jsonData property, etc... Speaking of which, you don't need jsonData, you just need to read the StdObj in the fullResponse property.

Sorry if this feedback comment is hard to read. I am tired today. Spent last month in the hospital.

SteveJobzniak:
Hmm another very huge downside... very very huge...

var_dump($response) will no longer show people a full view of the available data.

Although, var_dump($response->getFullResponse()) will still let them see all available data I suppose...

valga:
@mgp25 main pros are lazy loading and flexibility.

> How does it allow us to override getters?

It has 2 modes: `append` and `rewrite`. First one creates only methods that do not exist yet. Second one rewrites existing methods. Also, there is a way to skip certain properties: https://github.com/valga/Instagram-API/blob/535ba9e0733a4f3a0d9d3dd157f28c8794edc419/devtools/generateAutoPropertyHandlerMethods.php#L140
It can be re-used for methods.

> How does it handle mixing our custom functions with its autogenerated functions?

It always writes at the end of class. So you can just put some note like `HUMAN SHALL WRITE ABOVE THIS LINE`.

> It's bloated as hell right now. Every function is a copy-paste of itself.

As I said, it could be simplified. For example, most properties are scalar ones. There is no point in caching their values, so most getters could be simplified to something like:
```php
/**
 * @return mixed
 */
public function getName()
{
    return isset($this->_jsonData->name) ? $this->_jsonData->name : null;
}
```

> Lastly, I think perhaps a better idea is to just put all this code in the AutoPropertyHandler.

I don't think it is a good idea, because at the end it will become something like JsonMapper again. I don't want magic methods, I don't want Reflection in runtime. It is ok to have a lot of identical code, since it is auto-generated.

> Perhaps we could make that clearer by adding a dump() function to everything that derives from AutoPropertyHandler.

Yep, I just forgot to add a method for dumping JSON data to AutoPropertyHandler. And this way is even more flexible: you can dump 'full response' even for every tiny object in Response.

SteveJobzniak:
About reflection + magic methods: We're talking about microseconds of performance. It doesn't matter.

This new system would still be faster and better than JsonMapper. It wouldn't be like JsonMapper again. Because your system would only process/map properties the actual 1st time they are read. JsonMapper recursively maps the entire tree of objects every time even if not a single field is ever read from it.

About identical code: You can still cut down the lines of code by a lot by doing `protected _autoGet($propName, $className) { $this->{$propName}... }` and `public getSomething() { return $this->_autoGet('something', Model\Something); }`. It would avoid all Responses + Models being thousands of lines long.

Anyway, I think we should hold 3.0 release for as long as it takes to discuss this proposal. It's interesting as hell! Nice work so far.

valga:
> About the comment to separate the automatic methods: Could that be automatically added?

Yeah, it can be added. But there is another issue:
1. We have `a` and `c` properties. After running generator in `append mode` we will get `getA()` and `getC()` methds.
2. We add `b` property between `a` and `c`. As generator always appends code to the end of class, we will have `getA()`, `getC()` and `getB()`.
3. Someone runs generator in `rewrite` mode. We will have `getA()`, `getB()` and `getC()` methods.

Basically nothing has changed, but we have a reordered methods.

>About identical code

In most cases it could be simplified to 1-2 lines of code:
```php
// scalar
return isset($this->_jsonData->name) ? $this->_jsonData->name : null;
// object
return isset($this->user) || !isset($this->_jsonData->user) ? $this->user : $this->user = new User($this->_jsonData->user);
// array of objects
return isset($this->users) || !isset($this->_jsonData->users) || !is_array($this->_jsonData->users) 
    ? $this->users : $this->users = array_map(function ($value) {return new User($value);}, $this->_jsonData->users);
```
So, again, I don't see a problem with identical code.

Anyway, I need to take a break for a couple days to rethink everything.

SteveJobzniak:
@valga Hmm, here's an idea: If a property isn't documented as `@var SomeClass` or `@var something[]` (arrays), then the simplified getter code can always be generated instead.

My main dislike so far is that it currently adds 44 lines of code **per property**. So a file with 100 properties will get `+4400` lines of code. As it is right now I am leaning towards decline.

With optimization so that non-class/non-array properties use a simplified getter, it would still be around 30 lines of code per property on average.

With optimization so that it uses generic functions stored in a base class, and each is/get/set function is a wrapper which just calls the generics, it would still be about 20 lines of code per property for the "wrapper" functions.

With the biggest optimization, the original AutoPropertyHandler method (magic methods), we would get zero extra lines of code. At a non-noticeable, slight cost of performance.

Anyway, yes, feel free to take a break to think about this, because we've decided to go ahead with 3.0 release as soon as possible to avoid all people's tickets from outdated versions. So this will make it into a 3.x.x release anyway.

SteveJobzniak:
Oh one more thing... The json_decoded()'s full response StdObj must always be available via `getFullResponse()` regardless of how we implement all other properties.

Because it's used for unmapped (missing) properties that are used for things like random error fields that don't always exist and therefore aren't in any base classes, such as this:

https://github.com/mgp25/Instagram-API/issues/1199#issuecomment-296448210

valga:
I decided to close this PR, because it certainly can't be merged in its current state. However, we must rethink our properties handling.

1. Each model must receive an object from parsed JSON in its constructor and save it.
2. We must remove all properties from our models, because they just eat memory for nothing, as we already have the object with all data from JSON. Consider following example:
```php
class TestProperties
{
    public $a;
    public $b;
    public $c;
    public $d;
    public $e;
    public $f;
    public $g;
    public $h;
    public $j;
    public $k;
    public $l;
    public $m;
    public $n;
}

class TestJson
{
    public $json;
}

foreach ([TestProperties::class, TestJson::class] as $className) {
    $mem = memory_get_usage();
    $object = new $className;
    printf("Mem per object: %dB\n", memory_get_usage() - $mem);
    $object = null;
}
```
You can see that PHP reserves a memory for all properties (actually it creates them with `null` value):
```
Mem per object: 256B
Mem per object: 56B
```
3. We must place all properties as `@property Type $name` in PHPDoc. It will be used for autocomplete, because we can preserve BC by allowing users to access our virtual properties via `__get()` and `__set()` methods.
4. We don't need `is()`-ers, because they are equivalent to `(bool) get()`-ers.
5. It actually doesn't matter how we will implement getters and setters (via `__call`, or methods with generics, or independent methods), because we are waiting for Instagram's answer like 100x time longer then processing it later.
6. Do we really need setters?
7. It seems we can skip caching created child objects, because they are pretty cheap.

SteveJobzniak:
Okay so first of all before I reply to the point-list. Here's my benchmarks from last night:

```php
class foo extends \InstagramAPI\AutoPropertyHandler {
    public $bar_mitzvah_test = 'x';
}
$x = new foo();

$start = microtime(true);
for ($i = 0; $i < 2000; ++$i) {
    $x->getBarMitzvahTest();
}
$elapsed = microtime(true) - $start;
$elapsed *= 1000;
var_dump($elapsed);
```

The longer the camelcase name is, the longer it takes to process, but this test (getBarMitzvahTest) is about as long as the longest real property names, which is why I chose that one.

- 1 Million Calls to `getBarMitzvahTest()`: 13.178 seconds
- 10 Thousand Calls to it: 80 milliseconds

Then I thought "Okay, so on average we do less than 2 server API calls/responses per second (in fact, if you want to be safe against throttling, you should wait about 6 seconds between EACH call)". So let's assume 2 calls per second. That's very fast.

Now let's assume each server call returns an `items` array with 100 items.

And let's assume we want to loop through all 100 items and read 10 properties (via getters) from each item.

That's `2 APIs * 100 items * 10 get-calls` = `2000 magic get() calls`.

The cost for 2000 calls?

- 2000 Calls to `getBarMitzvahTest()`: 15 milliseconds

This is why I kept saying it's probably not worth getting rid of the magic methods. Sure they're "super slow" in computer-terms. But in human terms, and in the needs of this project (and at the rate we get data), they're still super fast, much faster than we can get more data from the server. And I see that you came to the same conclusion in your point-list above.

I disliked adding ~4400 lines of code to responses/models with 100 properties, since that's what non-magic methods would mean.

But in general, I really like the idea of doing deferred/lazy parsing of the JSON structures. So I'll now address your point list (well, after I eat breakfast! :D).

SteveJobzniak:
Okay man, the results above still made me think... "What if we cache the expensive function name-to-property name translations?". There's maybe 200 unique property names on Instagram, so it's not going to cost much RAM to cache the ones that people have called.

I therefore implemented a cache, which is on by default but can be turned off.

Before cache (results from post above):

- 1 Million Calls to `getBarMitzvahTest()`: 13.178 seconds
- 10 Thousand Calls to it: 80 milliseconds
- 2000 Calls to `getBarMitzvahTest()`: 15 milliseconds
- 2000 Calls to a shorter property name `getComments()`: 10 milliseconds
- 200 Calls to `getBarMitzvahTest()`: 2.1 milliseconds
- 200 Calls to a shorter property name `getComments()`: 1.3 milliseconds

After cache implementation:

- 1 Million Calls to `getBarMitzvahTest()`: 1.9 seconds
- 10 Thousand Calls to it: 13 milliseconds
- 2000 Calls to `getBarMitzvahTest()`: 2.9 milliseconds
- 2000 Calls to a shorter property name `getComments()`: 2.9 milliseconds
- 200 Calls to `getBarMitzvahTest()`: 0.6 milliseconds
- 200 Calls to a shorter property name `getComments()`: 0.6 milliseconds

Now we really cannot complain about magic method performance anymore! :) Calls now only take ~16% of the time they used to take.

https://github.com/mgp25/Instagram-API/commit/45f31076681034ced74cce0b43eb1f7ab0690a96

SteveJobzniak:
Now I'll answer the point-list! :)

1. Sure. If we derive all response/model classes from AutoPropertyHandler, then the APH constructor can take care of saving the JSON object internally.
2. Yes, it's true that PHP object properties eat up a bit of memory (for a null pointer I guess), but if we get rid of them we get another problem: We must either A) instantly parse the entire JSON tree into a proper object tree so that we can read it directly, or B) cache info about which fields we have lazy-parsed. And if we do A) then we gain nothing compared to JsonMapper, and if we do B) then we gain nothing compared to just having the properties on the object. Hmm. A possible solution is that we make the properties be references to the JSON object data, so "$this->foo = &$this->_jsonData->foo". At least then we don't duplicate the RAM. And we can do isset($this->foo) to know if we need to lazy-parse the JSON data. And also, this system would be an improvement over JSONmapper which already currently duplicates all data in memory.
3. Yep this sounds fine. Using `@property` tag is the exact purpose of that tag.
4. True but they are a convenience for users, so they should be kept. `isEnabled()` is a nice way to read code.
5. Exactly. See the two benchmark posts above. So I think we should use magic getters for all properties and all function calls. Especially after I just implemented the lookup cache which is a huge speedup. As for properties (`->foo`) access technique, they don't need any lookup caching.
6. Yes we need setters sometimes. They're used for things like `$loc => new Location(); $loc->setBlahblah(); $ig->timeline->uploadPhoto($file, ['location' => $loc]);`.
7. If a person repeatedly calls getSomething() and it needs to create an array of Something objects every time, it may be bad for performance. Needs investigating.

Here's a summary of what sounds good so far:

- Property access: Via magic __get and __set
- Function access: Via magic functions.
- Property documentation: Via `@property Type $name` for the class.
- Caching: We will need some type of cache to know what properties we have translated, see point 2 above.
- Reflection at runtime vs at build-time: Will need to be decided.

SteveJobzniak:
I am beginning to wonder if it's worth doing all of this. Benefits vs JsonMapper: Lazy loading of properties (nothing gets parsed until it's used), and reduces RAM usage per final object by something like 25-50%.

Then again if we're worried about RAM we shouldn't be storing the full freaking httpResponse inside of every object. xD We currently store the whole Guzzle response object, which is a maaaaaaassive tree of data which cannot be garbage-colleted as long as that reference remains in the code. But we have it there because we needed it. Sometimes RAM usage doesn't matter as much as code ease of use.

What are your thoughts? After I added the cache to AutoPropertyHandler, it certainly solved all performance issues. Perhaps we should leave the rest as it is? Unless you have some ideas.

valga:
Consider following example:
```php
$response = $instagram->timeline->getTimelineFeed();
$json = $response->getFullResponse();
unset($response);

$mapper = new \JsonMapper();
$mapper->bStrictNullTypes = false;

$mem = memory_get_usage();
$model = $mapper->map($json, new \InstagramAPI\Response\TimelineFeedResponse());
unset($model);
printf("Memory for Reflection data: %d.\n", memory_get_usage() - $mem);

$mem = memory_get_usage();
$model = $mapper->map($json, new \InstagramAPI\Response\TimelineFeedResponse());
printf("Memory for the object (JsonMapper): %d.\n", memory_get_usage() - $mem);
unset($model);

$mem = memory_get_usage();
$model = new Model($json);
printf("Memory for the object (Model): %d.\n", memory_get_usage() - $mem);

$mem = memory_get_usage();
$child = $model->getFeedItems();
printf("Memory for the child object (Model): %d.\n", memory_get_usage() - $mem);
unset($child);
unset($model);

$mem = memory_get_usage();
unset($json);
printf("Memory for the JSON itself: %d.\n", $mem - memory_get_usage());

class Model {
    private $json;
    public function __construct($json)
    {
        $this->json = $json;
    }
    public function getFeedItems()
    {
        return isset($this->json->feed_items) ? new Model($this->json->feed_items) : null;
    }
}
```
Output:
```
Memory for Reflection data: 58056.
Memory for the object (JsonMapper): 44960.
Memory for the object (Model): 56.
Memory for the child object (Model): 56.
Memory for the JSON itself: 55592.
```
So, basically JsonMapper uses 2x memory (or 3x, counting source JSON), because:
1. It caches parsed Reflection data for each property of each object, even if we did not use it all.
2. It creates a copy for each scalar property. I think that PHP is smart enough to increase a ref count on strings, but it still uses a pointer to source one.
3. `null` properties still eat some memory. If we have 100 properties in a model and 1 property in JSON, we have to reserve a memory for 100 properties.

As you can see, Model uses const memory, because we already have fully parsed JSON tree, so PHP just increases a ref count on specific nodes. So basically we need no caching for child objects, because they are already super cheap. But if we want to remain super cheap, we don't need any magic methods that cache something.

To conclude, we have a choice between super cheap models with 3x less memory usage and a clean codebase. But it's 150KB vs 50KB, so I don't know if it worth something.

SteveJobzniak:
Alright, I can confirm that PHP is smart enough to do reference counting for strings. There's a minimum size of the string container on PHP 5.6, which is about 224 bytes, which always exists for any string even if it's a super small string. But that's it. On PHP 7 the cost of copying a string is 0 bytes. Just a reference count.

```php
$mem1 = memory_get_usage();
$string = str_repeat('This is a message which is meant to test string reference counting!', 100);
$mem2 = memory_get_usage();
$string2 = $string;
$mem3 = memory_get_usage();

printf("Characters for the string: %d.\n", strlen($string));
printf("Memory for the string: %d.\n", $mem2-$mem1);
printf("Memory for the 2nd string: %d.\n", $mem3-$mem2);
```

PHP 5.6:

```
Characters for the string: 6700.
Memory for the string: 6992.
Memory for the 2nd string: 224.
```

PHP 7:

```
Characters for the string: 6700.
Memory for the string: 8192.
Memory for the 2nd string: 0.
```

SteveJobzniak:
Thank you so much for the research about memory usage.

I ran the test on my own timeline feed which apparently contains more data (which is good; it's a better test of overhead):

```
Memory for the response object: 128816.
Memory for Reflection data: 110088.
Memory for the object (JsonMapper): 71240.
Memory for the object (Model): 56.
Memory for the child object (Model): 56.
Memory for the JSON itself: 146776.
```

So to sum it up:

- The Guzzle HTTP object + all of our Response Class/Model properties (which are just Jsonmapped links/references to the internal FullResponse data), but NOT including the JSON StdObj (fullResponse) itself (since we keep a reference to it meaning the JSON portion is not GC'd): **128816 bytes**.
- The Jsonmapper reflection cache: 110088 bytes. We can ignore this number because it involves all of the cached reflections of all child-objects in the complex Timeline response. That's a lot of different objects. The memory usage for this cache stays static over time, since each class is only parsed once. So it's not fair to count this reflection cache value in the per-"Response/Model" object statistics. 
- Just the plain, Jsonmapped response object, WITHOUT any values in the Guzzle response and fullResponse properties: **71240 bytes**.
- From the JSONmapped object size above, and the "full response with Guzzle" at the top, we can see the size of Guzzle's object: 128816 bytes (mapped class + Guzzle data, no jsondata) - 71240 (mapped class, no guzzle data, no jsondata) = **57576 bytes** is the size of the Guzzle "httpResponse" data that we include in every Response class object.
- Memory for "just a plain, intelligent Model object technique": 56 bytes per object, recursively created only when requested via getters. Impressive! But then don't forget to add the 58 kB for the Guzzle data and the 146 kB for the JSON data.
- Memory for the JSON StdObj itself: **146776 bytes**.

That gives us these final numbers:

Current method:
- JSON "fullResponse" object: **146776 bytes**.
- Guzzle "httpResponse" object: **57576 bytes**.
- The "Response" class itself: **71240 bytes**.
- Total size: **275592 bytes**.

The "lazy-loading" runtime-parsing proposal:
- JSON "fullResponse" object: **146776 bytes**.
- Guzzle "httpResponse" object: **57576 bytes**.
- The "Response" class itself: **56 bytes** multiplied by how many objects we request/generate, such as "getItems() containing 20 items" being 20x56 bytes extra as well as the cost of the array it allocates. So say on average that we request 20 items + the base object itself. So 21 x 56 = **1176 bytes** is a fair example in most cases.
- Total size: **205528 bytes**.
- Space savings: **-70064 bytes**.

SteveJobzniak:
So we save 25% memory (205kB vs 275kB) by switching from JsonMapper to the Model technique.

But we would save 21% (218kB vs 275kB) just by deleting the Guzzle "httpResponse" value from all responses. And we only use it for checking HTTP status code further down the call stack, I think? So we could get the same savings simply by removing "httpResponse" from ResponseTrait and adding a "httpCode" value instead.

SteveJobzniak:
I decided to measure the WHOLE response object, WITH fullResponse and httpResponse:

```php
$response = $ig->account->getCurrentUser();
$mem = memory_get_usage();
unset($response);
printf("Memory for the response object: %d.\n", $mem - memory_get_usage());
```

Result for the WHOLE object, on a "small data" endpoint:

```
Memory for the response object: 16920.
```

Then I went into Client.php and commented out the line `// $responseObject->setHttpResponse($httpResponse);`

Result:

```
Memory for the response object: 7344.
```

So the biggest saving we can make, immediately and at no loss of code functionality, is to stop storing the Guzzle httpResponse.

I guess I'll do the rewrite unless you have any objection? I think we only use it for httpStatusCode. I'll double-check before I do the rewrite.

SteveJobzniak:
Interesting test (for this one I had to edit ResponseInterface and ResponseTrait to accept a null value for setHttpResponse):

```php
// $response = $ig->timeline->getTimelineFeed();
$response = $ig->account->getCurrentUser();
// $json = $response->getFullResponse();
$mem = memory_get_usage();
$response->setHttpResponse(null);
$memHttp = memory_get_usage();
unset($response);
$totalSize = $mem - memory_get_usage();
$httpSize = $mem - $memHttp;
printf("Memory for the whole API ResponseInterface object: %d.\n", $totalSize);
printf("Memory for the Guzzle getHttpResponse() object: %d.\n", $httpSize);
printf("Memory for the API ResponseInterface without the Guzzle object: %d.\n", $totalSize - $httpSize);
```

Result with getCurrentUser:
```
Memory for the whole API ResponseInterface object: 16920.
Memory for the Guzzle getHttpResponse() object: 9576 (57% of the whole object).
Memory for the API ResponseInterface without the Guzzle object: 7344.
```

Result with timeline feed:
```
Memory for the whole API ResponseInterface object: 271952.
Memory for the Guzzle getHttpResponse() object: 53480 (20% of the whole object).
Memory for the API ResponseInterface without the Guzzle object: 218472.
```

So as I said above, I think we should get rid of the `getHttpResponse/setHttpResponse` on the `ResponseInterface`. All we're using that feature for internally is the Http status code and we can store that as a simple integer and save plenty of RAM.

@valga Can you think of any purpose to keep the whole Guzzle response object in all of our ResponseInterface objects? I don't want to proceed with removing it until you've weighed in since you added this object and perhaps had a special reason for saving the whole thing.

valga:
> The memory usage for this cache stays static over time, since each class is only parsed once. So it's not fair to count this reflection cache value in the per-"Response/Model" object statistics.

Sometimes you have to parse reflection data every time you are making an API request. For example, I am using a queue system to emulate human's behavior (because `sleep()` is not effective), so I am building everything from scratch before making just one API call.

> Interesting test

It is obvious: Guzzle's object holds a response body. It's gzipped, but it still takes some memory.

> I think we should get rid of the `getHttpResponse/setHttpResponse` on the `ResponseInterface`.

I think we are going in a wrong direction. Removing `HttpResponse` from `ResponseInterface` will not reduce peak memory usage (in generic), it just frees some memory a bit earlier.

This PR was about lazy loading and parsing Reflection data just once with some auto-whatever script. But it has a cost of bloating the codebase with a tons of getter and setters, while saving just a bit of CPU power and memory. I've got your position on this, so I will not discuss it further.

SteveJobzniak:
@valga Hi, I didn't see your reply until now.

Anyway, the reason I mentioned Guzzle's `HttpResponse` object on the `ResponseInterface` was that it can be very big. Like 60kb per API call for big endpoints such as timelines, reelstrayfeed, and other big feeds. And because it sits on the response object that users use in their own code, it means that the memory _cannot_ be freed until the user deletes their API function response variable.

So if someone's code does `$feed = $ig->story->getReelsTrayFeed(); foreach ($feed->getItems() as $bleh) { $result = $ig->media->getInfo($bleh->getId()); }` they quickly end up with maybe half a megabyte or more of RAM just taken up by the `HttpResponse` objects. It depends on how they're storing function call responses, and how long they're keeping them in memory.

Simply removing that `HttpResponse` object was able to remove more RAM usage than the "lazy loading" rewrite in many cases.

But really, even in that case we're still talking about pretty small amounts of RAM which only get bloated in very specific scenarios with lots of API call responses stored in user-variables all at the same time.

And I actually think it's nice to be able to get the full `HttpResponse` from any API function call. So there is certainly a value to keeping it.

And as we both agreed, if we did the lazy-loading rewrite, the codebase would be bloated with tons of getter and setter functions that are hard to navigate (when we develop and read through our files with suddenly thousands of lines), while saving only a little bit of CPU and RAM. Especially after I added the AutoPropertyHandler resolver-cache, which means that function calls don't have to be constantly parsed with complicated regexps etc over and over anymore. It speeded up things a lot.

So it all seems pretty fine the way things are. It would be nice if we could solve reduce RAM usage even more, but the gains would be pretty small and the work would be very big, so yeah let's drop the idea. Thanks anyway for the cool ideas! :-)

SteveJobzniak:
@valga I just had an idea! Your core problem with the old system, if I understand it right, is that you needed to have multiple `Instagram()` objects simultaneously, and that you create/destroy those objects on demand, which means that you always have multiple `JsonMapper` class mapping caches that keep needing to be rebuilt, and take up lots of RAM.

How about this solution:

Take away `$this->_mapper = new \JsonMapper();` in Client.php and Mqtt.php.

Replace it with:

```php
class Client {
   protected static $_mapper;
   if (! self::$_mapper instanceof \JsonMapper) {
       self::$_mapper = new \JsonMapper();
       // do base config of it...
   }
}
```

That means a single universal instance will exist for all your copies of `Instagram()` objects.

Will that be an improvement for you? If so you're welcome to make a pull request for that instead.

valga:
The core problem for me is serialization. We have `getFullResponse()` only for root objects, so I can't get an original JSON data for some `$response->getObject()->getList()[0]->getAnotherObject()->getTarget()` object without headache. And serializing some objects cause a lot of `null` values to appear.

The other thing I miss is lazy loading, because I don't consume data on receipt: I serialize and deserialize  it several times. And each non-final step reads just a few properties, or doesn't read them at all.

And your suggestion for making `JsonMapper` will not work for me, because I run queue jobs in isolation.

SteveJobzniak:
Oh okay. You've got a very special setup. I can understand why you'd want smaller objects.

In general, I like a lot of things about this pull request except the tons of lines of code it added.

Perhaps if you change it to just output single-line, undocumented functions:

```php
/**
 * @property Model\Foo $foo
 * @property string $bar
 */
class SomeResponse
{
    function getFoo() { return isset($this->_json['foo']) ? new Model\Foo($this->_json['foo']) : null; }
    function isFoo() { return isset($this->_json['foo']) && (bool) $this->_json['foo']; }
    function setFoo($v) { $this->_json['foo'] = $v->asJSON(); }
    function getBar() { return isset($this->_json['bar']) ? $this->_json['bar'] : null; }
    function isBar() { return isset($this->_json['bar']) && (bool) $this->_json['bar']; }
    function setBar($v) { $this->_json['bar'] = $v; }
}
```

At least then the functions aren't adding 40 lines each. And the object setter above could be solved by ensuring that every object has an `asJSON()` function which just `return $this->_json;`

---

As for the JsonMapper: I added it to the task tracker anyway so that Mqtt and Client can share the same one: https://github.com/mgp25/Instagram-API/projects/4

SteveJobzniak:
Hmm, nevermind. We lose something very important if we get rid of JsonMapper:

`Instagram::$apiDeveloperDebug=true`. It's the only way we can discover when Instagram adds new, unmapped properties. With that flag, we get an exception when the base object has no $property for that JSON value.

We periodically run with that on, to find and add new fields.

Without JsonMapper, we miss that vital feature.

valga:
> We lose something very important if we get rid of JsonMapper:

It's about 10 lines of code in `__construct()`: just call `array_diff_key()` on properties and JSON data. And you can make a nice report tool to dump everything in one pass (vs throwing at first unknown property).

Anyway, this PR has a lowest priority for me at the moment. Just because everything already works.

SteveJobzniak:
That Facebook class structure was so intriguing that I was now inspired to code for fun. I was inspired because their static array of properties idea solves all issues (no need for runtime reflection when we have an array of valid properties!). I love when that happens. Very few things in programming excite me anymore. I've begun working on a proof-of-concept `AutoJsonMapper`, taking ideas from this pull request and Facebook's code, and we'll see how it goes...! I won't replace anything until we both agree on the replacement, since the current code already works. But I have a very solid idea in mind.

About the `array_diff_key()`: Good point. As long as we make it recursively check all depths of the object, it'll work, and will be even better than the current solution since it could report every missing layer as a single throw for the whole response, instead of throwing immediately per-field and requiring us to re-run the test code many times to add each new property.

SteveJobzniak:
Btw, just since this ticket has been acting as a sort of "dump our research here" thread, I'll add something cool: Decoding JSON to array is always smaller than decoding to objects! I was surprised. I assumed arrays would be larger since they always allocate more "index key space" than needed for all current keys, in case the arrays want to grow. I assumed objects only allocated a single key per property and was static. But still, somehow objects are larger. Perhaps objects are actually some kinda specialized arrays in PHP which would explain why they're like array + a little bit more data:

```php
<?php

$json = <<<EOF
{
  "clients": [
    {
      "id": "59761c23b30d971669fb42ff",
      "isActive": true,
      "age": 36,
      "name": "Dunlap Hubbard",
      "gender": "male",
      "company": "CEDWARD",
      "email": "dunlaphubbard@cedward.com",
      "phone": "+1 (890) 543-2508",
      "address": "169 Rutledge Street, Konterra, Northern Mariana Islands, 8551"
    },
    {
      "id": "59761c233d8d0f92a6b0570d",
      "isActive": true,
      "age": 24,
      "name": "Kirsten Sellers",
      "gender": "female",
      "company": "EMERGENT",
      "email": "kirstensellers@emergent.com",
      "phone": "+1 (831) 564-2190",
      "address": "886 Gallatin Place, Fannett, Arkansas, 4656"
    },
    {
      "id": "59761c23fcb6254b1a06dad5",
      "isActive": true,
      "age": 30,
      "name": "Acosta Robbins",
      "gender": "male",
      "company": "ORGANICA",
      "email": "acostarobbins@organica.com",
      "phone": "+1 (882) 441-3367",
      "address": "697 Linden Boulevard, Sattley, Idaho, 1035"
    }
  ]
}
EOF;

$copies = 1000;
$json = '{"deeptest":['.str_repeat($json.',', $copies).$json.']}';

function test($asArray = true) {
    global $json;
    $mem = memory_get_usage();
    $result = json_decode($json, $asArray, 512, JSON_BIGINT_AS_STRING);
    $mem = (memory_get_usage() - $mem);
    printf("%s: %d.\n", ($asArray?'array':'object'), $mem);
    return $mem;
}

$objectMem = test(false);
$arrayMem = test(true);

// Result:
// * 0 Copies:
// - PHP7: object: 5816.  array: 5616.  [object: +200 bytes]
// - PHP5.6: object: 7368.  array: 7120.  [object: +248 bytes]
// * 10 Copies:
// - PHP7: object: 55976.  array: 54176.  [object: +1800 bytes]
// - PHP5.6: object: 74704.  array: 72544.  [object: +2160 bytes]
// * 1000 Copies:
// - PHP7: object: 5050936.  array: 4866160.  [object: +184776 bytes]
// - PHP5.6: object: 6940248.  array: 6551768.  [object: +388480 bytes]

```

valga:
> Decoding JSON to array is always smaller than decoding to objects!

Because object is literally array for properties + class info. But objects with declared properties will take less memory, because there is no need to store names for these properties.

SteveJobzniak:
@valga Ah, so as I suspected then.

By the way my rewrite is going extremely well. I've finished the recursive object hierarchy code, the lazy loading, the getter, is-er and setter. Including full object-support in all of those functions. So people can create a new "someobject", use its setters to set values, then set that object on another object's property. It all works perfectly.

And not a single array value gets stored anywhere if the value is null/missing.

What remains to be done:
- Auto property handling of function calls and direct property access, for backwards compatibility.
- Automatic warnings if missing properties exist in JSON.

In the meantime, here is a teaser:

```php
require __DIR__.'/AutoJsonMapper.php';

$json = <<<EOF
{
  "justAString":"foo",
  "selfObject":{
    "justAString":"foo2"
  },
  "stringArray":[
    "a",
    "b",
    "c"
  ],
  "selfArray":[
    {
      "justAString":"foo2"
    },
    {
      "justAString":"foo2"
    },
    {
      "justAString":"foo2"
    }
  ]
}
EOF;

class Test extends AutoJsonMapper
{
    protected static $_jsonProperties = [
        'justAString' => ['string', false],
        'selfObject'  => ['\Test', false],
        'stringArray' => ['string', true],
        'selfArray'   => ['\Test', true],
    ];
}

$data = json_decode($json, true, 512, JSON_BIGINT_AS_STRING);

$x = new Test($data);

var_dump($x);
var_dump($x->getField('justAString'));
var_dump($x->isField('justAString'));
$x->setField('justAString', 'test');
var_dump($x->getField('justAString'));
var_dump($x);
var_dump($x->getField('selfObject'));
var_dump($x->getField('selfObject')->getField('justAString'));
var_dump($x->getField('stringArray'));
var_dump($x->getField('selfArray'));
$foo = new Test();
$x->setField('selfObject', $foo);
var_dump($x->getField('selfObject'));
var_dump($x->getField('selfObject')->getField('justAString'));
$foo = new Test(['justAString' => 'example']);
$x->setField('selfObject', $foo);
var_dump($x->getField('selfObject'));
var_dump($x->getField('selfObject')->getField('justAString'));
```

Result:

```
$ php test.php 
object(Test)#1 (1) {
  ["_jsonData":protected]=>
  array(4) {
    ["justAString"]=>
    string(3) "foo"
    ["selfObject"]=>
    array(1) {
      ["justAString"]=>
      string(4) "foo2"
    }
    ["stringArray"]=>
    array(3) {
      [0]=>
      string(1) "a"
      [1]=>
      string(1) "b"
      [2]=>
      string(1) "c"
    }
    ["selfArray"]=>
    array(3) {
      [0]=>
      array(1) {
        ["justAString"]=>
        string(4) "foo2"
      }
      [1]=>
      array(1) {
        ["justAString"]=>
        string(4) "foo2"
      }
      [2]=>
      array(1) {
        ["justAString"]=>
        string(4) "foo2"
      }
    }
  }
}
string(3) "foo"
bool(true)
string(4) "test"
object(Test)#1 (1) {
  ["_jsonData":protected]=>
  array(4) {
    ["justAString"]=>
    string(4) "test"
    ["selfObject"]=>
    array(1) {
      ["justAString"]=>
      string(4) "foo2"
    }
    ["stringArray"]=>
    array(3) {
      [0]=>
      string(1) "a"
      [1]=>
      string(1) "b"
      [2]=>
      string(1) "c"
    }
    ["selfArray"]=>
    array(3) {
      [0]=>
      array(1) {
        ["justAString"]=>
        string(4) "foo2"
      }
      [1]=>
      array(1) {
        ["justAString"]=>
        string(4) "foo2"
      }
      [2]=>
      array(1) {
        ["justAString"]=>
        string(4) "foo2"
      }
    }
  }
}
object(Test)#2 (1) {
  ["_jsonData":protected]=>
  array(1) {
    ["justAString"]=>
    string(4) "foo2"
  }
}
string(4) "foo2"
array(3) {
  [0]=>
  string(1) "a"
  [1]=>
  string(1) "b"
  [2]=>
  string(1) "c"
}
array(3) {
  [0]=>
  object(Test)#2 (1) {
    ["_jsonData":protected]=>
    array(1) {
      ["justAString"]=>
      string(4) "foo2"
    }
  }
  [1]=>
  object(Test)#3 (1) {
    ["_jsonData":protected]=>
    array(1) {
      ["justAString"]=>
      string(4) "foo2"
    }
  }
  [2]=>
  object(Test)#4 (1) {
    ["_jsonData":protected]=>
    array(1) {
      ["justAString"]=>
      string(4) "foo2"
    }
  }
}
object(Test)#3 (1) {
  ["_jsonData":protected]=>
  array(0) {
  }
}
NULL
object(Test)#4 (1) {
  ["_jsonData":protected]=>
  array(1) {
    ["justAString"]=>
    string(7) "example"
  }
}
string(7) "example"
```

SteveJobzniak:
As you can see, definition of a class is super easy:

```php
class Test extends AutoJsonMapper
{
    protected static $_jsonProperties = [
        'justAString' => ['string', false],
        'selfObject'  => ['\Test', false],
        'stringArray' => ['string', true],
        'selfArray'   => ['\Test', true],
    ];
}
```

If it's a native type, no casting of the value is done. It's just for our own/the user's own internal hinting.

If it's an object-type (starting with `\`), it assumes the object is an instance of AutoJsonMapper.

The true/false just sets whether the object is an array or not. I didn't want to use `\Foo[]` notation, to avoid having to constantly look for `[]` at runtime, and stripping that part of the string, etc... It's just more efficient to define properties as an array of "type+isArray".

Right now, if you look at the code above, the get/set/is-functions are exposed publicly. But they will be protected soon, when I re-implement magic methods and magic property access. Because those are definitely 100% going to be used. There's no way we're going to have a ton of functions in each class-file.

After my earlier cache-implementation (a few weeks ago), performance of the magic methods doesn't matter anymore. Lookups of function names magically is super-fast via the cache.

SteveJobzniak:
Okay... so far I've finished the recursive object hierarchy code, the lazy loading, the getter, is-er and setter. Including full support for arrays and objects. And auto-method handlers (get/set/is-functions). And direct property access (getter, setter, isset(), empty() and unset() are all supported on the virtual properties). And all class documentation is done.

This is what remains to be done:

- A constructor parameter for recursively checking the entire JSON data to ensure that all JSON properties in the data are defined in all classes.
- Converting all classes to this new `AutoJsonMapper` (I'll do that as a separate commit, because it's going to need manual rewriting of tons of classes, and all of their current overridden class functions like `getMessage()` that assume normal `$properties`).
- A new php-doc generator for classes that derive from `AutoJsonMapper` (perhaps a task for you).
- We'll still keep `AutoPropertyHandler` and the old doc-generator just in case we need them for anything in the future. They don't hurt us by existing there.

SteveJobzniak:
Okay... almost done!

Now the recursive "require JSON properties" checking is implemented:

```php
$json = <<<EOF
{
"self":{
  "yet_another_missing":"000"
},
"selfArray":[
  {
    "array_missing":"111"
  },
  {
    "array_missing":"222",
    "another_array_missing":"333"
  }
],
"property":"123",
"missing_property":"456",
"another_missing":"789"
}
EOF;

class TestUndefined extends \AutoJsonMapper
{
    protected static $_jsonProperties = [
        'self'      => ['\foo\TestUndefined', false],
        'selfArray' => ['\foo\TestUndefined', true],
        'property'  => ['string', false],
    ];
}

$data = json_decode($json, true, 512, JSON_BIGINT_AS_STRING);
$x = new TestUndefined($data, true);
```

`Fatal error: Uncaught RuntimeException: Missing JSON property definitions in "foo\TestUndefined": ("another_array_missing", "another_missing", "array_missing", "missing_property", "yet_another_missing").`

Phew... Okay, so the class is nearly complete and supports all features that the old `AutoPropertyHandler + JsonMapper` setup had. All that needs to be done now before I upload the class is to add support for non-typed properties to this new class. Since our responses/models have so many undefined variables.

SteveJobzniak:
Okay, it's done. As usual with programming, it took 30 minutes for a quick "proof of concept", and then it took 10 hours to perfect it and think of every scenario and every possible good/bad value that we need to check against.

I'm going to upload the class in a moment. It's literally _perfect_ as far as all of our requirements (low RAM usage, no runtime reflection, no runtime reflection cache, supports getters and direct property access, supports all type-conversions perfectly).

But... There's one very big problem with it... Which I unfortunately JUST thought about after spending 10 hours non-stop coding...

Because of lazy loading, there's ABSOLUTELY NO way to do this anymore: `$item->getUser()->setUsername('foo')`... Because every getter for objects/arrays creates a new object/array.

There's only one way to solve that: Implement a "resolved objects"-cache of the entire property-tree on the root object, so that they're only parsed a single time on the _root_ object. Like JsonMapper. But with this difference: We only cache properties that have been touched (set via setter or retrieved via getter).

So in the scenario above, the object would be something like this in memory:

```
class Foo {
   "_rootObject" => $this,
   "_resolvedData" => [ // Only has a value on root Object?
      "user" => class User() // this can probably use lazy-loading...
   ],
   "_jsonData" => [
      "user" => ...// original data,
      "other_field" => ...,
      "etc" => ...,
   ]
```

I don't know... this was a very unfortunate and big problem to discover after 10 hours of work!

Anyway, I'll upload the class now...

SteveJobzniak:
Okay @valga, the class is up... and ready for your eyes and your review.

https://github.com/mgp25/Instagram-API/blob/master/src/AutoJsonMapper.php

Perhaps you have some idea for how to solve the function chaining issue.

SteveJobzniak:
I have an idea for a great solution...

I'll simply rename `$_jsonData` to `$_internalData`, and initialize it to the JSON data array... But then when people use the getters, I'll automatically translate SECTIONS of the array into objects ON-DEMAND. Like this:

if `$type` of the requested property is an OBJECT, check if the `$value` is instanceof (`is_a()`) the requested class that it should be. If not, we replace that object with an instance of that class.

So imagine this init tree of pure JSON `$_internalData` array:

```
rootobject::$_internalData = [
   "subobject":[
      "foo":true,
      "deepersubobject":[
          "bar":true
      ]
   ]
]
```

Now the user does `$sub = $x->getSubObject()`, so we check it and see that it's NOT yet an instance of the requested class, so we replace THAT FIELD of the root-object with a new class instance which just instantly swallows the rest of the sub-array without any further recursive parsing:

```
rootobject::$_internalData = [
   "subobject":class SubObject() {
       $_internalData = [
          "foo":true,
          "deepersubobject":[
              "bar":true
          ]
       ];
    }
   ]
]
```

Now the user does `$sub->getDeeperSubObject()`, to get the final object, and that is what we replace the internal contents of `class SubObject()` with, which gives us the final object hierarchy:

```
rootobject::$_internalData = [
   "subobject":class SubObject() {
       $_internalData = [
          "foo":true,
          "deepersubobject":class DeeperSubObject() {
              $_internalData = [
                  "bar":true
              ]
          }
       ];
    }
   ]
]
```

As for non-object properties, those will simply be cast (via settype(), almost zero-cost) every time to guarantee that they're turned into and returned as the proper type no matter what they're stored as.

This will need some restructuring of the class... so... tomorrow...

SteveJobzniak:
Yep, I did a quick test of the lazy-tree conversion method and it solves everything. I'll finish that new method tomorrow and upload it asap.

It has been solved. You don't have to worry about the wall of text above. ;-)

valga:
Sorry, but I have no time and energy to review you class at the moment. Feel free to use it instead of `JsonMapper` if it has the lazy loading and nice serialization stuff. I completely trust you on this.

SteveJobzniak:
@valga Understood.

Anyway, I worked on it all day yesterday too. The class has been rewritten again (not uploaded yet). To solve all the million little things and tighten every scenario. Because it turns out there are a million things that need to be solved when you actually want to use a class in production. Such as supporting inheritance (inheriting the classes from all `A extends B` parents in the class-hierarchy), and supporting multiple inheritance (importing classes into itself) with full protection against circular references (the two objects are allowed to share the same ancestors (extends), and can even be part of diverging branches, but are not allowed to import each other). I also added things like strong verification of the entire property map (for validity), and changed the definition format to PHPdoc-style (`string[]`), and added support for multilevel arrays of the type (`\Foo[][]` is array of arrays of Foo), and added support for accessing properties that are not even defined in the class (that way people can always get them). And I refactored much of the code.

And now you added another to the list: Custom serialize/unserialize functions are needed. Thanks for discovering that now. I have an idea how to solve that.

I will release this class as my own separate project repository instead, because it has grown to be a JsonMapper killer and people will want to use it, so if I don’t make a repository then someone will just rip it out and upload it somewhere and claim credit for it. Besides, a repo is good for allowing people to make pull requests for it.

PR #1536, opened 2017-09-04
Realtime/Push: Fix reconnection timers

valga:
Recent refactoring was so huge, and I missed a couple of things with reconnection timer. This PR fixes it.


SteveJobzniak:
Thanks, nice catch! I just woke up but I'll comment on the Jsonmapper PR soon. :)

Issue #1539, opened 2017-09-05
Set location when uploading a photo.


skn2215:
Hi, please tell me why geolocation is not attached in the photo?

```php
var_dump($location);

object(InstagramAPI\Response\Model\Location)#145 (19) {
  ["name"]=>
  string(13) "Paris, France"
  ["external_id_source"]=>
  NULL
  ["external_source"]=>
  string(15) "facebook_places"
  ["address"]=>
  string(0) ""
  ["lat"]=>
  float(48.8567)
  ["lng"]=>
  float(2.3508)
  ["external_id"]=>
  NULL
  ["facebook_places_id"]=>
  string(15) "110774245616525"
  ["city"]=>
  string(0) ""
  ["pk"]=>
  string(7) "6889842"
  ["short_name"]=>
  string(5) "Paris"
  ["facebook_events_id"]=>
  NULL
  ["start_time"]=>
  NULL
  ["end_time"]=>
  NULL
  ["location_dict"]=>
  NULL
  ["type"]=>
  NULL
  ["profile_pic_url"]=>
  NULL
  ["profile_pic_username"]=>
  NULL
  ["time_granularity"]=>
  NULL
}

```

```php
$metadata['location'] = $location;
$ig->timeline->uploadPhoto($photo, $metadata);
```

mgp25:
@serega2215 can you post your debug log (request and response).

serega2215:
Oh sure

[log.txt](https://github.com/mgp25/Instagram-API/files/1277818/log.txt)

mgp25:
@skn2215 we have found the issue, we are updating this in a couple of minutes.

Regards

SteveJobzniak:
This has now been fixed, but not in the way that you may have expected.

1. We were ALREADY properly generating the `facebook_places_id` POST parameter. When we were given a PROPER Location object.
2. You were using the `Location::searchFacebook` and `Location::searchFacebookByPoint` functions to find your locations. That's the wrong way.
3. We were not clear enough about telling users that those functions are WRONG, so I have now added documentation to all 3 location-search functions, telling people which one to use for media.
4. The library now throws an `InvalidArgumentException` if you try giving it a Facebook location object in the future.

Here's the code changes we did:

https://github.com/mgp25/Instagram-API/commit/5fa36a5a32a43cd34ed6ddcedd41a04723f905f9

[...]

PR #1541, opened 2017-09-06
fix: set geolocation

skn2215:


SteveJobzniak:
Thanks a lot. We'll have to verify this change, and we also need to change it in all the other code places that send locations.

Interestingly, the old code for adding locations on uploads was last tested by me a month or two ago and worked then. So perhaps they recently changed it.

mgp25:
Verifying. Give me 5-10 minutes

SteveJobzniak:
@mgp25 Great. My biggest concern is that I'm not sure that all locations have a FacebookPlacesId value. Non-Facebook based locations don't havr any, if I remember correctly.

mgp25:
@SteveJobzniak gitter

SteveJobzniak:
Okay, we've verified. This change is not completely correct and only fixes 1 of 4 code locations. We'll code the fix, but thank you very much for finding the problem!

mgp25:
@skn2215 i have closed your PR, because the fix require multiple changes. We'll be adding it in the next couple minutes. 


Regards

SteveJobzniak:
We found the real problem. I'll answer in your ticket: https://github.com/mgp25/Instagram-API/issues/1539

PR #1564, opened 2017-09-15
Ability to read parameters of resulting image after process MediaAutoResizer

safhome:
When I post horizontal & vertical images as album in timeline, i've found that Instagram's web-interface shows them incorrectly: it's stretching second and other images to aspect ratio of first one. So I decide to set all my images to one format: autoresize first and then for other sets aspect ratio of first image. But to do so I need to get resulting aspect ratio. So I realized ability to read parameters of resulting image after process MediaAutoResizer.
Another added feature is "autoAdjustAspectRatio" to automatically adjust values of minAspectRatio and maxAspectRatio into allowed interval. This function is enabled by default but someone who wants to get error rather then right image can switch it off.

Swapping constants and variables in all comparisons is result of "composer codestyle".


SteveJobzniak:
Thanks, I will investigate and think about this when I have more energy, and possibly implement the idea in another way. All of the swapped if-statements is making it hard to read the actual changes. I'm surprised codestyle did that, since it isn't doing that on my machine. We may have inherited a new rule from the core `@Symfony` rules we base on. Heh. I'll check tht too.

safhome:
Instagram forces all album photos to have identical aspect ratios only in web-interface, same album in android application looks OK with images with different aspect ratios.

SteveJobzniak:
@safhome Hi, thanks for the extra details. I haven't researched this yet, but would just like to add that today I saw an album uploaded from the app by someone. One photo was 717x717 and another was 900x900, so both were 1:1 aspect ratio but totally different resolution. It seems like you're right that the official app crops all album content to be the same ratio when an album is created, but it definitely doesn't care about matching resolution. So you only need to care about aspect.

If you or someone else wants to check for sure what the app does, you could try making an album with a portrait and a landscape photo from your camera roll, and see what Instagram does to the final media's aspect ratios. Could be interesting.

SteveJobzniak:
@mgp25 Okay, there's a problem with `composer codestyle` now. If you update to the latest version of the php-cs-fixer, it will swap all code, like `if ($result[$i] === ' *') {` becomes `if (' *' === $result[$i]) {`. So don't run it/commit anything until we've figured out what rule we must tweak in the codestyle config.

@valga I think we probably inherited this from some update to the `@Symfony` codestyle settings, but I think it's still probably a good idea to inherit from them since we'll automatically get all good evolutions to their codestyle settings. This particular change was bad, but we'll just find it and fix it... :-)

SteveJobzniak:
I think I found the problematic rule:

```
yoda_style [@Symfony]

Write conditions in Yoda style or not based on configuration.

Configuration options:

equal (bool, null): change equal (==, !=) statements; defaults to true
identical (bool, null): change identical (===, !==) statements; defaults to true
less_and_greater (bool, null): change less and greater than (<, <=, >, >=) statements; defaults to null
```

Haven't checked yet. Going to bed.

valga:
I hope that `yoda_style` will not be included into recommended StyleCI ruleset.

SteveJobzniak:
Hey guys. I've released v3.1 with some various fixes and a fix for the "composer codestyle" command.

As for this pull request: I'll analyze it when I have time and energy. I can't promise when. The idea of getting the dimensions/aspect of the final image is good, but it needs to be written in a way that isn't so invasive into the class. Thinking about it quickly, perhaps something like a separate function, `getStatistics()` which returns `null` if getFile() hasn't been called, or otherwise returns an array of stuff like `['is_original' => false, 'width' => 1000, 'height' => 1000, 'aspect_ratio' => 1.0000000]`.

And @valga, if StyleCI starts doing something like that, they would make it configurable, so there's no danger. :-)

SteveJobzniak:
I have had a discussion with mgp25 tonight. The API is complete and up to date, and we've been working on it non-stop for so long. All of our instant support and instant coding and instant feature additions etc is totally killing us. So we'll slow down a bit now that v3 is out and the API is totally stable.

We're simply going to change the pace around here a bit. These tickets will remain open for a while, and we'll take care of them when we _feel_ like relaxing and doing some coding. Any ultra-critical bugfixes will of course still get high priority, but minor things or new features will happen when time is available.

This particular ticket will take some time to get through. Don't worry, it _will_ get done when we feel like relaxing with some programming, and we don't need to be reminded. This change of pace is simply for our health's sake. We've worked so hard for so long. Thank you very much for your contribution!

SteveJobzniak:
Hi. I ran the fixed `composer codestyle` and committed it directly to this pull request, so that we can look at just the actual changes:

https://github.com/mgp25/Instagram-API/pull/1564/files

Now I can review the changes:

- `autoAdjustAspectRatio` parameter: Hmm. In the current code, we will throw an exception if the user gives the class a bad custom aspect ratio. With this new parameter, you make it possible to silently ignore any bad user-provided min/max ratios and instead auto-select the lowest/highest valid ratio. But I don't see any reason why? Why wouldn't you want the class to actually warn you if your code gives it a bad parameter? It's up to the user of the class to pass in a proper parameter value, so unless you can explain a good use-case, I'll vote 👎  for this new parameter proposal.
- The other proposal about being able to get statistics for the final file: 👍 Yes, that's a good idea. But as I said above, it won't be implemented with all of these new variables and functions. We'll add a `MediaAutoResizer::getWithStatistics()` function which returns an array with the file AND all statistics about it.

Here's the todo list for the next stable release. We've added `getWithStatistics()` to that goal:

https://github.com/mgp25/Instagram-API/projects/3?

Thanks for the idea, and you can look forward to seeing it added in the next release (whenever it's ready...). :-)

SteveJobzniak:
@safhome Okay the soon-to-be-merged pull request from @valga solves this.

https://github.com/mgp25/Instagram-API/pull/1660

Instead of some kinda `getWithStatistics()`, you will simply do this:

1. Step 1 is to resize the first photo/video of the album via a resizer, and getting its `getFile()`.
2. Step 2 is to do a `new PhotoDetails($thatFile)` or `new VideoDetails($thatFile)`.
3. Step 3 is to use the `$details->getAspectRatio()` as the minimum and maximum ratio in your further resizings. And you should also use the `allowNewAspectDeviation = true` option.

PR #1567, opened 2017-09-17
Realtime/Push: Fixes and improvements

valga:
This PR fixes a couple of issues with Realtime client. Also it forces Realtime and Push clients to run only from CLI.


SteveJobzniak:
This is perfect. Thank you for being an awesome perfectionist. 👍

PR #1569, opened 2017-09-18
Add ImageAutoResizer to the uploadPhoto example

javilumbrales:
Hi,

I thought adding this to the example would be very handy for newcomers...

Cheers,
J


SteveJobzniak:
Hi. That's a great idea! The class is definitely too hidden right now, so adding it to the examples is great. But you wrote it for v2.x (well, in your original commit), and didn't include any documentation, so it would just confuse users. Furthermore, we recently (with v3.x) added support for Story processing, which is actually the upload-type that's in the most need of the autoresizer (because every story must be transformed into ~9:16 ratio to look right in the app).

But the idea is great, so I've written documented examples in the uploadPhoto and uploadStory examples:

https://github.com/mgp25/Instagram-API/commit/6683a6d71705278359501590426268419bbdd2fc

Thanks again for the suggestion! :)

javilumbrales:
You are absolutely right, and what you've done is very very useful! Thank you!

PR #1572, opened 2017-09-19
Basic video resize keeping aspect ratio

javilumbrales:
This is far from perfect, and I'm sure it could reuse a lot more from the `MediaAutoResizer` (that would require refactoring of that class which I'm not familiarised enough to do), but so far it does the bare minimum to allow uploading videos that have a higher resolution.


SteveJobzniak:
This is really cool. We were actually going to add video support to MediaAutoResizer and make it auto-detect the format, to detect if the input is a video or an image. Because the algorithms for cropping and expansion in that class are already universal, and the class simply needs a few steps to detect if the input is a video, and to create an ffmpeg command for doing cropping or expansion in that case.

Your code here will be a really great reference during the MediaAutoResizer refactoring. I know that I wrote down an ffmpeg cropping command somewhere in these tickets and pull requests. A search for "ffmpeg crop involves:stevejobzniak" would find it.

We won't merge this separate class, since we would have to delete it (and break backwards compatibility) later. I also don't like that it's unfinished and a bit weird (such as inheriting froma nother class but not calling the parent constructor or any of its functions). Perhaps you'd like to work on the MediaAutoResizer refactoring instead? I know it's a pretty big class, but it's actually not that difficult to understand. The constructor sets options, then there is a function which checks if processing will be necessary, then a function which calculates crop/expand, and finally a function which processes the image.

- The class needs new constants for min/max video width.
- It needs to detect input file format, possibly based on file extension. Something like "if jpg, jpeg, gif, png or bmp, treat as image, otherwise treat as video".
- The "needs processing?" function needs to also handle video detail reading, instead of just image dimensions.
- The processing code needs to be refactored so that it calculates the new dimensions and crop-offset/expand parameters, and then calls a final function which does either a video command or does image processing.
- ffmpeg equivalents for cropping, expansion and crop-shifting need to be implemented. I think ffmpeg has commands for all of that.
- All of the isrotated code needs to be ignored if it is a video, since those are never based on rotated pixel data, so simply always set isrotated to false if it is a video.

If you are interested in trying this on, then we would really appreciate your help because we're overworked after months of non-stop work, and have decided to slow down for a while to rest.

If not, then we'll simply leave this ticket open and use it as a reference when we have time and energy and feel like working on the MediaAutoResizer refactoring.

Speaking of which, @valga: You're probably overworked and tired too. But in case you're not and you feel like coding for fun, then you also know the MediaAutoResizer intricately and would be a great mind for the video resizing support.

Either way, it will get done eventually. I just severely need to rest and recover my health right now. ;-)

valga:
@SteveJobzniak See #1574. It has the same set of features, such as crop&expand (even with custom color).

javilumbrales:
wow! that is way more advanced than what I've done! I think my PR doesn't make any sense now, I just created it for myself and thought of sharing it and perhaps improve it together with some help, but looking at yours mine is a no go.

@SteveJobzniak I completely appreciate all that you've said and all makes sense. The only reason I created a separated class was mainly to avoid likely breaking the existing one and because the way I see it Images are one thing and Videos are another one (SRP), and there's of course stuff in common. Pretty much what @valga has done with the `ResizerInterface` is what I had in mind, but I was still miles ahead :).

Anyway, thank you very much both for your awesome work. In any case your health is and always will be more important, I hope you get well soon.

Best regards,
J

SteveJobzniak:
@valga Ohh. You have a brilliant mind as always! <3 I'll speak to you in that ticket in a moment.

@javilumbrales Hehe no problem, your pull request was very appreciated anyway and gave us the "kick" to put higher priority on the video extension of `MediaAutoResizer`. By the way, his new code isn't used via `ResizerInterface`. The main class is still `MediaAutoResizer`, and it auto-chooses what type of resizer to use based on if it's an image or video input file. So it's just a single, easy to use class for any filetype, as intended. ;-)

SteveJobzniak:
@javilumbrales I have to say something else: I was impressed with your codestyle and ability. I would love to see you around, because you have great potential, and after a while of testing perhaps you'll even want to join the team if you're interested then. We're 3 people right now and always need skilled programmers. In that case I would suggest trying your hand at some of the open tickets on the Issues tab. And if something perhaps "isn't perfect" in a pull request, we can always talk through it to improve it. Everyone is new to a project's codebase at some point. ;-)

SteveJobzniak:
@javilumbrales If you're using `dev-master`, the new MediaAutoResizer with video support is there now, but it's a work in progress (some stuff is not right yet; we need to make it check video format validity etc too). So it won't be officially released until that's done. If you need the feature already,  I suggest that you switch to `dev-master` instead of stable.

javilumbrales:
@SteveJobzniak thank you very much for your kind words, far more from what I deserve. This is definitely a very challenging and exciting project, and that's why I got encouraged to raise this PRs, trying to collaborate and contribute to it, even if only with tiny bits. Having said that I work professionally as full time dev and unfortunately that doesn't tend to allow me much free time to work on "the cool stuff". As much as I'd love to, specially the reverse-engineering stuff looks like something I'd really enjoy, but I'm afraid I can't commit/promise anything at the moment, as chances are I wouldn't be able to deliver it and would end up being a disapointment for everyone. 

Nonetheless, I'm keeping an eye on you guys and whenever I have the chance to contribute in any way, I'll definitely try my best to help you out. You are doing a really good job here and I really appreciate it.

Re video support, thank for letting me know, I'll give that a go.

Best regards,
J

SteveJobzniak:
@javilumbrales Ahh, you work as a full-time developer. That explains your code quality! :-) You would have made a really great addition to the team. But I completely understand the time limitations. It's okay. :-)

Thanks again for submitting this pull request. If you have any other great ideas or change your mind, you know where we are. 😄  See you around!

PR #1574, opened 2017-09-19
MediaAutoResizer: Support videos

valga:
Work in progress...


SteveJobzniak:
Wow, you know what? This is absolutely fantastic! The separation into separate classes and interfaces really cleans up the code. I especially love the new Rectangle and Dimensions classes. Those sure cleaned up a lot of code locations and made them much clearer!

I was just on my way to bed but I look forward to reading this in depth later.

Here are some initial things to check out:

- The `mime_content_type` call is risky. The user must have a `magic.mime` file in their PHP installation. So instead, perhaps we should do something like what's in a few user comments here: http://php.net/manual/en/function.mime-content-type.php. All kinds of people are making more universally reliable mime checkers themselves, by downloading the Apache mime list and writing a converter which turns their list into a PHP array (which can be included as a regular PHP class file). We could do that as a `devtool/generateMagicMime.php`. And then, have our own mime checker which looks at the file-extension and compares it to the mime list.

- I noticed that the video processor right now doesn't have any conditions for when to process. It always returns false. I guess we should tweak it to say "yes, we need processing" if the video isn't mp4 + h264 + aac (I think those are Instagram's video file format rules, but that needs to be researched).

Thanks a lot for putting your brilliant mind to this. :-)

valga:
> The `mime_content_type` call seems like it may be risky.

According to [docs](http://php.net/manual/en/fileinfo.installation.php), Fileinfo extension is built in as of PHP 5.3. But I made a couple of checks to ensure that: `mime_content_type()` function exists and returns valid value, with fall back to extension-based detector.

> I noticed that the video processor right now doesn't have any conditions for when to process.

Yep, rotated videos aren't supported too. Both of these features are good candidates to another PR.

> This is an issue for later: There are places in our regular code that refers to `MediaAutoResizer::SOMECONSTANT`.

I believe I fixed them all.

SteveJobzniak:
All of the newly added commits above look correct. Awesome! 👍 

Okay here is my full review:

- The fallback to regex extension checker: Looks good since it covers all common extensions.
- `Utils::getVideoFileDetails()`: It's awesome that you discovered that the ffprobe duration sometimes exists in the "format" probe result instead of in a stream. But the rewrite has a bug. In the old code, finding at least 1 video stream meant that we successfully (well, we thought) had the "width, height and duration". With the new code, we skip "duration" if it's not available in the video stream. But then we also skip it if it's not available in the "format" section. So the newly added code at the bottom of the function needs to be fixed a bit, to something like: `"if we found no duration, then { if format-section has duration, use that; else: throw runtimeexception that we cannot find any duration for this video }"`.
- Rotated videos: Can that really happen? The whole reason for "isRotated" flag was because JPEG data is weird and can be stored as 100x200 but have a special EXIF flag that tells the viewer to render it as 200x100. So the only case where videos can be rotated is if they do the same thing: ffprobe reporting 100x200 but actually rendering as 200x100. If videos can indeed be rotated (or if you don't know for sure) then please edit `VideoResizer` to put a developer reminder-comment `// FIXME: Research and implement how to handle rotated videos.`
- Bad video formats: Yes, a different PR/later work by any of us is fine. But please put `// FIXME: Ensure that video is mp4 + h264 + aac` in VideoResizer's "needsProcessing?" function, so that we remember that it's work that needs to be done later.
- In VideoAutoResizer, please also put a developer reminder/comment above `'%s -y -i %s -vf %s -c:a copy -f mp4 %s 2>&1',`, to say `// FIXME: Force to h264 + aac audio, but if audio input is already aac then use "copy" for lossless audio processing. Video format can't copy since we always need to re-encode due to video filtering.` (that comment is long but will only be needed until the ffmpeg command has been finalized).
- The `InstagramAPI\Media` namespace: It sounds like a **very** nice and logical namespace for **all** of our internal media-handling stuff. So that we have all resizing, "get media info" etc classes in a single place. So perhaps we should move the `src/Request/Metadata/` namespace Media/Photo/VideoDetails classes to the new Media namespace? What do you think of that idea?
- The `rotate()` functions you recently added above don't make sense name-wise. It's not rotating (a rotation would act in 0-360 degrees). They're actually swapping dimensions so that width becomes height, and height becomes width, and X becomes Y and Y becomes X. And they can swap back and forth repeatedly. So I'm trying to think of a more logical name, perhaps: `Dimensions::swapDimensions()` and `Rectangle::swapRectangle()` (but the latter sounds confusing since it swaps the rectangle and the input x/y). Or perhaps just `::swap()` on both. Or `::swapValues()` on both (I think that last one is my favorite, just because it's so clear about what it does to the values, and it can be the same function name on both classes).
- Lastly, something for improving code clarity: Currently we have `$this->_width` and `$this->height` and `$this->_aspectRatio` in `MediaAutoResizer`, which is very confusing since people can't easily know if they refer to target or input dimensions without reading all of the code. One way to handle that would be to rename them to `$this->_inputWidth`, `$this->_inputHeight` and `$this->_inputAspectRatio`. Another even better way to handle it would be to make it all into objects: `$this->_inputDimensions = new Dimensions($this->_resizer->getMediaDetails()->getWidth(), $this->_resizer->getMediaDetails()->getHeight()); $this->_inputAspectRatio = $this->_inputDimensions->getWidth() / $this->_inputDimensions->getHeight();`, and then in the "is rotated?" section further down in MediaAutoResizer we can just do `if ($this->_resizer->isRotated()) { $this->_inputDimensions->swapValues();  $this->_inputAspectRatio = 1 / $this->_inputAspectRatio; }`.

Wow... long review is long... but I'm very excited about this new media resizer rewrite. It's excellent. :-)

valga:
> Rotated videos: Can that really happen?

Yes, videos could have [a metadata with rotation flag](https://addpipe.com/blog/mp4-rotation-metadata-in-mobile-video-files/). ffmpeg already has some [autorotation stuff enabled by default](https://trac.ffmpeg.org/ticket/515), so we should check how it works.

> So perhaps we should move the `src/Request/Metadata/` namespace Media/Photo/VideoDetails classes to the new Media namespace?

I think it is some kind of breaking change, so it should be avoided.

Other things are addressed.

SteveJobzniak:
Wow. Fantastic work!

That new rotation-refactoring is excellent. I love that MediaAutoResizer doesn't have to care about rotation in its calculations anymore, and that it's all handled in the internal `_processResource` which just swaps the dimensions/rects it was given. That's so clean. I love it!

Just to be sure there are no bugs after all this refactoring (especially in the crop-calculation isRotated + isVerFlipped/isHorFlipped cropFocus code), I'll run all 16 of the JPEG orientation test files (https://github.com/recurser/exif-orientation-examples) through it before merging, with various cropFocus tests. If they all give identical results, then it works:

And about this:

> > So perhaps we should move the `src/Request/Metadata/` namespace Media/Photo/VideoDetails classes to the new Media namespace?

> I think it is some kind of breaking change, so it should be avoided.

It doesn't matter. The users have never been told about our internal Metadata classes, so it doesn't matter if we break backwards compatibility for those because they are not part of our public user-facing API.

SteveJobzniak:
Okay, @valga my dear friend! All image tests succeeded! ❤️ I'll merge this now and also add a little devtool I created for easily testing all JPEG orientations to make sure they're all correctly processed.

Thanks a lot for doing this amazing rewrite. Seriously amazing.

As for the `Internal\Metadata` namespace thing: I'll do the change after this merge. Because these media classes belong together, and they're not for public use by users. :-)

SteveJobzniak:
Okay I've been carefully reading through every line and documenting and tweaking a few things where necessary so that the class is even cleaner, so here's a 2nd review:

https://github.com/mgp25/Instagram-API/commit/6b4e14b3ad35928ab4b001afa0351af132ffea2a

I did notice a bad habit. Sometimes you document functions as `* Returns true, if something...` in their summary-line. That's really bad because the summary is for a single-sentence description that shows up in IDEs.

For example, let's say I type `$resizer->getMinSize()`, then the IDE will say "Returns true, if something"... Instead, it should say "Get the minimum size", obviously.

The return-value is documented in the `@return` of PHPdoc, as you know. So never put it in the summary. ;-)

Look at the commit I did to see the changes to those "Return X" documentations, for a better style while going forwards.

Anyway, I will now proceed with uploading the little media resizer auto-tester I created. A useful tool for us to have.

Thanks again so much for this kickass work, man. :-) 👍

valga:
That was so fast, and I even did not update the description.

> Basically, if a function summary begins with the word "Returns", then that's wrong and should be rewritten another way. ;-)

Noted.

SteveJobzniak:
@valga You did an absolutely incredibly job on this refactoring. Seriously great work.

valga:
@SteveJobzniak I managed to collect some info about rotated videos.
1. Rotation flag is just a metadata value, so `ffprobe` always returns the same dimensions for all possible rotation angles.
2. Our video thumbnail generator doesn't take rotation flag into account.
3. Instagram relies on `ffmpeg` autorotation feature.
4. The app crops rotated videos to `original_width` and `original_height` aspect ratio.
5. `ffmpeg` applies autorotation after all filters.

For example, we have 640x480 raw video stream with rotation metadata = 90 (that makes video 480x640 in players, etc):
1. `ffprobe` returns 640x480.
2. Thumbnail is 640x480.
3. Uploaded video is 480x640.
4. The app crops it to 480x360 on display.

SteveJobzniak:
@valga That's excellent, I added a link to your comment to the project tracker:

https://github.com/mgp25/Instagram-API/projects/4?

By the way, please open www.gitter.im and log in with your Github account and send me a message (@SteveJobzniak) as soon as you see this. I'm working on your Realtime code and have some questions.

valga:
It should be noted that points 1-4 are not related with MediaAutoResizer. Our video uploader produces unexpected results with rotated videos, such as: thumbnail orientation doesn't match the video one, the video itself is getting cropped on screen.

And point 5 just means that MediaAutoResizer should swap axes before applying video filters. Also, it should handle horizontal and vertical flips for 180 and 270 degrees rotations.

PR #1586, opened 2017-09-26
Push: FBNS token&auth improvements

valga:
A couple of FBNS improvements:
1. Save FBNS token to the storage to be used in `login()` flows. Also it makes possible to register FBNS token only when it is updated (instead of doing that on every successful connection attempt).
2. Move the responsibility for updating FBNS auth from FBNS client to Push wrapper.


src/Push.php
```diff
@@ -117,7 +117,24 @@ protected function _getFbns()
             }
         });
         $fbns->on('fbns_token', function ($token) {
-            // Register the received token.
+            $oldToken = null;
+
+            try {
+                $oldToken = $this->_instagram->settings->get('fbns_token');
+            } catch (\Exception $e) {
+                $this->_logger->error(sprintf('Failed to read FBNS token: %s', $e->getMessage()));
```
SteveJobzniak: This always fails on the 1st connection, before a token has been created. Right?

In that case, why call it an 'error'? Perhaps it should say "info: There is no previous FBNS token."

src/Instagram.php
```diff
@@ -608,6 +608,24 @@ protected function _sendPreLoginFlow()
     }
 
     /**
+     * Registers available Push channels.
+     *
+     * @throws \InvalidArgumentException
+     * @throws \InstagramAPI\Exception\InstagramException
+     */
+    protected function _registerPushChannels()
```
SteveJobzniak: I think this belongs more in `src/Request/Internal`, just like other functions we use there such as `$this->internal->syncUserFeatures();`. :-)

The only functions we have in Instagram.php are ones related to the login itself and its flow (what functions it should call). We don't have internal details like this in the main class...

src/Push.php
valga: It doesn't fail on 1st connection: it returns `null` when value is missing or empty. So this try-catch is just for safety.

src/Instagram.php
valga: Well, it is certainly related to login flow, because you should't call it in other scenarios.

src/Push.php
SteveJobzniak: Ohh, good point. Gotcha.

src/Instagram.php
SteveJobzniak: Hmm... Yeah, I guess... We should think about it a bit more and analyze where it belongs. We've put many internal "login flow" things in Internal already.

valga: Initially I made it inline, but it turned out that we have a couple of almost identical blocks: https://github.com/valga/Instagram-API/blob/474a2439c273dfecc8c3d41c20f0ce4595900f36/src/Instagram.php#L669-L685 and https://github.com/valga/Instagram-API/blob/474a2439c273dfecc8c3d41c20f0ce4595900f36/src/Instagram.php#L711-L721 so I made the method. Perhaps we could just extend it to include all duplicates. Not sure on this one.

SteveJobzniak: Ah now I understand what you mean... Yeah, it belongs "inline" there, but of course we have 2 different login flows, so I agree with the decision of putting it inside a separate function in Instagram.php since it's very-very-very internal and is part of the login flow logic.

SteveJobzniak: I just realized one more question before merging: I noticed that this function checks "If a fbns_token exists in storage, then register the 'mqtt' push channel"... But this means that:

1. User runs PushExample, and gets a push token,
2. Imagine that the user always only runs non-Push code after that...
3. Still... it will register the push token every time they login()... even though the Push-receiver is not running.

Shouldn't this function be fixed to only run `$this->push->register` if we're currently running an InstagramAPI instance that has a Push listener?

valga: It sounds more like a token expiration stuff for me. Remember that `update_push_token` from `_messages` field? https://github.com/mgp25/Instagram-API/blob/cb8908bc1a816935c43b1e873ed0f0b4a55ba89b/src/ResponseTrait.php#L140-L144

It seems that we need to reset the FBNS token when that message has been received.

SteveJobzniak: Yeah, seems possible that registering simply "refreshes" the "this is our token: X"...

My worry was that calling `$this->push->register()` will make Instagram think that we're about to connect a Push-listener with that ID immediately after our register-call. And if we never again connect Push after registering our token, we may be a "bad" client in their eyes.

What do you think? I am not asking all this to be "difficult" about merging. :-) I am asking because I want to be sure we don't accidentally get users into trouble by making all clients constantly re-register to Push even when they never use Push anymore.

valga: > What do you think?

Hmm, we could a new storage key to store the token timestamp. And the we just stop calling `register()` on it when the token is too old.

SteveJobzniak: That sounds like a great idea! I trust you to figure out the implementation. Probably either count the time since we were given the token by their server, or perhaps count the time since we last used the `Push` class on that user (the `Push` class could save the timestamp to the storage, just like it already stores the token itself; then we know the user themselves used Push intentionally). So either store time-since-new-token-by-server, or time-since-last-used-Push-class-on-that-user... Maybe the latter is best.

SteveJobzniak: I edited my comment above after investigating more... And noticed that the new code would conflict with the lazy-register...

- The old Push class always registered push-tokens every time it got a token.
- The new one registers only if the new token differs from the old. If it's the same as the old, it just assumes it's already been registered by the login() flow.
- But... the push-registration in the login flow ONLY happens in full or refresh-logins. Not in "basic timeline-only resume". So we'll often call `login()` but it won't register the token. Then the `Push` class runs and sees that the token is identical so it doesn't call register either... It seems like a bug.
- So... the `Push` class needs something else: It needs knowledge about whether the token has been registered yet during this session..

SteveJobzniak: Okay, I see that after the "expiration system", my point above is solved...

Now, the Push class knows that if it sees a value in `fbns_token` then it has been registered by either the login flow or by itself... If no value, then the Push "token" event handler takes care of registering it.

Perfect. I think this solves everything.

I'm collapsing to bed now so I'll finish the merge tomorrow, but everything seems perfect now at first glance!

valga: > But... that push-registration in the login flow actually ONLY happens in full or refresh-logins.

The app does the same. Actually, FBNS token doesn't seem to change, because it consists of parts from FBNS auth. So basically it changes only when FBNS auth changes. And it is safe to register it only in refresh-logins.

SteveJobzniak: Ah perfect. I was thinking we'd need to check "last token timestamp" during quick-resume logins too, and instantly clear it.

But what you say makes sense: If it's a quick-resume login, then we've already registered the token.

Let's see...

- Imagine it's 3 days since last login() and last token timestamp: Their login() will do a "refresh flow: and will clear the fbns token. Then the Push client will detect the missing token and will register the new/same one.
- Imagine that it's 20 hours since last login() and last token timestamp: Their login() will do a "refresh flow" (because it's more than 30 minutes since last login() call). The refresh flow sees a <24h token, and registers it and keeps it. The Push client (if any) will detect the token in storage and will know that we still have it and that we have registered it during this "session".
- Imagine that it's 25 minutes since last login() call and that our token is less than 24 hours old: We know that our previous login() call happened 25 minutes ago, and we know that IT has already checked our token timestamp + re-registered it and kept the token value in the storage.

So yeah... with your system, everything works. The token is re-registered and kept every full/refresh login(), and just (correctly) assumed to have been recently re-registered during any "quick resume-login" (<30 minutes old).

All is good!

I added some commits to docs + function orders. Seems okay to you?

src/Push.php
```diff
@@ -100,15 +104,48 @@ protected function _getFbns()
...
             try {
                 $this->_instagram->push->register('mqtt', $token);
+                // Save the new, registered token to the storage.
+                try {
+                    $this->_instagram->settings->set('fbns_token', $token);
+                } catch (\Exception $e) {
+                    $this->_logger->error(sprintf('Failed to update FBNS token: %s', $e->getMessage()), [$token]);
+                }
             } catch (\Exception $e) {
                 $this->emit('error', $e);
```
SteveJobzniak: Actually... I changed everything here to make it only save the new token to storage after successful registration. But perhaps we should `$this->_instagram->settings->set('fbns_token', '');` if registration fails?

SteveJobzniak: Well, that, and moving `$oldToken = null` back to the top of the function body... because I just noticed the style-rule that forces "try" to have an empty line above it... :-)

SteveJobzniak: @valga Well, what do you think? Should I add token-clearing if token-registration fails (here and in the login flow)?

valga: Hmm, let's add it just to login flow.

SteveJobzniak: @valga Yeah, good idea. That makes sense. If we got it from the server in the Push-loop then obviously we can trust that the token is still OK and is safe to save to storage instantly. And then the login-flow will just re-verify that token IF it keeps failing. Sounds good to me!

SteveJobzniak:
Okay, I did two more tiny tweaks. Now it's absolutely perfect.

Thanks for these token&auth improvements. They really kick ass now. Totally reliable and very nice system for updating and tracking token activity and registrations. I love it. Time to merge! :)

PR #1587, opened 2017-09-26
Device: Add a DeviceInterface

valga:


mgp25:
@valga Are you still working in the PR, or it is just these changes?

valga:
Just these.

src/Devices/Device.php
```diff
@@ -7,7 +7,7 @@
  *
  * @author SteveJobzniak (https://github.com/SteveJobzniak)
  */
-class Device
+class Device implements DeviceInterface
```
SteveJobzniak: This now implements an interface, which means that those functions need `@inheritdoc` in this class.

src/Instagram.php
```diff
@@ -43,7 +43,7 @@ class Instagram
     /**
      * The Android Device for the currently active user.
      *
-     * @var \InstagramAPI\Devices\Device
+     * @var \InstagramAPI\Devices\DeviceInterface
```
SteveJobzniak: The second problem with this pull request is just a comment:

Please verify using `ag -s "\\bDevice\\b" -C2` (install ag if you don't have it). Look at all of the returned results. You will see many code locations that still refer to `Device` instead of `DeviceInterface` in parameters etc...

valga: Well, I really shouldn't trust mine IDE in find usages...

src/Devices/Device.php
valga: Done.

src/Instagram.php
SteveJobzniak: Yeah, I always use `ag` on the command line. Nice regex searcher, source-code/text-file aware, fast, nice output, and great for finding all instances before I rename anything. I highly recommend always running it. I wouldn't trust my own code rewrites if I didn't have "ag". :)

src/Devices/UserAgent.php
```diff
@@ -32,12 +32,8 @@ class UserAgent
     public static function buildUserAgent(
         $appVersion,
         $userLocale,
-        Device $device)
+        DeviceInterface $device)
     {
-        if (!$device instanceof Device) {
-            throw new \InvalidArgumentException('The device parameter must be a Device class instance.');
-        }
-
```
SteveJobzniak: Oops, haha, I think I added that before we added strict typing to the function parameter. ;)

PR #1611, opened 2017-10-09
Client: Exclude UnpredictableKeys from analysis

SteveJobzniak:
- This excludes "missing_definition" errors when analyzing classes that
  belong to the "UnpredictableKeys" types. That way we don't get warned
  about missing keys in those classes, since they obviously WILL always
  be missing keys! ;-)


mgp25:
🍰

SteveJobzniak:
💘

PR #1614, opened 2017-10-12
Restore Item&StoryLocation, Create AymfItem

aitte2:


mgp25:
Hello @aitte2, i havent tested the PR yet, will do in a couple of houts, but at the very first sight:

- https://github.com/mgp25/Instagram-API/pull/1614/files#diff-221df3522d3abe7ae70784e43c89fccaR50 In this case `attribution` is a string. I'll provide a response later.

aitte2:
Hi @mgp25 I restored Attribution to fix the damage from earlier commits. I didn't say that Attribution is right, because personally I haven't been able to find any response with attribution in it. :) After merging you should look at many different endpoints and when something cannot be mapped you should do what I did with `Item -> AymfItem:: override "caption" to be "string"`.

PR #1627, opened 2017-10-19
Update DirectThreadItem.php

stpnov:
Add Preview Medias

_If the wrong typography, I'm sorry, this is my first time_


SteveJobzniak:
Oh wow you manually wrote the `@method` lines? Some of them are in the wrong order so I think so... We have a tool for that. Type `composer codestyle`. xD

Anyway the most important part is this:

```
'preview_medias'                => 'Item[]',
```

Is that type correct? Does it work? Did you test `getPreviewMedias()` and successfully get an array of Item objects?

stpnov:
Yes, everything works correctly.

SteveJobzniak:
Thanks, that is very kind of you! What do you mean that you have 10% of my rewritten code? Which code? What? :D Are you saying you have rewritten this as a different library with 90% different code?

stpnov:
Yes that's right. I used this library and added additional functions that you do not have, well, and partially adapted to my own needs.

SteveJobzniak:
Okay cool. With the new license of this library, that 10% of rewritten code you're using is actually licensed with RPL-1.5: https://github.com/mgp25/Instagram-API#license

It is not a freeware license. It means "giving something back in return for what you are getting". In this case, two ways: License payment, or open sourcing your own project completely (all of it).

Which do you prefer? ;-)

stpnov:
I wanted to pay, but I couldn't even communicate normally with mgp25. He constantly ignores

SteveJobzniak:
Oh okay, that's great to hear. Well it's just because he gets so many emails (25+ per day). I never understood why he put his email publicly everywhere.

You can make the 100 EUR license fee payment here instead, and then we'll consider it taken care of: https://www.paypal.me/Armindale/0usd

SteveJobzniak:
That was one of the fastest lie detector tests I have ever performed. Went from realtime chat to graveyard. 🤣

stpnov:
I apologize, I just saw. 100 euros is more than I earn on the project, in which I use your library)

SteveJobzniak:
@stpnov Okay, well in that case in order to keep using the new RPL code you can contribute code like you're doing here. It's just about fairness. Giving something back. :-)

Btw: Have you seen `composer codestyle`? Run that command. Don't manually write the `@method` lines. ;-) It does it automatically.

stpnov:
I want to help) codestylecould not find `Command "codestyle" is not defined.`

SteveJobzniak:
@stpnov Oh. You must run it in the root folder of the project (same folder as `composer.json`). :)

PR #1637, opened 2017-10-22
Correction of the encoding for Emoji support

stpnov:
Correction of the encoding to support emoji if you use the PDO for your own purposes, like me


SteveJobzniak:
Hi, sorry this is not acceptable for multiple reasons. I'm sorry.

1. You forgot to also edit `COLLATE="utf8_general_ci"` in the table-creation in MySQL.php.
2. This is not the purpose of the PDO class. You are not supposed to re-use your object if you can't accept its rules.

https://github.com/mgp25/Instagram-API/blob/f3035e1aa3e9abce2c632a3b1f7cbd9bb0864de9/src/Settings/Factory.php#L86-L88

So your solution is simple: Make your own PDO object (connection). Let our library have its own with regular UTF-8 and let your code have its own with UTF-8 4-byte encoding.

SteveJobzniak:
Hmm, small update: I've been reading and apparently there's no performance downside to using utf8mb4 in MySQL. I'll be making a commit (credited to you for the idea) with the proper rewrite to UTF8MB4!

stpnov:
😊

SteveJobzniak:
😄

SteveJobzniak:
@stpnov Ok after lots of verification and research, I've finished the proper rewrite:

https://github.com/SteveJobzniak/Instagram-API/commit/55d79f10b365a95e056e0fc100a9f6a00f0b546d

Thanks a lot for making me aware of this MySQL problem. I had no idea that its utf8 mode was not true UTF-8. :-)

Okay so you can now use this PDO connection with utf8mb4 mode. Thanks a lot again. This was a great contribution.

And about your other idea (https://github.com/mgp25/Instagram-API/pull/1638): Now you know how to re-use PDO objects. :-) You should be ready to try this now!

PR #1638, opened 2017-10-22
If you use the storage for your own purposes

stpnov:
If you use the storage for your own purposes (PDO connection)

P.S. Randomly sent 2 requests and forgot to put the dot :с


SteveJobzniak:
Hi, uhm I'm sorry to shoot down your pull requests but this change makes no sense. ;-)

You're supposed to do it in reverse:

```php
$pdo = new PDO(); // <-- this is your PDO
$ig = new \InstagramAPI\Instagram($debug, $truncatedDebug, [
    'storage' => 'mysql',
    'pdo'     => $pdo, // <-- give it to us... and we will use it
]);
```

You can read more here:
https://github.com/mgp25/Instagram-API/blob/master/examples/customSettings.php

And here:
https://github.com/mgp25/Instagram-API/blob/f3035e1aa3e9abce2c632a3b1f7cbd9bb0864de9/src/Settings/Factory.php#L78-L89

SteveJobzniak:
To clarify: We don't want public access to the raw storage (with all of its closeLocation() etc functions), so `getStorage()` is a bad idea.

And `getPdo()` is a bad idea, since you should give us YOUR PDO instead, as I showed above.

SteveJobzniak:
Although as I said in your other pull request, your particular scenario is not a good situation for PDO re-use, because you're using 4-byte UTF 8 encoding and we're using regular 3-byte UTF 8 encoding. So you should use 2 separate PDO connections: https://github.com/mgp25/Instagram-API/pull/1637

PR #1657, opened 2017-10-27
Response/Model/DirectThreadItem: Add new property

itsMoji:


SteveJobzniak:
Thanks man, I have not verified that these are correct but they _look_ totally correct and consistent with everything else. And thanks for assigning data types to everything. Perfect. That's what we're supposed to do from now on. :-) I did some tedious work to assign types to lots of old values in a previous commit: https://github.com/mgp25/Instagram-API/commit/cef2679a315ae049277b27fe7819cc94f2ab3d08... But quite a lot of `''` untyped fields still exist. Thanks for not adding to that pile, hehe. :D

SteveJobzniak:
PS: Lol I just noticed that you manually wrote the `@method` stuff to the files. Because after merging I ran the codestyle command and it re-ordered your `@method` lines because they had been manually added. xD

You should just run `composer codestyle` in the root of the project directory and it will auto-add those things.

itsMoji:
@SteveJobzniak oh :)) i didn't know about `composer codestyle`

SteveJobzniak:
@itsMoji Yeah lol we forgot to advertise it. It was recently added to https://github.com/mgp25/Instagram-API/blob/master/CONTRIBUTING.md#codestyle-command

This is another good thing to do:
https://github.com/mgp25/Instagram-API/blob/master/CONTRIBUTING.md#git-setup

PR #1660, opened 2017-10-28
Refactor media handlers and forbid to upload illegal media

valga:
List of changes:
1. Forbid to upload rotated (or flipped) media and non MP4+H.264+AAC videos.
2. Make `MediaAutoResizer` abstract. Use corresponding resizer for you media.
3. Force video encoder output to H.264+AAC.
4. Refactor media validation.


SteveJobzniak:
Wohoo. Beautiful. I love you. This is looking great!

SteveJobzniak:
Btw, because of your commitstyle of many small commits I guess we should squash this one too?

Alternatively you could start running interactive rebases and squashing related commits to single commits. That is how I turned my 66-commit work to 37 cleaner units of changes. :)

src/Media/MediaDetails.php
```diff
@@ -69,6 +69,39 @@ public function getBasename()
...
+     * Check whether the media has swapped axes.
+     *
+     * @return bool
+     */
+    abstract public function hasSwappedAxes();
+
+    /**
+     * Check whether the media is horizontally flipped.
+     *
+     * *****      *****
```
SteveJobzniak: Hahah, I am reviewing now and so far (I have read 30% of all commits) everything is so beautiful and perfect.

I love this ascii art!

Just one note: This is illegal PHPdoc. Each * is interpreted as on Github and affects bold/italics. xD

To make it legal just put it inside the same character you would use on Github to add preformatted text. Meaning, put three ``` ` ``` above and below this ascii art. Then it will be preserved in generated docs. :-) PHPdoc has the same formatting rules as Github markdown.

src/Media/MediaDetails.php
```diff
@@ -3,10 +3,63 @@
...
+     * This range was decided through community research, which revealed that
+     * all Instagram stories are in ~9:16 (0.5625, widescreen portrait) ratio,
+     * with a small range of similar portrait ratios also being used sometimes.
+     *
+     * We have selected a photo/video story aspect range which supports all
+     * story media aspects that are commonly used by the app: 0.56 - 0.67.
+     * (That's ~1080x1611 to ~1080x1928.)
+     *
+     * However, note that we'll target the "best story aspect ratio range"
+     * by default and that you must manually disable that constructor option
```
SteveJobzniak: This comment refers to a MediaAutoResizer constructor option about enabling extended story ratio range. It needs tweaking now that it's no longer in that class.

src/MediaAutoResizer.php
```diff
@@ -34,60 +35,6 @@ class MediaAutoResizer
     const EXPAND = 2;
 
     /**
```
SteveJobzniak: Hmm, maybe move `BEST_STORY_RATIO`, `BEST_MIN_STORY_RATIO`, `BEST_MAX_STORY_RATIO` to `MediaDetails` too? Now we have some of the constants in here and some in there.

Sure, the best ratio ones are about resizing but maybe we should still have them all in the same place...

src/MediaAutoResizer.php
```diff
@@ -229,20 +229,8 @@ public function __construct(
         }
         $this->_verCropFocus = $verCropFocus;
 
-        // Target feed. Turn it into a string for easier processing,
-        // since we only care about story ratios vs general ratios.
-        switch ($targetFeed) {
-        case Constants::FEED_STORY:
-        case Constants::FEED_DIRECT_STORY:
```
SteveJobzniak: This is a bug now. Direct story and story must be treated the same which is why I put them all as a string so that we wouldn't have to check for both constants everywhere.

src/Media/Photo/PhotoResizer.php
```diff
@@ -224,28 +224,33 @@ protected function _processResource(
...
-            case 6:
-                $output = $this->_rotateResource($output, -90, $bgColor);
-                break;
-            case 7:
-                $output = $this->_rotateResource($output, -90, $bgColor, IMG_FLIP_HORIZONTAL);
-                break;
-            case 8:
-                $output = $this->_rotateResource($output, 90, $bgColor);
-                break;
+        if ($this->_details->hasSwappedAxes()) {
```
SteveJobzniak: Hmm, how about this:

```php
$angle = $flip = null;
if ($this->_details->hasSwappedAxes()) {
    ...
} else {
    ...
}
if ($angle !== null || $flip !== null) {
    $output = $this->_rotateResource($output, (int) $angle, $bgColor, $flip);
}
```

That way there's no risk anyone will bug it out in the future by contributing some tweak that doesn't set both variables. Maybe worth doing.

src/Media/Video/VideoDetails.php
```diff
@@ -81,13 +73,13 @@ public function hasSwappedAxes()
...
     {
-        return $this->_rotation === 90;
+        return $this->_rotation === 90 || $this->_rotation === 180;
     }
 
     /** {@inheritdoc} */
     public function isVerticallyFlipped()
     {
-        return $this->_rotation === 180;
+        return $this->_rotation === 180 || $this->_rotation === 270;
```
SteveJobzniak: I don't really understand the logic here.

Here's a chart of rotation:

rotational_symmetry_1

Which gives us:

```
0 degrees = no vertical flip, no horizontal flip
90 degrees = no vertical flip, no horizontal flip
180 degrees = yes vertical flip, no horizontal flip
270 degrees = no vertical flip, no horizontal flip
360 degrees = no vertical flip, no horizontal flip
```

Hmm? Right?

Because horizontal flip would mean that the horizontal pixels are flipped so that pixel MIN is the right side and pixel MAX is the left side. And rotation doesn't cause a flip horizontally. But it does cause a vertical flip at 180 degrees.

As for photos, with EXIF tags for flip/rotate, of course THOSE have flipping...

But I don't think videos have it just due to rotation angles, do they?

valga: 32135816-57ca6618-bc05-11e7-80c5-98fc2db5a516

src/Media/Photo/PhotoResizer.php
```diff
@@ -3,11 +3,14 @@
...
 
 use InstagramAPI\Media\Dimensions;
+use InstagramAPI\Media\MediaResizer;
 use InstagramAPI\Media\Rectangle;
-use InstagramAPI\Media\ResizerInterface;
 use InstagramAPI\Utils;
 
-class PhotoResizer implements ResizerInterface
+/**
+ * @property PhotoDetails $_details
```
SteveJobzniak: What on earth? Why isn't this just defined as `/** @var PhotoDetails */ protected $_details;` ?

valga: Because it is already defined in `MediaResizer`, and I just need to change a type hint for it.

src/Media/Photo/PhotoResizer.php
```diff
@@ -24,63 +27,30 @@ class PhotoResizer implements ResizerInterface
...
-    /** @var array Background color [R, G, B] for the final image. */
-    protected $_bgColor;
-
-    /** {@inheritdoc} */
+    /**
+     * Constructor.
+     *
+     * @param string $inputFile Path to an input file.
+     * @param array  $options   An associative array of optional parameters.
+     *                          See MediaResizer::__construct() description.
```
SteveJobzniak: This is good! Just one thing, which I learned while making a PHPdoc-compatible project: Make sure that any variables, functions or class name references are enclosed in ``` ` ```. As if you wrote this note on Github. Because PHPdoc uses Github flavored markdown, so from now on I've been writing all such references properly.

In other words:

```
 * @param array  $options   An associative array of optional parameters.
 *                          See `MediaResizer::__construct()` description.
```

src/Media/Video/VideoDetails.php
```diff
@@ -203,7 +235,29 @@ public function validate(
...
+            throw new \InvalidArgumentException(sprintf(
+                'Instagram only accepts videos encoded with H.264. Your file "%s" has "%s".',
+                $mediaFilename, $this->_videoCodec
+            ));
+        }
+
+        // Check the audio stream (if available).
+        if ($this->_audioCodec !== null && $this->_audioCodec !== 'aac') {
+            throw new \InvalidArgumentException(sprintf(
+                'Instagram only accepts videos encoded with AAC. Your file "%s" has "%s".',
```
SteveJobzniak: To not confuse newbies, let's add the word audio? `Instagram only accepts videos encoded with AAC audio`

src/Media/Video/VideoDetails.php
```diff
@@ -203,7 +235,29 @@ public function validate(
         // users. Do NOT use it to read from the hard disk!
         $mediaFilename = $this->getBasename();
 
-        // TODO: Ensure that video is mp4 + h264 + aac.
+        // Check the video stream. We should have at least one.
+        if ($this->_videoCodec !== 'h264') {
+            throw new \InvalidArgumentException(sprintf(
+                'Instagram only accepts videos encoded with H.264. Your file "%s" has "%s".',
```
SteveJobzniak: `Instagram only accepts videos encoded with the H.264 video codec.`

src/Media/Video/VideoDetails.php
```diff
@@ -203,7 +235,29 @@ public function validate(
...
+            throw new \InvalidArgumentException(sprintf(
+                'Instagram only accepts videos encoded with AAC. Your file "%s" has "%s".',
+                $mediaFilename, $this->_audioCodec
+            ));
+        }
+
+        // Check the containter format.
+        if (strpos($this->_container, 'mp4') === false) {
+            throw new \InvalidArgumentException(sprintf(
+                'Instagram only accepts videos packed into MP4. Your file "%s" has "%s".',
```
SteveJobzniak: `Instagram only accepts videos packed into MP4 containers.`

src/Media/Video/VideoDetails.php
```diff
@@ -203,7 +235,29 @@ public function validate(
...
-        // TODO: Ensure that video is mp4 + h264 + aac.
+        // Check the video stream. We should have at least one.
+        if ($this->_videoCodec !== 'h264') {
+            throw new \InvalidArgumentException(sprintf(
+                'Instagram only accepts videos encoded with H.264. Your file "%s" has "%s".',
+                $mediaFilename, $this->_videoCodec
+            ));
+        }
+
+        // Check the audio stream (if available).
```
SteveJobzniak: So Instagram supports silent videos? I've seen an error saying `Not a single valid audio stream found` from their server, posted by a user. So maybe they don't allow silent.

valga: Seem so. It even have a field for it: https://github.com/mgp25/Instagram-API/blob/dddcdf31c509105f37c7edf9de5e0fede2030b74/src/Request/Internal.php#L527

src/Media/Video/VideoResizer.php
```diff
@@ -141,29 +141,84 @@ protected function _processVideo(
...
+        $bitrate = $this->_calculateVideoBitrate($canvas);
+        $result[] = sprintf(
+            '-vcodec libx264 -preset veryfast -vb %.0fk -maxrate %.0fk -bufsize %.0fk',
+            $bitrate / 2,
+            $bitrate,
+            $bitrate * 1.5
+        );
+
+        // Force AAC for audio.
+        if ($this->_details->getAudioCodec() !== 'aac') {
```
SteveJobzniak: Stylistically I think it makes more sense to program if-statements as `if coded is aac, copy it, else change it`. I mean swapping it around to `=== 'aac'`.

src/Media/Video/VideoResizer.php
```diff
@@ -141,29 +141,84 @@ protected function _processVideo(
...
+            // The encoder 'aac' is experimental but experimental codecs are not enabled,
+            // add '-strict -2' if you want to use it.
```
SteveJobzniak: Hmm the situation with AAC on ffmpeg is actually kinda complicated.

`aac` may indeed be experimental, but not the proper one (`libfdk_aac`) with the highest audio quality: https://trac.ffmpeg.org/wiki/Encode/AAC

But the problem is that `libfdk_aac` is only available if compiled with `--enable-libfdk-aac`.

Here are the proper encoding parameters for FDK-AAC:

`
-c:a libfdk_aac -b:a 256k -cutoff 18000
`

src/Media/Video/VideoResizer.php
```diff
@@ -141,29 +141,84 @@ protected function _processVideo(
...
+     * @param Dimensions $canvas
+     *
+     * @return float Desired bitrate, in kbps.
+     */
+    protected function _calculateVideoBitrate(
+        Dimensions $canvas)
+    {
+        $result = $canvas->getWidth() * $canvas->getHeight();
+        if ($canvas->getWidth() > $canvas->getHeight()) {
+            $result *= 9.7222;
```
SteveJobzniak: What is this black magic? :P Algorithm from their APK?

Why not use `constant quality, variable bitrate` encoding? X264 doesn't like being told a specific bitrate. It likes quality-targets. Via `-crf` flag.

Like this (example is from a script I wrote a few months ago):
`ffmpeg -i "${INPUT}" -c:v libx264 -preset slow -crf 14 -c:a libfdk_aac -b:a 256k -cutoff 18000 "${OUTPUT}"`

examples/uploadVideo.php
```diff
@@ -30,7 +30,16 @@
...
-    $ig->timeline->uploadVideo($videoFilename, ['caption' => $captionText]);
+
+    // If you want to guarantee that the file is valid (correct format, codecs,
+    // width, height and aspect ratio), then you can run it through our
+    // automatic video resizer class. It is pretty fast, and only does any work
+    // when the input video file is invalid, so you may want to always use it.
+    // You have nothing to worry about, since the class uses temporary files if
+    // the input needs processing, and it never overwrites your original file.
+    // Also note that it has lots of options, so read its class documentation!
+    $resizer = new \InstagramAPI\Media\Video\Video($videoFilename);
```
SteveJobzniak: Typo. `\Video\Video`

examples/uploadVideo.php
```diff
@@ -30,7 +30,16 @@
     // Note that all video upload functions perform some automatic chunk upload
     // retries, in case of failing to upload all video chunks to Instagram's
     // server! Uploads therefore take longer when their server is overloaded.
-    $ig->timeline->uploadVideo($videoFilename, ['caption' => $captionText]);
+
+    // If you want to guarantee that the file is valid (correct format, codecs,
+    // width, height and aspect ratio), then you can run it through our
+    // automatic video resizer class. It is pretty fast, and only does any work
```
SteveJobzniak: The `it is pretty fast` comment is true for PhotoResizer but definitely not for VideoResizer, lol. That one can take minutes. xD

src/Media/Constraints/DirectStory.php
```diff
@@ -0,0 +1,18 @@
...
+    public function getTitle()
+    {
+        return 'direct story';
+    }
+
+    /** {@inheritdoc} */
+    public function getMinDuration()
+    {
+        return Direct::MIN_DURATION;
+    }
```
SteveJobzniak: Shouldn't this also import `Direct::MAX_DURATION`?

src/Media/Video/VideoDetails.php
SteveJobzniak: Mmmm okay! :-) 👍

src/Media/MediaDetails.php
SteveJobzniak: Ok this problematic comment has been moved into `src/Media/Constraints/Story.php` :)

src/Media/Video/VideoDetails.php
SteveJobzniak: Oh, that's fantastic, I love the colored image. We should maybe keep that inside `devtools/` as a reference! :P

SteveJobzniak:
Okay my review of all commits so far is complete. The comments above sometimes refer to outdated commits but whose problems have moved to new files, since I left the reviews on a commit by commit basis while reading.

Man, this is looking incredible so far. I love this whole refactoring! It cleans it all up so nicely.

src/Media/Constraints/DirectStory.php
valga: It inherits `getMaxDuration()` from `Story` constraints.

valga:
Alright, there are a couple of things left:
1. I don't like `BEST_STORY_*` handling, but I have no idea how to fix it. Perhaps we should move them into constraints.
2. Loop too short videos. It seems we need to use `movie` filter, because `stream_loop` option is buggy. But I have no energy to do that.

And StyleCI has disabled `align_double_arrow` option for `binary_operator_spaces` fixer, so its failing check is unrelated to this PR.

src/Media/Video/VideoResizer.php
```diff
@@ -153,11 +153,11 @@ protected function _getOutputFormat()
         // Force AAC for the audio.
         if ($this->_details->getAudioCodec() !== 'aac') {
             if ($this->_ffmpegWrapper->hasLibFdkAac()) {
-                $result[] = '-c:a libfdk_aac -b:a 160k';
+                $result[] = '-c:a libfdk_aac -vbr 4';
```
SteveJobzniak: This new VBR value looks good. At VBR 4 it has a cutoff of 15.5 kHz, which is all the high freq audio we need, so no need to use the cutoff parameter manually. http://wiki.hydrogenaud.io/index.php?title=Fraunhofer_FDK_AAC#Bandwidth

src/Media/Constraints/Story.php
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Highest allowed story aspect ratio.
+     *
+     * This range was decided through community research.
+     *
+     * @var float
+     */
+    const MAX_RATIO = 0.67;
+
+    const MIN_DURATION = 3.0;
```
SteveJobzniak: Maybe document the duration constants here with `/** @var float */` at least?

SteveJobzniak:
Okay, excellent work. About these things:

1. Yeah I mentioned it here too https://github.com/mgp25/Instagram-API/pull/1660#discussion_r147556538, perhaps we should simply put them in `Constraints\Story` (since DirectStory also derives from it). And have them as:

```php
class Story implements ConstraintsInterface
{
    const ...;
    public function getBestStoryAspectRatio();
    public function getMinBestStoryAspectRatio();
    public function getMaxBestStoryAspectRatio();
}
```

And then the `MediaResizer`, which has a few various "if targetfeed is story, use best ratio constant as target ratio":

```php
if constraints instanceof Story
$targetAspectRatio = constraints->getBestStoryAspectRatio();
```

I think that is the cleanest solution we will get...?

2. I was checking out the movie looping, but hmm it seems like a bad filter: https://ffmpeg.org/ffmpeg-filters.html#movie (It says that if looping is used, all timestamps will reset each time it loops. So the final file will have timestamps like `0:00:00, 0:00:10, 0:00:00, , 0:00:10` instead of `0:00:00, 0:00:10, 0:00:20, , 0:00:30` ... so more research is needed another time...

About StyleCI: Okay so we should just totally ignore styleCI from now on and only use composer codestyle.

src/Media/Constraints/DirectStory.php
SteveJobzniak: Ohh okay!

SteveJobzniak:
Okay I have finished a re-review of all previous comments + all new commits.

All is well. Only places to read are:

https://github.com/mgp25/Instagram-API/pull/1660#discussion_r147577510

https://github.com/mgp25/Instagram-API/pull/1660#discussion_r147577818

https://github.com/mgp25/Instagram-API/pull/1660#issuecomment-340254590

src/Media/Video/VideoResizer.php
valga: There is no cutoff parameter.

valga:
1. Media resizer should know nothing about stories. Perhaps we can do something like that:
```php
interface ConstraintsInterface {
    ...
    // return 0.5625 for stories and 1 for other media
    public function getRecommendedRatio();
    // return 0.0025 for stories and 0 for other media (because we can always crop them to 1:1)
    public function getRecommendedRatioDeviation();
    // return true for stories and false for other media
    public function useRecommendedRatioByDefault();
}
```
And then rename `useBestStoryRatio` param to `useRecommendedRatio`.
2. I see no problem about those timestamps.

src/Media/Video/VideoResizer.php
SteveJobzniak: @valga There is, unless they've recently removed it since I last used `-cutoff` in May:

https://trac.ffmpeg.org/wiki/Encode/AAC#fdk_aac

cutoff

valga: I mean, there is no cutoff param in my code. So there is nothing to change.

SteveJobzniak: Oh yeah this wasn't a change note, just a comment that the new VBR idea is good and needs no manual cutoff. :)

SteveJobzniak:
1. `getWithStatistics()`: Yeah, you're right about the `new PhotoDetails()` since those now have aspect ratio functions etc. I'll remove that task and notify the user that wanted it!
2. Do you think we should mention the `PhotoDetails()` and `VideoDetails()` trick in the uploadAlbum example? Ugh. Feels tedious to write about some kinda "Loop through all of your files, convert the first via resizer, then load its details, then apply the same ratio to every other file...". I don't want to even think about more programming now, but I can add this to the task tracker...
3. Timestamps: Okay. As long as they don't confuse Instagram or its apps... that's my worry. Maybe ffmpeg has a "PTS timestamp re-generate" filter. I think I've seen that in the past.
4. Functions not documenting their throws: Yeah, sadly any place that reads JSON data can throw that exception if the data cannot be mapped. I didn't document it because it would need changes in almost every file in this project. And I thought it's such a rare event that it's not worth mentioning. And that perhaps we can wrap that exception into an InstagramException instead (in `Client::mapResponseObject()` or better yet in `Request::getResponse()`) by catching mapping errors and re-throwing as `InstagramException` with the same message. That way, the only places that need to document `LazyJsonMapperException` are functions that read values from responses internally instead of just giving it to users... Well, I'll add this to the tracker.
5. About StyleCI: I mentioned it to mgp. I think we should switch to `fabpot.io` which runs PHP-CS-Fixer rules and therefore perfectly matches `composer codestyle`.
6. Now for the final and most important note: Yes, I like the "recommended ratio" concept, but with some tweaks...

```php
interface ConstraintsInterface {
    ...
    // return 0.5625 for stories and -1 for other media
    public function getRecommendedRatio();
    // return 0.0025 for stories and -1 for other media
    public function getRecommendedRatioDeviation();
    // return true for stories and false for other media
    public function hasRecommendedRatio();
}
```

```php
MediaResizer constructor:
        $this->_constraints = ConstraintsFactory::createFor($targetFeed);
        $forceTargetAspectRatio = null;
        if ($useRecommendedRatio && $this->_constraints->hasRecommendedRatio) {
            $forceTargetAspectRatio = $this->_constraints->getRecommendedRatio();
            $allowedMinRatio = $forceTargetAspectRatio - $this->_constraints-> getRecommendedRatioDeviation();
            $allowedMaxRatio = $forceTargetAspectRatio + $this->_constraints-> getRecommendedRatioDeviation();
        } else {
            $allowedMinRatio = $this->_constraints->getMinAspectRatio();
            $allowedMaxRatio = $this->_constraints->getMaxAspectRatio();
        }
        $this->_forceTargetAspectRatio = $forceTargetAspectRatio;

MediaResizer _calculateNewCanvas:
   /**
    * @param float     $inputAspectRatio
    * @param float|null $forceTargetAspectRatio If non-NULL, forces a specific target aspect ratio rather than aiming at the nearest legal ratio within the allowed range.
    */
    protected function _calculateNewCanvas(
        $operation,
        $inputWidth,
        $inputHeight,
        $inputAspectRatio, // NEW
        $isMod2CanvasRequired,
        $minWidth = 1,
        $maxWidth = 99999,
        $minAspectRatio = null,
        $maxAspectRatio = null,
        $forceTargetAspectRatio = null, // NEW
        $allowNewAspectDeviation = false)


And near the top:
   $targetAspectRatio = $forceTargetAspectRatio !== null ? $forceTargetAspectRatio : $inputAspectRatio;

And then carefully change all code that does stuff like:
   $minAspectRatio !== null && $targetAspectRatio < $minAspectRatio
into:
   $minAspectRatio !== null && $inputAspectRatio < $minAspectRatio

Now we'll have two different vars: The input and the target. So the code changes
must be careful to avoid bugging it out, since the checks should be done against input
and the resizing against target... hm.
```

valga:
2. Well, according to [this article](https://www.theverge.com/2017/8/29/16221562/instagram-portrait-landscape-photo-albums-carousel), Instagram still requires at least the same orientation for all album media. So it worth noting that in docs. And it's strange that we had no "only square" restriction for album media earlier.
4. We definitely should wrap 3rd party exceptions into ours.
6. Well, the app still crops timeline media to square by default, so we should have the default ratio for all non-story media set to 1.

SteveJobzniak:
Hehe I noticed that if you skip numbers in the list, it just goes to the next number. Your point 6 above says 4. I'll try a workaround. xD

1. x
2. OK we should mention it. It's in the task tracker so we don't forget. 
3. x
4. Yep. It requires some planning but I think the easiest 1st step (later) will be to wrap it inside `Request::getResponse()`. That way the core low-level `Client::` mapping function doesn't lie, but the Request takes care of the re-wrapping... Maybe. Or maybe we don't care if the Client lies and should grab it there. Well, screw it for now we'll discuss it later...
5. x 
6. Okay I fully understand your idea now and support it! You mean: Story media will say "use recommended ratio by default" and recommended ratio is 9:16. Other media will say "do not use recommended ratio by default" and recommended ratio is 1:1 square (classic instagram format). And no users need to run with `useRecommendedRatio` because it auto-selects Yes for stories and No for others. Perfect. Sorry I misunderstood earlier. Your idea is perfect. Go for it! :-)

By the way, really great work here. You went really far into this rewrite and refactoring. I'm impressed as always.

SteveJobzniak:
Hi valga, I'm about to sleep but I just saw the new commit. I'll review it fully tomorrow but there are many issues with it. I'll quickly drop in a few comments now and then do a full review tomorrow.

src/Media/Constraints/StoryConstraints.php
```diff
@@ -70,6 +103,24 @@ public function getMaxAspectRatio()
...
+    /** {@inheritdoc} */
+    public function getRecommendedRatioDeviation()
+    {
+        return self::RECOMMENDED_RATIO_DEVIATION;
+    }
+
+    /** {@inheritdoc} */
+    public function useRecommendedRatioByDefault()
+    {
+        return false;
```
SteveJobzniak: Must be true for stories. The extended non-9:16 story range is a bad idea and should be opt-in. That's why 9:16 should be default.

src/Media/MediaResizer.php
```diff
@@ -878,7 +849,7 @@ protected function _calculateNewCanvas(
             // Use floor() so that height will always be above minAspectRatio.
             $useFloorHeightRecalc = true;
             // Determine target ratio; in case of stories we always target 9:16.
-            $targetAspectRatio = $targetFeedIsStory ? self::BEST_STORY_RATIO : $minAspectRatio;
+            $targetAspectRatio = $this->_targetAspectRatio !== null ? $this->_targetAspectRatio : $minAspectRatio;
```
SteveJobzniak: Nevermind, when I said there were "many problems" I misread this as if you had changed the `$targetAspectRatio = $inputWidth / $inputHeight;` line to this, and thought "but that means the rest of the code no longer checks the input aspect ratio!". Sorry, I misread, it's time to sleep. ;-)

Yes this is correct.

So let's see...

1. It initializes target aspect ratio to input aspect ratio, to try to preserve it as much as possible.
2. Then it checks if target (aka input) is outside allowed range.
3. If it is outside allowed range, then it changes the new target to be either min, or max, or the special target ratio if a special forced ratio is enabled.

Yeah, this seems correct! <3

SteveJobzniak:
Okay @valga I've reviewed and merged this absolutely nuts work. You really go overkill (as do I) whenever something needs tweaking, and you've really created something very beautiful here. This refactoring is fantastic. Instagram users are going to be so happy to finally have strong photo and video validation and processing. Must have been exhausting to do this work, but wow, this is _great_... and now it is merged!

SteveJobzniak:
After merging I finished a final post-merge review, and ran some video and photo tests. It's all very, very, very perfect. Awesome work. Seriously. ;-)

I also found a funny side effect, caused by the fact that FFprobe and FFmpeg supports images: If you create `$resizer = new \InstagramAPI\Media\Video\VideoResizer('cat.jpg', $options);` it creates a MP4 file which displays the photo. Its duration is `00:00:00.04` (when I ffprobe the output file). But it seems that's totally fine to do that after the `$times` loop is implemented. It says "$times = 75", and yeah 0.04 seconds times 75 is 3 seconds!

valga:
I would say it has become closer to perfection, but it is still not perfect.

By the way, I have a proposal for class names: `UploadReadyMedia`, `UploadReadyPhoto`, ..., `VideoThumbnail`.

SteveJobzniak:
@valga Hey. :-) I've been thinking that we should discuss this again, so it's a good thing that you brought it up. We should brainstorm. If you have any more names, add them to this list.

UploadReady:

- `MediaResizer` = `UploadReadyMedia`
- `PhotoResizer` = `UploadReadyPhoto`
- `VideoResizer` = `UploadReadyVideo`
- `ThumbResizer` = `VideoThumbnail`

```php
$img = new UploadReadyPhoto('cat.jpg');
$ig->timeline->uploadPhoto($img->getFile());
```

Valid:

- `MediaResizer` = `ValidMedia`
- `PhotoResizer` = `ValidPhoto`
- `VideoResizer` = `ValidVideo`
- `ThumbResizer` = `ValidThumbnail`

```php
$img = new ValidPhoto('cat.jpg');
$ig->timeline->uploadPhoto($img->getFile());
```

Instagram:

- `MediaResizer` = `MediaProcessor`
- `PhotoResizer` = `InstagramPhoto`
- `VideoResizer` = `InstagramVideo`
- `ThumbResizer` = `InstagramThumb`

```php
$img = new InstagramPhoto('cat.jpg');
$ig->timeline->uploadPhoto($img->getFile());
```

Processed:

- `MediaResizer` = `MediaProcessor`
- `PhotoResizer` = `ProcessedPhoto`
- `VideoResizer` = `ProcessedVideo`
- `ThumbResizer` = `ProcessedThumb`

```php
$img = new ProcessedPhoto('cat.jpg');
$ig->timeline->uploadPhoto($img->getFile());
```

Managed:

- `MediaResizer` = `ManagedMedia`
- `PhotoResizer` = `ManagedPhoto`
- `VideoResizer` = `ManagedVideo`
- `ThumbResizer` = `ManagedThumb`

```php
$img = new ManagedPhoto('cat.jpg');
$ig->timeline->uploadPhoto($img->getFile());
```

Automatic:

- `MediaResizer` = `MediaProcessor`
- `PhotoResizer` = `AutomaticPhoto`
- `VideoResizer` = `AutomaticVideo`
- `ThumbResizer` = `AutomaticThumb`

```php
$img = new AutomaticPhoto('cat.jpg');
$ig->timeline->uploadPhoto($img->getFile());
```

Fixed:

- `MediaResizer` = `MediaProcessor`
- `PhotoResizer` = `FixedPhoto`
- `VideoResizer` = `FixedVideo`
- `ThumbResizer` = `FixedThumb`

```php
$img = new FixedPhoto('cat.jpg');
$ig->timeline->uploadPhoto($img->getFile());
```

Optimized:

- `MediaResizer` = `MediaProcessor`
- `PhotoResizer` = `OptimizedPhoto`
- `VideoResizer` = `OptimizedVideo`
- `ThumbResizer` = `OptimizedThumb`

```php
$img = new OptimizedPhoto('cat.jpg');
$ig->timeline->uploadPhoto($img->getFile());
```

---

Many of these are ugly, too long, too unclear, or give the wrong impression, but I just brainstormed every tech "validation" terminology I could think of.

Out of all of these, my absolute favorites are `InstagramPhoto`, `InstagramVideo`, `InstagramThumb`. I don't care that they "don't represent a photo that comes _FROM_ their server" (which you mentioned earlier).

The reason they have the name "Instagram" in them _isn't_ that they're _FROM_ their server.

It's that they're going _TO_ Instagram's server.

So the reason for the word "Instagram" is that they represent a photo that is verified/validated/processed/manipulated according to _Instagram's RULES_ so that they can be uploaded _TO Instagram's server_, hence "InstagramPhoto", "InstagramVideo", and "InstagramThumb". The only reason to manipulate the files is to make them valid for Instagram usage, hence `Instagram*`. And these class names make _perfect_ sense for newbie users and fits in with all other library classes, to create a "logical unit":

```php
use InstagramAPI\Instagram;
use InstagramAPI\InstagramID;
use InstagramAPI\Media\Photo\InstagramPhoto;
use InstagramAPI\Media\Video\InstagramVideo;

$ig = new Instagram();

$photo = new InstagramPhoto('cat.jpg');
$ig->timeline->uploadPhoto($photo->getFile());

$feed = $ig->timeline->getSelfUserFeed();
$item = $feed->getItems()[0];
$code = InstagramID::toCode($item->getPk());

$video = new InstagramVideo('morecats.avi');
$ig->story->uploadVideo($video->getFile());
```

Another great benefit is this:

When a user imports tons of different libraries (our Instagram library, plus maybe some HTTP libraries, some media manipulation libraries, etc). Then the `Instagram*` classes remain perfectly named. They won't get confusingly lost in a sea of other classes and people won't have to do `use InstagramAPI\Media\Photo\ValidPhoto as InstagramValidPhoto;` to make the names unique/sane.

But I won't proceed with any rename until we agree. I just feel very strongly that the best name so far is `Instagram*`. :-O

valga:
1. "Yes" for `VideoThumbnail`. Because it is internal, and we forbid to upload illegal videos anyway.
2. "No" for `MediaProcessor`. It must be `*Media`.
3. Still have no great idea about that `*`.

SteveJobzniak:
1. Agreed.
2. Agreed too, because `MediaProcessor` makes it sound like it's a class that can handle more than one input file, as we said before. And `ProcessedMedia` is also wrong since it's not always processed.

How about this:

```php
class UploadReadyMedia {} // internal
class InstagramPhoto extends UploadReadyMedia {} // external
class InstagramVideo extends UploadReadyMedia {} // external
class VideoThumbnail extends InstagramVideo {} // internal
```

Because I like the name UploadReady for internal purposes. It's just when we get external for user-purposes that I dislike it for the newbies' and for "library cohesiveness" sake.

By the way, I had some thoughts... that it would be nice to add the ability for Upload functions to accept `InstagramPhoto` and `InstagramVideo` objects and automatically handle the `getFile()` calls. But that's a later thought. It would be pretty nice though: `$ig->timeline->uploadPhoto(new InstagramPhoto('cat.jpg'))` instead of `$img = new InstagramPhoto('cat.jpg'); $ig->timeline->uploadPhoto($img->getFile())`. We don't have many upload-related functions so it would be a pretty easy job. Do you think we should do it? The only part I really dislike is that we'd turn the filename argument from `string` into `InstagramPhoto|string`.

Alternatively we could totally ban the uploading of string-filenames, to prevent people from doing invalid uploads. The `PhotoDetails/VideoDetails` checks are fast enough that it shouldn't matter for people if we require actual objects.

SteveJobzniak:
>  prefix must be the same for all 3 classes.

Yeah I thought about that too but `VideoThumbnail` doesn't follow that rule at all. So I don't see any problem with any of these choices:

```php
class UploadReadyMedia {} // internal
class InstagramPhoto extends UploadReadyMedia {} // external
class InstagramVideo extends UploadReadyMedia {} // external
class VideoThumbnail extends InstagramVideo {} // internal
```

```php
class InstagramReadyMedia {} // internal
class InstagramPhoto extends InstagramReadyMedia {} // external
class InstagramVideo extends InstagramReadyMedia {} // external
class VideoThumbnail extends InstagramVideo {} // internal
```

```php
class InstagramMedia {} // internal
class InstagramPhoto extends InstagramMedia {} // external
class InstagramVideo extends InstagramMedia {} // external
class VideoThumbnail extends InstagramVideo {} // internal
```

```php
class InstagramMedia {} // internal
class InstagramPhoto extends InstagramMedia {} // external
class InstagramVideo extends InstagramMedia {} // external
class InstagramThumbnail extends InstagramVideo {} // internal
```

Maybe even the last one if we want 100% consistency.

---

> We can just implement __toString() method in MediaResizer

Ouch yeah, it can't throw exceptions.

But that's why we should perhaps simply always require valid objects in all upload methods. Because what possible reason can we have to allow people to upload "Some 1.5 GB movie.mkv" or "photo.png" via string-filename?

> We must check whether it is just the same orientation or the same aspect ratio for album media.

Good idea! We can simply loop over all files the user has added to the album array and check their aspect ratio to see if they all match.

Btw, if we require that users always provide objects, then it would be trivial to do this check in a speedy way. We'd simply read the aspect ratio values from the finished objects somehow.

In fact, we could change it like this:

1. All functions require `InstagramPhoto` and `InstagramVideo` objects.
2. Merge `getFile()` into `MediaResizer::__construct()` so that the file is instantly created. That way there's no need for us or them to ensure that `getFile()` has been called.
3. Make the "getFile" processing fill in stuff like `$this->_outputAspectRatio` and other functions we need, and simply make `getOutputAspectRatio()`.
4. So, to check an album, we simply loop over all objects, and check `getOutputAspectRatio` of all of them and ensure they're all within a few % of the 1st album file.

And for users to make an album all they have to do is make the object for file 1, then append the others to the array via `getOutputAspectRatio()` from the 1st file as their target aspect ratio for the other files.

---

> We already propagate target feed, and there is no need to propagate the aspect ratio, because legal video will produce legal thumbnail with the same dimensions.

Good point, I didn't think about that! Of course the thumbnail size will match the already-legal video. :-) Doh.

> As for timestamp, it is a nice idea worth implementing.

Yeah!

> I don't find it useful [to handle targetFeed]

Well it would certainly complicate our job and our design of the classes... although if we do switch to strict objects, we could also enforce the targetfeed so that users 100% cannot create invalid media by marking a Story as target when uploading Timeline objects etc.

Well... all of this is just making me tired now. Everything is as fine as it was yesterday too if we just stop after renaming the classes. ;-)

valga:
We can't switch to strict objects, at least for videos, because re-encoding 720p video could take several minutes. Preparing a file to upload and uploading that file are 2 different steps, and they must be done separately. Otherwise network or Instagram errors could lead to yet another one processing job.

So the only nice idea (apart from renaming classes) is to allow setting custom timestamp for video thumbnails. Maybe I will do it with video loops.

SteveJobzniak:
Yeah let's skip the ideas of switching to objects completely.

About names: Should we choose the 100% consistent one (last option above)? I think it _would_ definitely be nice to have totally consistent names.

As for the albums: The real app just ensures that all media matches the 1st aspect ratio. That's what happens if you look at albums from the real app. All files have the same aspect ratio. And online, they display in a viewer which is adjusted to that ratio. If some files have a different ratio, as our library allows, then they'll look stretched on Instagram.com (but not in the app).

SteveJobzniak:
@valga (repeating question) Should we choose the 100% consistent one (last option above)? I think it would definitely be nice to have totally consistent names. ;-)

valga:
@SteveJobzniak Missed it, sorry. Just pick the one you like.

SteveJobzniak:
@valga No worries. Alright, check this commit to see the changes:

https://github.com/mgp25/Instagram-API/commit/d03477f1b3e51d856ca9043b4637d6c6df48a229

https://github.com/mgp25/Instagram-API/commit/bd7fa6fb61725e599ac8e498376f4f9a0b5e26f1

As for the `LazyJsonMapperException` we were discussing, we'll see how to wrap/document that... Currently, all of our library code here is doing stuff like `getResponse(new SomeLazyClass())`. And as you know, the Lazy constructor is marked as "throws". That is no problem, though. An exception can only be thrown in three situations in the Lazy constructor:

- If the class map is invalid and cannot be compiled. That will never happen, because we use `lazydoctor` in our codestyle script and in the pre-commit hook, which not only documents but also diagnoses every class and every property in the entire project and ensures everything is compilable.
- The second possibility is if the class is constructed with JSON data AND the `$requireAnalysis` flag is TRUE AND there are problems with the class map compared to the data. But as you can see, we NEVER construct objects that way. So that way cannot throw either.
- The final way that can throw, is if the class has a custom `_init()` function and THAT one throws an exception. But we don't have even a _single_ init function in this project since we don't use that feature (and I don't think we ever will).

So, in short: Class constructors will never throw. Not in the realtime/push clients and not in the Request core.

Alright, next up: Our usage of `->assignObjectData()` in `Client::mapServerResponse()`. We already do that in a try-catch block and rewrap it as an `InstagramException`. Although, we do assignObjectData with `$requireAnalysis = false`, which means that it will actually never throw. The catch is actually for OUR own throwing of a CUSTOM Lazy analysis exception (which ignores missing keys in the "unpredictable container" classes).

So, `mapServerResponse()` will never throw a Lazy exception in the code that maps the data.

That leaves only one area that CAN throw: Everywhere in the code where we interact with data in the Lazy object. If the interactions are invalid (ie a `getSomething()` (or internal `_getProperty()`) which is mapped as `string` but is actually a `{"object":true}` due to Instagram having changed that property, or a `setSomething([])` when the map is a `string`). Those are ultra rare though. Almost so rare that we don't have to care. I think I've seen Instagram change a data type (while keeping the old variable name) a single time, if even that...

But yeah, `mapServerResponse()` could throw if the various getters at the end of that function fail. And various other code in internal functions that do reading of data could throw (such as video uploader and similar things that read JSON properties) IF the data changes type in the future.

There's almost zero risk that data will change type, but it could happen. Luckily there are only a few places of library code that read properties.

SteveJobzniak:
@valga I've now taken care of the Request.php throws:

https://github.com/mgp25/Instagram-API/commit/aa2527bb3a4ff6d4888990dac725d339330c92d9

These are the remaining tasks, but I can't do any more. My plans from the 9th when I "left" haven't changed. They were just shifted. I was back between the 14th (fever and bored) and 24th (my work finally merged) and then from 24th to now to help work on your pull request with you as I had promised you.

It's not impossible that I'll tinker with something, but it's less and less likely because my burnout is severe.

Issue #927, opened 2017-02-07
Search Hashtags


Triden:
The latest news on the hashtag. They made a lot of different experiences. Able to do so. Intercepted request to post the official publication of the application and its substituted picture. In search, it is not shown. If you do not replace all the displays. This proves that the image data is important. What's stored is not found. There are no assumptions about you? Exif, who was able to get empty.

mgp25:
@Triden test the following thing: Upload a photo with your device. Download that image using the API, and then reupload the image with the API. Since you have downloaded the imagen from the app, it should contain the exif, and therefore it should work.

Try it.

Triden:
It does not make sense. When loaded, the server overwrites this data and change a photo. We intercepted on the proxy level. And there copy a photo bits. Restores the image of bits. Viewed in photo, but have not been able to find. But! This image appears in the search! But! If the proxy level to replace the image. Override bits on his own. It be published, but in the search to no find it! It is difficult to explain in English. By this I concluded. The fact that the picture is why it is important. Do you understand me?

mgp25:
@Triden im not going to research it or do it either. If you want this, someone else or you should do it.

As i said before, i lost interest in Instagram

Regards

Triden:
It is sad. I'm understood, thank you.

mgp25:
@Triden you can always do it yourself, research about it, reverse engineer, and code, it will take you some time, but you will get what you want

Regards

ppeixoto:
@Triden Seems they are using OpenGL to encode the image somehow. I think this is the way Instagram knows if the image was posted from the original App or not.

PR #1672, opened 2017-11-02
Push: Document more pushes

valga:


valga:
Hi @crouchingtiger. I started to receive pushes a couple of days ago. I wonder if you have something to add to that list.

crouchingtiger:
@valga Looks good! Only one you're missing is ``reply_to_comment_with_threading`` with payload
``comments_v2?media_id=1111111111111111111_1111111111&target_comment_id=11111111111111111``

valga:
@crouchingtiger Thanks! It seems that I need to collect even more pushes.

src/Realtime.php
```diff
@@ -279,10 +203,15 @@ public function sendLikeToDirect(
         $threadId,
         array $options = [])
     {
-        return $this->_sendItemToDirect(array_merge($options, [
-            'thread_id' => $threadId,
-            'item_type' => 'like',
-        ]));
+        try {
+            $command = new Command\sendLike($threadId, $options);
```
SteveJobzniak: `sendLike` seems to be a typo (lowercase S)

mgp25: @SteveJobzniak gitter. Important.

SteveJobzniak:
@valga Excellent. So is this one ready to merge (the single commit)?

valga:
@SteveJobzniak I did not manage to receive `reply_to_comment_with_threading` notification yet, but I guess it could be added later (perhaps along with `livewith` one).

SteveJobzniak:
@valga Hmm. I see. Well, decide if we should merge now or wait for that?

PS: Are you using the signature (or newer) than what we have in dev-master constants? They added comment threading in mid-August and our signature is from Aug 30th or so. But they may need an even later one for that notification.

valga:
@SteveJobzniak I can't update to the latest signature because of some missing headers and parameters, so I am using the one from June or July, that's why I can't receive new pushes. And I see no possibility to update to the latest version in foreseeable future, so let's merge this PR.

Speaking of missing headers, see #1696.

SteveJobzniak:
@valga Ah, well there's no danger with running with `dev-master` without the HTTP-speed-stuff headers. Instagram already detects this API and all 3rd party APIs anyway and doesn't care. ;-)

Anyway, okay let's merge 👍

valga:
@SteveJobzniak As long as we know nothing about how Instagram detects us, it's better to stay as close to the app as possible. Even if we think that they don't care.

SteveJobzniak:
@valga We actually know how, and I can say it without any danger of revealing too much: They have server AI that analyzes behavior and API calls and headers and navigation patterns and parameters and analytics (graph) and IPs and lots of other parameters, and easily knows when something is a human or a 3rd party library.

That is why media uploaded via any 3rd party API is never listed in its `#hashtag` feeds. They know. And since they know, and have specifically coded `if (not real app) { hide from hashtag feed, and apply stricter antispam filters }`, instead of `if (not real app) { ban }`, we can safely assume they don't mind third party APIs. They have the situation under total control and are successfully combatting spam via their constantly evolving AI.

I can assure you that every single one of your accounts (you mentioned having several within a shared eventloop) are 100% detected, despite all your efforts to queue up actions in a human-like way, avoid latest constants until headers are emulated, etc. None of that matters. They know...

However, I fully agree that emulation accuracy is important because it helps us be good, clean citizens on their service. That is why that matters. :-)

crouchingtiger:
@SteveJobzniak @valga That's kind of half true. I don't use this API directly so I'm not sure about this repo specifically, but videos and carousel posts aren't affected by the issue you mention on my own API implementation. Only photos suffer the issue you mention (which does indeed indicate there is some kind of analysis going on, but it's not the most sophisticated in the world). 

As a test - clear out all data for Instagram on Android (Settings > Apps), change your device ID (free apps on the Play Store), block your requests to ``graph.facebook.com/logging_client_events`` on a real device, and then check what happens when you upload photos.

valga:
@crouchingtiger I remember some guy who avoided hashtag problem by uploading the photo extracted from the app's captured requests, so their detection could also rely on some file metadata. And it is super easy to detect `CREATOR: gd-jpeg v1.0 (using IJG JPEG v62), quality = 95` JPEG comment.

SteveJobzniak:
@valga hmm

crouchingtiger:
@valga I tried the same too a while back (September 2016 when the API change came into effect), but didn't have success at the time. I'm going to investigate further, thanks.

crouchingtiger:
@SteveJobzniak I wouldn't worry *too much* about the spammers. Instagram's algorithsm know when hashtags are being abused, and new accounts can't upload more than 100 posts in 24h. Spammers would probably prefer using bots, which are commenting/liking/following en masse, rather than uploading rubbish content.

valga:
@SteveJobzniak I am not sure about it. Here is [the issue](https://github.com/mgp25/Instagram-API/issues/927).

@crouchingtiger I don't do any uploads, so I can't confirm it. Anyway, as analytics is not realtime (and can't be used with 100% accuracy because of that), we have only JSON data and the JPEG left.

SteveJobzniak:
@crouchingtiger True, but Instagram seems to worry about it since they have this system of preventing unauthorized images on hashtag feeds. Someone would abuse it if we "crack the system". And then Instagram would just change the system to make it secure again. :-S

@valga Interesting. Very hard to understand what that guy is writing but he does seem to be saying that it's related to bits in the image. He also seems to be saying: There was no EXIF data (he said `Exif, who was able to get empty.`). He also speaks about capturing the image with a proxy (before it was sent to Instagram) and looking at it in an image viewer and being unable to see the pixel "bits".

Perhaps the image itself has some bits (steganography perhaps, embedded in the JPEG data stream)... I know there are programs that can put short text files inside JPG files. So perhaps Instagram uses that method to authenticate real app-photos vs fake ones.

He seems to be saying that when he copied the JPG binary created by the real app, he was able to make it appear in the hashtag search.

He also seemed to be saying that the steganography bits/data is scrubbed when Instagram's server processes the image and places it on the CDN. Since the CDN does not hold the original file. it holds a resized/recompressed copy.

Also: Analytics may still be related, because even though analytics themselves are delayed, their server still knows: "csrftoken XXXXXXXX has never sent us any analytics", so they know which sessions are fake, and may use that to punish media uploaded from those tokens. We also have another way to prove they're detecting us: They have much stricter limits for "spammy" actions such as like/follow when using 3rd party APIs instead of the real app.

We emulate the JSON arguments perfectly.

So my guess is analytics (as a "flag" of "probably not a real client"), and possibly also steganography inside the JPEG... The steganography makes a lot of sense (as a way of protecting/authenticating media) and is probably the main mechanism.

crouchingtiger:
@SteveJobzniak It's impossible to compare the uploaded data with the data received when downloading the image from Instagram's CDN, as there's post-processing going on remotely (even for the non transcoded 1080px original resolution image). eg. You'll find some additional Facebook IPTC info was embedded.

I already went through the path of steganography last year. I felt like I exhausted all options - I even placed a random JPG from a random online image search with entirely different EXIF info into the filesystem right before upload, and it seemed okay on the searches. 

Regarding the analytics - that's why I mentioned you try blocking the analytics URL on a real device that's entirely reset (https://github.com/mgp25/Instagram-API/pull/1672#issuecomment-345785188).

Newly created accounts I created from a new device, that I never touched with any API, also had their posts hidden in searches - so it's not all black and white that posting from a device - account activity plays a part. Rule of thumb - if you can post fine to the searches from a device, then it's down to a device issue. I encourage you to try the approach above if you have a real Android device though. Analytics logging could always get blocked on a corporate level eg. places that block Facebook. I'm pretty sure people posting on those networks can still get their hashtags seen ;)

valga:
@SteveJobzniak I think that steganography is an overkill: you can filter out like 90% of "bad" photos just by checking for EXIF/comments/embedded profiles, because real app doesn't have anything of that, while almost every photo editing software places its tags into a JPEG. Also, the app uses its own JPEG encoding library (libjpeg-turbo or mozjpeg) with specific settings, so it might be detectable too.

And as @crouchingtiger said, Instagram could enforce some restrictions on newly created/inactive/unpopular/hashtag abusing/whatever accounts.

SteveJobzniak:
Well, if either of you want to investigate further, there is definitely a possibility that they simply look at exif, creator, and iptc profiles.

crouchingtiger:
@SteveJobzniak @valga My API client now shows tagged posts in hashtag feeds. I didn't perform any changes, but it seems it might have been fixed during the "follow hashtag" feature rollout. You might want to check this API to see if it's working too.

PR #1678, opened 2017-11-05
Refactor media uploads, add resumable photo uploader

valga:
* Refactor upload exceptions system: wrap runtime errors into `UploadFailedException`.
* Prepare for new uploaders based on the resumable one.
* Add resumable photo uploader.
* Make video thumbnail uploads implicit.


src/Request/Internal.php
```diff
@@ -1179,6 +1178,145 @@ public function configureWithRetries(
...
+        $targetFeed,
+        InternalMetadata $internalMetadata)
+    {
+        switch ($targetFeed) {
+            case Constants::FEED_TIMELINE_ALBUM:
+                $result = false;
+                break;
+            default:
+                $result = $this->ig->isExperimentEnabled(
+                    'ig_android_photo_fbupload_universe',
```
crouchingtiger: 👍

valga: I know, that's why I have `$result = false;` for `FEED_TIMELINE_ALBUM` for both photo and video.

By the way, yet another video upload method is being rolled out: segmented one. Basically, it is resumable chunked uploader, where each chunk duration is 5 seconds. It is used for videos longer than 10 seconds.

crouchingtiger: In which version did you find this?

valga: 22.0.0.15.68 already has it enabled, but just for timeline so far.

SteveJobzniak: @valga How on Earth is it making each chunk 5 seconds long? It would have to create a new video container per-chunk and put exactly 5 seconds of data (at current FPS) into it. It doesn't sound logical that they're doing all that effort. What really matters is filesize. Perhaps each chunk is a certain size?

valga: @SteveJobzniak They are using [MediaCodec](https://developer.android.com/reference/android/media/MediaCodec.html) to render a video into a file. It has a buffered output, so you can close current file and open new one on the fly. And it is obvious that you can do that after a given duration, because `onOutputBufferAvailable` event has [MediaCodec.BufferInfo](https://developer.android.com/reference/android/media/MediaCodec.BufferInfo.html) parameter with `presentationTimeUs`.

We can do it with `ffprobe`, as it can dump packets at specified timestamps:
```
$ ffprobe -v quiet -read_intervals "5%+#2,10%+#2,15%+#2" -show_packets -print_format json timeline.mp4
{
    "packets": [
        {
            "codec_type": "audio",
            "stream_index": 1,
            "pts": 216064,
            "pts_time": "4.899410",
            "dts": 216064,
            "dts_time": "4.899410",
            "duration": 1024,
            "duration_time": "0.023220",
            "size": "182",
            "pos": "214640",
            "flags": "K"
        },
        ...
        {
            "codec_type": "video",
            "stream_index": 0,
            "pts": 167424,
            "pts_time": "14.696629",
            "dts": 165888,
            "dts_time": "14.561798",
            "duration": 768,
            "duration_time": "0.067416",
            "size": "7232",
            "pos": "782838",
            "flags": "K"
        }
    ]
}
```

valga: Disregard my comment about `ffprobe`, because we need to upload video and audio streams separately. To conclude: each stream must be splitted into 5sec segments, and then each segment should be uploaded with resumable uploader.

SteveJobzniak: Wow. Sounds very painful. :-\

valga: Resumable uploader can't be used in parallel uploads. They are aiming to use multiple threads for video uploads, hence the segmentation.

Also their backend could support some on-the-fly transcoding, hence the segmentation by timecodes.

SteveJobzniak: Ohh okay. But they could already use parallel threads via "Range" offsets during uploading. Oh well not everything Instagram does is logical. :P

valga: Well, resumable and segmented uploaders are owned by the Facebook, so you should blame them. They are shutting down Instagram's own infrastructure step by step: good old media uploaders are being used just by albums and old app versions. And each Instagram user has a shadow user in the Facebook.

valga:
Hi @SteveJobzniak. This PR is ready for a final review. Apart from resumable uploader, I refactored upload exceptions a bit.

src/Request/Internal.php
SteveJobzniak: True. I have also seen a few Instagram media recently with Facebook's CDN URL encryption keys. Which cannot be downloaded without providing the correct `?query`. Facebook is doing to Instagram what Microsoft did when it bought Skype: Putting their own team of engineers on monkey-adding their own infrastructure to an existing codebase. ;) Some things are well done. Others are a mess...

SteveJobzniak:
@valga Fantastic news, thanks! I will review either today or tomorrow!

src/Request/Internal.php
```diff
@@ -1106,8 +1191,8 @@ public function configureWithRetries(
         while (true) {
             // Check for max retry-limit, and throw if we exceeded it.
             if (++$attempt > self::MAX_CONFIGURE_RETRIES) {
-                throw new \InstagramAPI\Exception\UploadFailedException(sprintf(
-                    'Configuration of "%s" failed. All retries have failed.',
+                throw new \RuntimeException(sprintf(
+                    'All %s configuration retries have failed.',
```
SteveJobzniak: The `$entity` variable is either `album`, `video` etc, _or_ `$videoFilename` (from the caller). Because it can be a filename which may be very long, perhaps all `$entity` usages in messages (everywhere in the codebase) should be wrapped in double quotes? Like `All "/home/foo/video files/my videos/what.mp4" configuration retries have failed.` (Vs: `All /home/foo/video files/my videos/what.mp4 configuration retries have failed.`)

src/Request/Internal.php
```diff
@@ -1146,8 +1231,8 @@ public function configureWithRetries(
                     // Instagram uses "ok" status for this error, so we need to check it first:
                     // {"message": "media_needs_reupload", "error_title": "staged_position_not_found", "status": "ok"}
                     if (strtolower($result->getMessage()) === 'media_needs_reupload') {
-                        throw new \InstagramAPI\Exception\UploadFailedException(sprintf(
-                            'Configuration of "%s" failed. You need to reupload the media (%s).',
+                        throw new \RuntimeException(sprintf(
+                            'You need to reupload the %s (%s).',
```
SteveJobzniak: Here, `$entity` can lead to messages like `You need to reupload the /home/foo/video files/my videos/what.mp4 (unknown error).` Hmm...

src/Request/Internal.php
```diff
@@ -1172,13 +1257,194 @@ public function configureWithRetries(
             sleep($delay);
         }
 
-        throw new \InstagramAPI\Exception\UploadFailedException(sprintf(
-            'Configuration of "%s" failed.',
+        // We are never supposed to get here!
+        throw new \LogicException(sprintf(
+            'Something went wrong during %s configuration.',
```
SteveJobzniak: `$entity` again.

src/Request/Internal.php
```diff
@@ -393,30 +418,90 @@ public function uploadSingleVideo(
...
+                $internalMetadata->getVideoDetails()->getFilename(),
+                ['targetFeed' => $targetFeed]
+            );
+            $photoFilename = $videoThumbnail->getFile();
+
+            try {
+                $internalMetadata->setPhotoDetails($targetFeed, $photoFilename);
+                $this->uploadPhotoData($targetFeed, $internalMetadata);
+            } finally {
+                @unlink($photoFilename);
```
SteveJobzniak: Hmmm... This part is not necessary, even though I agree that it's somewhat nice to be explicit.

As soon as `$videoThumbnail` is garbage-collected (destroyed), it will unlink its file. Remember? ;-) And that one takes care of only unlinking if the output-file differs from input-file path.

So here's the two choices:

1. Delete the `try{}finally{unlink();}` completely and let the destructor handle it.
2. Or, be explicit by calling `$videoThumbnail->deleteFile()`. It will automatically delete the file only if it differs from input path, and will reset the inner state so that `getFile()` will work again after that if we ever want to use it again.

I think option 2 sounds nice, since it's nice to explicitly free the temp file "early". What do you think?

SteveJobzniak:
Review is complete. Very clean work. Only two super tiny issues. 👍

valga:
Well, this particular PR is not suitable for squashing, so let me do something with it.

src/Request/Internal.php
valga: All configure exceptions will be wrapped into `UploadFailedException`, so `$entity` is not needed here.

src/Request/Internal.php
valga: All configure exceptions will be wrapped into `UploadFailedException`, so `$entity` is not needed here.

src/Request/Internal.php
SteveJobzniak: Ahh okay hehe :)

SteveJobzniak:
@valga Thanks, I confirm the fixes and will rebase and merge the individual commits now. Beautiful work. :-)

And thank you very much for rebasing and making the individual commits clean. :-)

PR #1685, opened 2017-11-09
Segmented video uploader

valga:
It works by splitting the video stream into a segments by time (5 seconds by default). The audio stream goes into its own segment (in one piece). After that, all segments are going to be uploaded with the resumable uploader.

Short videos (less than 10 seconds by default) aren't going to this uploader, just as an album videos.


valga:
It is ready for review.

There is another upload-related change (we need to provide `retry_context` to all uploaders), but it requires a lot of refactoring around our internal metadata (that is actually a good thing). I will address it later in a separate PR.

SteveJobzniak:
Hi. Fantastic news. But I am sick at the moment, I will get to this as soon as I feel better.

src/Media/Video/VideoDetails.php
```diff
@@ -235,6 +230,14 @@ public function validate(
         // users. Do NOT use it to read from the hard disk!
         $mediaFilename = $this->getBasename();
 
+        // Make sure we have found at least one video stream.
+        if ($this->_videoCodec === null) {
+            throw new \InvalidArgumentException(sprintf(
+                'Instagram requires video with at least one video stream. Your file "%s" doesn\' have any.',
```
SteveJobzniak: `doesn\'` misses a "t" :)

src/Request/Internal.php
```diff
@@ -1670,6 +1675,103 @@ protected function _uploadVideoChunks(
...
+     *
+     * @return \InstagramAPI\Response\GenericResponse
+     */
+    protected function _uploadSegmentedVideo(
+        $targetFeed,
+        InternalMetadata $internalMetadata)
+    {
+        $videoDetails = $internalMetadata->getVideoDetails();
+
+        // We must prepare a segments before running any requests.
```
SteveJobzniak: `a segments`? Perhaps say something like `We must split the video into segments`

src/Request/Internal.php
```diff
@@ -1670,6 +1675,103 @@ protected function _uploadVideoChunks(
...
+        // This request gives us a stream identifier.
+        $startRequest = new Request($this->ig, sprintf(
+            'https://i.instagram.com/rupload_igvideo/%s?segmented=true&phase=start',
+            Signatures::generateUUID()
+        ));
+        $startRequest
+            ->setAddDefaultHeaders(false)
+            ->addHeader('X-Instagram-Rupload-Params', json_encode($uploadParams))
+            // Dirty hack to make a POST request.
+            ->setBody(stream_for());
```
SteveJobzniak: This is fine, I guess... I think I even like it more than something like `setForcePost(true)` because this technique is less risky. If we have a flag we can screw that up. But we can't screw up this setBody. Thoughts?

src/Request/Internal.php
```diff
@@ -1725,6 +1827,69 @@ protected function _uploadResumableVideo(
...
+        try {
+            Utils::getFFmpegWrapper()->version();
+        } catch (\Exception $e) {
+            return false;
+        }
+
+        // There is no need to segment short videos.
+        $minDuration = $this->ig->getExperimentParam(
+            'ig_android_video_segmented_upload_universe',
+            'min_duration_threashold_sec_for_segmentation',
```
SteveJobzniak: `min_duration_threashold_sec_for_segmentation` Is that correct?

The correct spelling is `min_duration_threshold_sec_for_segmentation` but maybe Instagram's devs made a typo...

src/Request/Internal.php
```diff
@@ -1842,4 +2008,99 @@ protected function _getVideoUploadParams(
 
         return $result;
     }
+
+    /**
+     * Split the video file into a segments.
```
SteveJobzniak: `a segments`, remove "a". :)

src/Request/Internal.php
```diff
@@ -1842,4 +2008,99 @@ protected function _getVideoUploadParams(
...
+                escapeshellarg($videoDetails->getFilename()),
+                escapeshellarg(sprintf(
+                    '%s%s%s_1audio.000.mp4',
+                    $outputDirectory,
+                    DIRECTORY_SEPARATOR,
+                    $prefix
+                ))
+            ));
+        } catch (\RuntimeException $e) {
+            // Find and remove a segments (if any).
```
SteveJobzniak: `a segments`, say "all segments" :)

src/Request/Internal.php
```diff
@@ -1842,4 +2008,99 @@ protected function _getVideoUploadParams(
...
+            // Re-throw the exception.
+            throw $e;
+        }
+
+        // Collect segments.
+        $files = glob($pattern, GLOB_BRACE);
+        $result = [];
+
+        try {
+            // Wrap them into a VideoDetails.
```
SteveJobzniak: Hmm more with this `a VideoDetails` grammar habit xD The word "a" isn't right in that situation, unless some other word follows it, such as "a VideoDetails object". Otherwise, you just have to say "VideoDetails".

src/Request/Internal.php
valga: This typo is not mine. I just copied the string as is from the APK.

src/Request/Internal.php
valga: The only reliable way is to make request method implicit by changing `Instagram::request()` signature to something like:
```php
public function request(
        $method,
        $url)
```
But it would require a lot of changes in existing code.

SteveJobzniak: @valga Yeah let's not do that (a nightmare). The auto-guessing is fine. And the setBody trick is fine...

src/Request/Internal.php
SteveJobzniak: Okay, so as I suspected, it was their typo, hehe.

valga:
I fixed all grammar problems.

SteveJobzniak:
@valga Thank you so much. Beautiful work.

PR #1695, opened 2017-11-14
Realtime: Refactor everything

valga:
Instagram is testing dynamic realtime subscriptions: for example, one can subscribe to some media stats (such as likes or video views counters) update or to notifications about people typing comments. Also, some new direct message commands were introduced.

But adding these new features will cause a lot of pain with current codebase, so this PR aims to make the Realtime client easier to maintain. Refer to the commits list for more details.

BC breaks:
1. All Realtime payload has been moved into a `\InstagramAPI\Realtime\Payload` namespace.
2. First parameter for the `unseen-count-update` event is now `$inbox`.
3. The `Realtime::updateSequenceId()` method has been renamed to `receiveOfflineMessages()` to match its purpose.


valga:
Alright, it's time to review and merge it.

src/Instagram.php
```diff
@@ -951,6 +951,25 @@ public function isExperimentEnabled(
...
+     * @param string $param
+     * @param mixed  $default
+     *
+     * @return mixed
+     */
+    public function getExperimentParam(
+        $experiment,
+        $param,
+        $default = null)
+    {
```
SteveJobzniak: This code will fail in a special scenario:

Param exists and was given the value `null` by Instagram. And `$default` is non-null.

In that case, this code will see Instagram's null and isset will be false, then it chooses the non-null default value.

To protect against that, it needs something like `isset($this->experiments[$experiment]) && is_array($this->experiments[$experiment]) && array_key_exists($param, $this->experiments[$experiment])`. Which is sadly three function calls.

Perhaps it is not worth worrying about. I don't think Instagram ever uses NULL.

valga: We must choose the default value when experiment parameter is missing or its value is `null`. No problem here.

SteveJobzniak: Okay, excellent! 👍 :-)

SteveJobzniak:
@valga Oops, I just saw your comment that it's time to review and merge this. I didn't notice that comment until now. Is it still true? If so I will review after you confirm that it's still ready for review. :)

valga:
Yep, it is still ready for review.

SteveJobzniak:
Okay sweet! ❤️  I will review tomorrow. Can barely stay awake today. Sick ;(

mgp25:
Mobile did this, not me! Please reopen and revert the push. Sorry

SteveJobzniak:
@mgp25 lol what the hell... a squashed merge ;( not even the rebase kind...

I will open the computer and force-push the previous commit to remove the merge...

mgp25:
Yep, sorry about that, i was looking the feed in the ios app and it seems my finger is to big and pressed by mistake the merge mini button

Sorry

SteveJobzniak:
@mgp25 I have reverted and force-cleared the history from the merge. So now it's clean again :S

@valga It seems like fat-fingered (heh) mgp has made it impossible to use this pull request. There doesn't seem to be any way to reopen it, and it seems like a new pull request from the same branch is needed.

mgp25:
Im sorry if i caused any trouble to any of you @valga @SteveJobzniak. As said before, pressed by mistake of the ios app.

SteveJobzniak:
@valga On another note: LazyDoctor puts the class name (such as `* LivePayload.`) as the class summary whenever it documents a class that had no doc block (or has no summary line). But after that, it sees a summary line so it won't update/insert the class name again if it changes in the future.

I thought about it before, when writing the code, but didn't think it was a big problem. But now this pull request renames some classes and their summaries have the old class names.

I suppose I can make a regex in LazyDoctor which checks if the 1st line (summary) is exactly the format: `CamelCaseName<period>`, and in that case erase the old line and insert the latest classname. What do you think about the risk of false positives? I don't think people document their classes with `* Lol.` or whatever (a single word with no spaces, followed by a period), so the risk of false positives is very low... Hmm...

What do you think? I can quite easily add a regex to make it fix these renamed class names...

valga:
@SteveJobzniak Everything that has been generated automatically must be also updated automatically.

SteveJobzniak:
@valga I agree but how do I know if `* ClassName.` was added automatically vs a guy who just decided to write `* Functions.`? ;-) That's the reason I didn't auto-update it. But I think the issue is very tiny. Because:

1. We only auto-document CLASSES.
2. We only auto-document classes derived from LazyJsonMapper.
3. People would probably only use `* Functions.` if they ever write a 1-word summary. And there's no risk that they'll put a function-collection in a class that's derived from LazyJsonMapper. I can't remember personally having ever seen or written a 1-word class summary.
4. Again, since it's only doing this to classes derived from LazyJsonMapper, it seems totally safe.

So yes I will write and release a small update to LazyJsonMapper now, to add auto-updating of the summary line when it's just a single word followed by a period... If there are any false positives, that's less of an issue than outdated class names in the summary.

SteveJobzniak:
@valga Okay, https://github.com/SteveJobzniak/LazyJsonMapper/releases.

Please run `composer update -o` and update the new pull request with the fixed names that it produces. We'll talk over there. I need to go back to bed now (I woke up at 3 AM)... And tomorrow I will review the code!

PR #1696, opened 2017-11-14
Request: Update default headers

valga:
It has been almost 3 months since we found them, but we still have nothing.


SteveJobzniak:
Thanks, you're right, we should at least have these new defaults!

src/Request.php
```diff
@@ -365,9 +365,13 @@ public function addHeader(
     protected function _addDefaultHeaders()
     {
         if ($this->_defaultHeaders) {
+            $this->_headers['X-IG-App-ID'] = Constants::FACEBOOK_ANALYTICS_APPLICATION_ID;
             $this->_headers['X-IG-Capabilities'] = Constants::X_IG_Capabilities;
             $this->_headers['X-IG-Connection-Type'] = Constants::X_IG_Connection_Type;
             $this->_headers['X-IG-Connection-Speed'] = mt_rand(1000, 3700).'kbps';
+            $this->_headers['X-IG-Bandwidth-Speed-KBPS'] = '-1.000';
+            $this->_headers['X-IG-Bandwidth-TotalBytes-B'] = '0';
+            $this->_headers['X-IG-Bandwidth-TotalTime-MS'] = '0';
```
crouchingtiger: @valga You can see what these are being used for here: https://engineering.instagram.com/improving-performance-with-background-data-prefetching-b191acb39898

SteveJobzniak: @crouchingtiger Ohh great discovery, thanks! So it seems like they are using it to collect worldwide connection/usage metrics to check how quickly content is being delivered (in ms) and how large the replies are.

PR #1707, opened 2017-11-24
Request/Live: New functions enableComments() & disableComments()

itsMoji:


valga:
@mgp25 Could you please verify that?

mgp25:
@valga Verified, this is okay.

SteveJobzniak:
@itsMoji Thanks! <3

@valga That's intentional, for a few reasons:

- Properties can come from many places. They are either added to the class map by the class itself, or inherited from parents, or imported from other classes (via `OtherClass::class` import commands), or from parents of those other classes. To massively reduce memory usage with all this property duplication going on, all properties have as little object fields as possible, and any properties that are identically defined/or inherited/or imported are shared (reused) between multiple classes so that there's only 1 definition-instance for that property in memory. Therefore, properties have no "Owner" field (since tons of different classes can use the same property instance), so there's no way to know if a property came from the class itself, its parent, or from an import, or from an import's parent. Therefore, we literally can't know where the properties came from. Unless we add some more metadata which points to the source of the definition, but then we increase memory usage.
- It's also intentional because it's a very good idea: Almost nobody except hired professionals use something like PHPStorm or NetBeans. Most people use basic editors like Notepad++, Vim, Emacs, Atom, etc. They have no class method/property autocompletion (some can do it with plugins but most don't install that). So those people literally have to read the .php files to see what functions are available in each class. That's why I decided to include all available methods on each class.
- Lastly, the autodocumenter has a special `-o` flag which means "document functions even when they have been overridden by the class itself". We run with that, so that even if we define a function `getMessage`, we'll still see `getMessage` in the list of `@method` too. Why? Because it means that everything which inherits from Response.php will show `getMessage` as an available method, even though that particular one has been overridden in the base. Otherwise people would see getters for everything except the message.

Basically: IDEs that do autocompletion ignore all of these extra duplicate declarations. Users on the other hand are greatly helped by them.

PR #1711, opened 2017-11-27
Add pagination to the blocked list

svetlana-1604:


SteveJobzniak:
Hi Svetlana, thanks for doing this work and sharing it!

However, two things are needed:

- Add maxid parameter documentation. Look at the other functions that have a maxId param, and copy it. It should be above the `@throws`, with a blank line between the `@param` and the `@throws`. Again, look at the other functions.
- Run `composer codestyle`. I see that you manually wrote the new properties into the response file, and you missed one. Just run the command and it will be done automatically. And be sure you have `dev-master` when running the command, otherwise you will run an outdated codestyle command.

More help is available here:

https://github.com/mgp25/Instagram-API/blob/master/CONTRIBUTING.md

svetlana-1604:
Hi Steve,

Thanks for the comments. I've made the required changes.

SteveJobzniak:
Thanks a lot for your contribution! And thanks for fixing the variable type in the `@param`. ;-)

PR #1723, opened 2017-12-04
Requests: Mass update

valga:
Based on `22.0.0.15.68` version.

BC breaks:
- The following methods are removed: `Direct::getShareInbox()`, `Direct::getRecentRecipients()`, `Discover::getPopularFeed()`, `Discover::getHomeChannelFeed()`, `People::getMutedStoryList()`.
- The `People::searchFacebook()`, `People::getSuggestedUsersFacebook()`, and `People::getRecentFacebookSearches()` methods are renamed to `Discover::search()`, `Discover::getSuggestedSearches()`, and `Discover::getRecentSearches()` respectively.
- The `Location::searchFacebook()`, and `Location::searchFacebookByPoint()` methods are renamed to `Location::findPlaces()`, and `Location::findPlacesNearby()` respectively.
- Rank token now matches the app behavior. There are 3 types of rank token:
    1. Local UUID. You must generate it with `Signatures::generateUUID()` and use for all paginations and/or searches on the same page. Affected methods: `Hashtag::getFeed()`, `Location::getFeed()`, `People::getFollowing()`, `People::getFollowers()`.
    2. Remote UUID. You must use a value taken from a previous call response. Affected methods: `Hashtag::search()`, `Location::findPlaces()`, `Location::findPlacesNearby()`, `People::search()`, `Discover::search()`.
    3. Internal (means you don't have to worry about it). It has good old `%ACCOUNT_ID%_%UUID%` format, used just for `location_search/` endpoint.

Improvements:
- New methods `Direct::getThreadByParticipants()`, `Direct::markVisualItemsReplayed()`, and `Discover::clearSearchHistory()`.
- It is now possible to paginate over `Hashtag::search()`, `Location::findPlaces()`, `Location::findPlacesNearby()`, `People::search()`, and `Discover::search()` results.


SteveJobzniak:
Wow this is absolutely incredible. Incredible. Incredible! ❤️

valga:
@SteveJobzniak Hi. I need you to write a proper documentation for all new `$rankToken` params.

SteveJobzniak:
@valga Hi. Okay, but since I have packed down my computer 2 days ago to live without as much electronics, to help my health, it is kinda hard for me to do that. ;O I have been around mostly as support/idea/discussion guy since mid-december when I "left", and now all personla projects are done too, so the computer is down. This communication now and recent merges have been mostly done from iPad.

But anyway, the documentation seems like a good candidate for the extradocs folder (which we made for any long docs that are too complex to explain in the source code), since the doc for rankToken will be long. Your "BC Breaks" summary in the first post is already very good. Just needs bit more info about what you mean by "from a previous response" (ie add some example of how to get those tokens).

Simply make a tzt file and link to that in the same way we linked to extradocs from another function:

https://github.com/mgp25/Instagram-API/blob/383be65522e556a25c9aaf8a08945e7312f5a14c/src/Request/Push.php#L62

Seems ok?

valga:
@SteveJobzniak I made some examples, but I see no reason in extradocs for rank token, because it's just one line of code: you either generate random UUID or take it from the previous response.

SteveJobzniak:
Yeah you made some very good examples! I saw them. :-)

But somehow users need to be pointed at them or at an extradoc which explains it all, because the current comments and the fact that they are in so many functions now, will cause approximately 1 new user-ticket per day asking how to use those parameters. Hehe. Working on opensource has taught me that there are lots of idiots in the world and that they _all_ love to waste developer time asking us for help - unless we are super clear in the code (and even then there's always someone who asks anyway). So since the new system is a bit complex ( ut necessary, I like your changes!), it needs to lead the user to the usage-example files so that people can find the code for how to use those params.

I can think of three possible ways solve it: an extradocs file which explains it (would be the clearest way to avoid user confusion, since it allows to write some nice, longer explanation about the system). Or a `@see` on all ranktoken which tells the user to view the code example (may not be possible, since see-tags require a classname or http URL to be valid). Or a longer explanation in the `@param` (kinda nasty but we have some other long, repeated param descriptions so it may be okay).

valga:
@SteveJobzniak I decided to use `@see` tags to point to rank token sources and examples.

And it is ready for review.

SteveJobzniak:
@valga That is great, I will review tomorrow.

Two small notes now:

- At least one of the ranktoken `@param` is marked as `@param null $rankToken`.
- I like the use of `@see`, but as I mentioned earlier it seems like `@see` does not support `@see examples/paginate.....`. Read: http://docs.phpdoc.org/references/phpdoc/tags/see.html. Luckily we don't truly care about valid tags (so that we can generate compiled phpdoc html files for this library) so I guess we can keep the `@see` method of documenting these.

Tomorrow I will collect the energy to review properly :)

valga:
@SteveJobzniak 
- Just one. I fixed it.
- Well, `examples/paginateWithExclusion.php` is a valid relative URI, so there is no issue here. Also see `https://github.com/phpDocumentor/phpDocumentor2/issues/757`

SteveJobzniak:
@valga Reviewed and merged as a single commit in https://github.com/mgp25/Instagram-API/commit/a640aae361cb895c64234d453b416e5044c33ae7, with all of your 1st post info put into the squashed commit! :-)

I had some questions, but decided to merge anyway:

- What's `recovered_from_crash`? With this new change, every "refresh login flow" sends that flag. Will that make our emulator appear as if it constantly comes back from a crash? Or perhaps the real app uses that when it gets killed by Android's memory manager... Just asking in case we're looking weird by always saying that we are returning from a crash. ;-)
- I thought about suggesting something like `rankTokenUsage = localRankTokenUsage` and `paginateWithExclusion = remoteRankTokenUsage` for the example-files, since they demonstrate local and remote rank-tokens... but it would have involved more edits than what I can do via GitHub's web GUI...

When reviewing this, I actually learned a little trick from you: `if (isset($options['recovered_from_crash']) && $options['recovered_from_crash']) {` is equivalent to `if (!empty($options['recovered_from_crash'])) {`. I had no idea that `empty()` can take missing variables exactly like `isset()`. Heh. Nice!

Anyway, I am very impressed with this huge update to the library's compatibility and correctness. You've done an incredible task here! It's seriously impressive. All this rank-token handling brings this library to a new level of correctness.

@mgp25 We should now update the Contants.php to `22.0.0.15.68` (which this PR was based on). I don't have those constants though. I have 25 and newer. So if you have time someday, please extract the vars from that particular version (IG_VERSION, VERSION_CODE, IG_SIG_KEY, X_IG_Capabilities, LOGIN_EXPERIMENTS and EXPERIMENTS).

Also, @valga... just so that you know, I've noticed that my health - even though it's doing a bit better - still isn't good enough to even allow me to hang around to help with any more reviews. I did it this time because I am grateful for your awesome work, and that's why I've still been around for a while... but every review can drain all of my energy for one or more days and give me health setbacks. It's my own fault for feeling responsible to help even though I have a serious health situation. But I need to continue to leave my electronics behind, more and more, to have a chance at any form of recovery. So you may still see me chat from time to time, but I am going to be in the background now without doing reviews. Wish me luck with my health. It seems to be going better and better, but closing the computer was the first step. I need to rest and not have any projects or responsibilities, so I'll have to become a ghost now. Take care bro. I wish you and mgp a very merry christmas! :-)

valga:
> With this new change, every "refresh login flow" sends that flag.

This flag is being sent just after logging in. And yes, it means that the app was crashed or killed or force closed or... just installed.

> I thought about suggesting something like

Initially I named them `localRankTokenPagination` and `remoteRankTokenPagination` but then renamed to their current names, because it makes more sense. 

It is sad to hear that you are leaving us, but your health is more important, good luck with it. I will certainly miss your reviews. And by the way, can you give me or mgp a push access to the LJM repo?

SteveJobzniak:
Hi!

> This flag is being sent just after logging in. And yes, it means that the app was crashed or killed or force closed or... just installed.

Okay so it's as I hoped. That is perfect. :-) Funny Instagram with their weirdly named variables.

> Initially I named them localRankTokenPagination and remoteRankTokenPagination but then renamed to their current names, because it makes more sense.

Yeah, paginating with exclusion uses remote rank token so `paginateWithExclusion` makes sense, and `rankTokenUsage` makes sense that it's for locally generated tokens. It's fine. :-)

> It is sad to hear that you are leaving us, but your health is more important, good luck with it.

Thank you. I am making progress but it's slower than I'd like it. The programming has been a very helpful and fun way help with stress but now I need to live without electronics to truly rest my mind and body. There's a good chance of recovery but it's going to take time and truly resting now...

> I will certainly miss your reviews.

Thanks, well at least I can say that your work is super high quality and has needed less and less reviewing, so I would fully trust you unreviewed on any project I ever work on. :-)

> And by the way, can you give me or mgp a push access to the LJM repo?

And about that! Yes, of course you can have push access to LazyJsonMapper! :-) I will send the invite now, and I also sent one to mgp in case something happens to you. I just need to request that you (both) truly learn the codebase and follow the code style and documentation style of that project if you commit anything, since I spent a _lot_ of time and energy carefully making it fully documented and modular and stable. There's a lot of comments that explain each step in the code and why it's necessary and why somethings are critically important etc. But of course I trust that if you commit anything, you'll do it after becoming very familiar with the code. I remember your earliest contributions here at this library and you were great from the beginning. You're very intelligent. :-)

Invite sent!

PR #1759, opened 2017-12-30
Added Facebook login functionality

charlie759:
Added the ability to login via Facebook.  I hope this is helpful for some users.


Halcyonflow:
Wow, good for you for the effort, but this is really terribly written.

1. massive code duplication. why not just add something like a `$ig->login_type = 'regular'; // or 'facebook'` and use that variable for doing some slightly different actions where necessary. instead of literally duplicating 3 huge functions just to make one-liner changes to each of them?
2. it doesn't set `$ig->username` value. but search in the source code for `->username`. that param is critical in many places. you can't just leave it unset. it must be set.
3. it stores the token as a username in the database as-is. should at least prefix it something like `'facebook.'.$fb_access_token` to separate it from regular usernames.
4. it uses `addPost('waterfall_id', $this->waterfall_id)` but there is no such thing as `$this->waterfall_id`. it doesn't exist. (and shouldn't exist. waterfall ids are not meant to be permanent. from my own research they are just used to link multi-step form actions in the real app to identify them all as part of the same request.)
5. it attempts to store `waterfall_id` in the database with `$this->settings->set('waterfall_id', Signatures::generateUUID(true));` but the database does not have that field, so that line will throw an exception.
6. the endpoint it calls for "login" is `'fb/facebook_signup/'` which looks like trouble. "signup"? So what about people who try to login with facebook accounts who have never loged in before? I guess the endpoint will throw them to some "create your username" endpoint, but this patch doesn't implement that.

Sorry this is really bad... 👎

atyachin:
Thanks! this would be very useful for us.
Would you please provide an example of how you get the $fb_access_token ?
(I assume that the user successfully finished the OAuth process in his browser, but not sure how to get the code from the redirection URL - which is a Facebook/Insta URL..). Maybe I'm missing something?

nemanjan00:
@atyachin, I think taking a look at code of tind3r extension for chrome would be useful in this case. 

It does the same thing. 

(I do not think it is opensource but you can just unzip it and read the code)

atyachin:
@nemanjan00, Thanks!. A browser extension can definitely capture the OAuth authorization code, however it requires the user to install an extension - something I'd like to avoid... I'm trying to figure out a way to capture the authorization code from server side only.

@charlie759 - would you please elaborate on the use case you had in mind on how to get the $fb_access_token ?

charlie759:
@Halcyonflow I agree actually, this is pretty terribly written, sorry about that.  PHO isn't my strong point by a mile, maybe I shouldn't have sent the PR.  I might close this, fix the issues, and then make a new one.  About the username; it cannot be set until after login as most people would login with their Facebook email address but I'm pretty sure that should be ok.

@atyachin Take a look at: https://developers.facebook.com/docs/facebook-login/web/accesstokens and the docs in general.

nemanjan00:
@atyachin, in that case, you also need to login to facebook on a server...

Halcyonflow2:
@charlie759 I think it's best to keep it open if you're not experienced in PHP. That way everyone can discuss the ideas and work together.

About the username: It is going to be hard to get it. Most logins do a "resume" which just calls a single API: "get timeline". Which I just checked does not tell you your account's username. So during almost every login, the only thing which knows the username is the `login($username, $password)` call the account owner does... so facebook logins are screwed.

But I researched a bit. Turns out the cookie jar may be able to solve this...?

This one which expires on March 19th, 2018:

`{"Name":"ds_user","Value":"myusernamehere","Domain":"i.instagram.com","Path":"\/","Max-Age":"7776000","Expires":1521419865,"Secure":false,"Discard":false,"HttpOnly":false}`

This one which expires on January 1st 2018:

`{"Name":"igfl","Value":"myusernamehere","Domain":"i.instagram.com","Path":"\/","Max-Age":"86400","Expires":1514803376,"Secure":false,"Discard":false,"HttpOnly":false}`

It seems like the best one is `ds_user`. It has the longest duration and has a name that suggests to me that its purpose is to store the username.

Please check your cookies for your facebook login account and see if the ds_user value has your correct Instagram username. Then I think all you need is to add code for these two things:

1. After the user cookie jar is loaded (probably at the bottom of `_setUser()` after the Client has been updated?), try to read the `ds_user` cookie if it exists and use it to set `$this->username`. There's a `getCookie()` function in the client. But beware, there may be risks about reading outdated cookies containing the wrong username which would be bad. So only read the username from the client cookies if `$this->login_type` is `facebook`. Otherwise keep the accurate one the user provided to `login()` as usual. Alternatively, perhaps it should simply make the "$this->username = getCookie('ds_user')" call in _sendLoginFlow() in all cases, at the bottom after the regular login/resume process is complete... because then we know for sure that the cookie value has been updated by Instagram and is not stale from an old cookie jar...
2. When you call the `fb/facebook_signup/` function, the endpoint response probably contains the username that just logged in. So read it from there in that case. Unless you use the alternative method above and just read the username from the cookies at the end of the login flow...

I don't know exactly how code for that would best look like. Someone more involved in this project may tell you if my idea is good or not. But I think the above would solve it: Full logins that call the `fb/facebook_signup/` endpoint would get the username from the response. Resumed facebook sessions would read the username from the cookies. All other logins would get it from `login($username, $password)` as usual...

charlie759:
The `ds_user` cookie is set on login with Facebook as shown below:

```
[csrftoken=131fU8qqZLJP3XRtep4HRnjHtmU9sFfj; expires=Sun, 30 Dec 2018 11:12:20 GMT; path=/; secure, ds_user_id=276431532; expires=Sat, 31 Mar 2018 11:12:20 GMT; path=/, sessionid=CENSORED; expires=Sat, 31 Mar 2018 11:12:20 GMT; path=/; secure; httponly, ds_user=nieks759; expires=Sat, 31 Mar 2018 11:12:20 GMT; path=/, urlgen="{\"time\": 1514718742\054 \"94.0.218.23\": 5607}:1eVbXS:UgcJN-314c8ql6hYnAoB4EXi1KA"; path=/, mid=WkjGFQABAAF3q6sU4HloH0oLdiUC; expires=Sat, 26 Dec 2037 11:12:20 GMT; path=/, rur=ATN; path=/]
```

The username is also attainable from the login response.  I think what you suggested should definitely work.

mgp25:
@SteveJobzniak @Halcyonflow2 (steveJobzniak) you can login with your normal account, no need to use secundary accounts, just saying

charlie759:
@mgp25 Totally your decision whether or not to add it, I just thought it would be a nice feature to have.

mgp25:
@charlie759 the feature is okay, but it need to be improved so it fits perfectly with the project, unfortunately i dont have time to do any coding at the moment. You can post all information (documentation) here, the changes and everything, and when its perfect we'll merge it.

charlie759:
@mgp25 OK, sounds good to me.

charlie759:
This is just another note, even though the PR hasn't been fixed yet/isn't ready to be merged, you **MUST** use the Instagram App ID when initializing the Facebook Login Javascript SDK which can be found in `Constants.php`. It is: `124024574287414`.  This ensures that you get the correct access token.

davisjackie321:
Cool

shadowc0de:
@charlie759 , how the next with this? i'm interest

bolds07:
any progress?

atyachin:
bolds07 - the FB login would indeed be very useful feature.
As you mentioned, it seems that the main challenge is getting the Facebook token.
Could you suggest any practical method of getting that token that doesn’t involve installing a browser extension?
(the main challenge I see is that the code is returned only in a trusted redirection URL provided by Facebook, and I see to easy way getting in the middle...)
Any direction here would be appreciated.

bolds07:
@atyachin You need to have a man in the middle.

I think the most elegant way would be work on React API of facebook https://developers.facebook.com/docs/react-native/login and find a way to port the required functions to work on browser

But is also possible to any client side login on facebook sdk
https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow

This way instead of giving the oauth screen to the user,, you need to get the information from user and interact with the flow by yourself...

For sure it will demand work since it may lead to two-factor-login or any kind of challenge from facebook private API (since the ip wont be the client's)

But this repository reversed almost whole instagram private API, it wont much different to handle the facebook oauth mechanism... And it worth (in my opinion) since although almost whole api is reversed you still lack of the most basic function... login

aaddaa:
It was
image

it became
image

b07n37:
What is the issue getting FB token? This is just a POST with login/pass and specific parameters. Would be nice to keep it rolling...

b07n37:
are those style issues blocking the PR? @charlie759 did you test the whole flow actually?

mgp25:
@b07n37 his changes were almost 2 years ago. This code would need a full review and rebase

PR #1762, opened 2018-01-02
Fixing API headers

BaderMudarra:
Based on most recent version for android # 26.0.0.13.86
The following two headers are not being sent any longer, so I removed them.
* X-FB-HTTP-Engine
* Accept

Also, `Accept-Encoding` header is sent with `gzip` value instead of `gzip,deflate`


aitte2:
https://github.com/mgp25/Instagram-API/blob/d28ab4803843310f32e328e6742c05ba55d16555/src/Constants.php#L12

mgp25:
Yes, in the current version, headers still exist. When we add everything that is in that version, we'll merge this PR

Regards

davisjackie321:
Yea

aitte2:
@mgp25 hi. Isnt this PR needed  after https://github.com/mgp25/Instagram-API/commit/d0d289422b933d261c03e173d548434bd790bcb6 ? Or are the headers back to being the same again?

PR #1785, opened 2018-01-20
Archive class

mgp25:
Group all archive related functions into this class


mgp25:
@valga let me know if this is okay. Also is it okay for you if i start writing the summary for the release 4? Most of the issues are because people using outdated release.

valga:
@mgp25 Hi, it is ok to prepare a new release. I will review the PR later.

src/Request/Collection.php
```diff
@@ -43,14 +43,7 @@ public function create(
...
     {
-        return $this->ig->request('collections/create/')
-            ->addPost('module_name', $moduleName)
-            ->addPost('added_media_ids', json_encode($mediaIds))
-            ->addPost('name', $name)
-            ->addPost('_uuid', $this->ig->uuid)
-            ->addPost('_uid', $this->ig->account_id)
-            ->addPost('_csrftoken', $this->ig->client->getToken())
-            ->getResponse(new Response\CreateCollectionResponse());
+        return $this->ig->archive->createCollection($name, $mediaIds, $moduleName);
```
valga: Since we are going to release new major version, it is ok to break BC by removing the whole `Collection` class.

src/Request/Archive.php
```diff
@@ -0,0 +1,214 @@
...
+     * Edit the name of a collection.
+     *
+     * @param string $collectionId The collection ID.
+     * @param string $name         New name for the collection.
+     *
+     * @throws \InstagramAPI\Exception\InstagramException
+     *
+     * @return \InstagramAPI\Response\EditCollectionResponse
+     */
+    public function editCollection(
```
valga: Let's merge addMediaToCollection() and removeMediaFromCollection() methods into this one, so its signature will change to:
```php
public function editCollection(
    $collectionId,
    array $params)
```
Where `$params` is an associative array with `name`, `add_media` and `remove_media` keys.

src/Request/Story.php
```diff
@@ -151,6 +151,70 @@ public function getReelsMediaFeed(
...
+     * @param mixed  $mediaIdForCover One media ID in Instagram's internal format (ie ["3482384834_43294"]).
+     * @param array  $addedMediaIds   One or more media IDs in Instagram's internal format (ie ["3482384834_43294"]).
+     * @param array  $removedMediaIds One or more media IDs in Instagram's internal format (ie ["3482384834_43294"]).
+     * @param mixed  $module          (optional) From which app module (page) you're performing this action.
+     *
+     * @throws \InstagramAPI\Exception\InstagramException
+     *
+     * @return \InstagramAPI\Response\ReelsMediaResponse
+     */
+    public function editHighlight(
```
valga: There is no need to duplicate methods. Let's remove it.

src/Request/Story.php
```diff
@@ -151,6 +151,70 @@ public function getReelsMediaFeed(
...
+     * @param array      $mediaIds        Array with one or more media IDs in Instagram's internal format (ie ["3482384834_43294"]).
+     * @param string     $title           Title for the highlight.
+     * @param null|mixed $mediaIdForCover One media ID in Instagram's internal format (ie ["3482384834_43294"]).
+     * @param mixed      $module
+     *
+     * @throws \InstagramAPI\Exception\InstagramException
+     *
+     * @return \InstagramAPI\Response\ReelsMediaResponse
+     */
+    public function createHighlight(
```
valga: It is better to replace this method with `addToHighlight()` with the following signature:
```php
public function addToHighlight(
    $storyId,
    $highlightId,
    array $options = [])
```
And then add `@see Archive::createHighlight() To create a highlight.`

src/Request/Archive.php
```diff
@@ -0,0 +1,214 @@
...
+
+/**
+ * Archive-related functions.
+ */
+class Archive extends RequestCollection
+{
+    /**
+     * Create a highlight reel.
+     *
+     * @param string     $userId          Numerical UserPK ID.
```
valga: We don't have such param, also in other methods.

src/Request/Archive.php
```diff
@@ -0,0 +1,214 @@
...
+     * @param array      $mediaIds        Array with one or more media IDs in Instagram's internal format (ie ["3482384834_43294"]).
+     * @param string     $title           Title for the highlight.
+     * @param null|mixed $mediaIdForCover One media ID in Instagram's internal format (ie ["3482384834_43294"]).
+     * @param mixed      $module
```
valga: `$module` should be a string. Also, do you have a complete list of allowed modules?

And you have a lot of `mixed`s in other params. I think all of them are actually strings.

src/Request/Archive.php
```diff
@@ -0,0 +1,214 @@
...
+     *
+     * @throws \InstagramAPI\Exception\InstagramException
+     *
+     * @return \InstagramAPI\Response\ReelsMediaResponse
```
valga: Wrong return type, also in other methods.

src/Request/Archive.php
```diff
@@ -0,0 +1,214 @@
...
+     */
+    public function createHighlight(
+        array $mediaIds,
+        $title,
+        $mediaIdForCover = null,
+        $module = 'story_viewer')
+    {
+        if ($mediaIdForCover === null) {
+            $mediaIdForCover = $mediaIds[0];
```
valga: You don't check for array length, so one can easily pass an empty array, and it would crash here.

src/Request/Archive.php
```diff
@@ -0,0 +1,214 @@
...
+        }
+        $cover = [
+                    'media_id'  => $mediaIdForCover,
+                    'crop_rect' => '[0.0, 0.19543147, 1.0, 0.8045685]',
```
valga: Are these numbers fixed? May be it is better to use `mt_rand()`? Also is it really going to be double JSON-encoded?

src/Request/Archive.php
```diff
@@ -0,0 +1,214 @@
...
+                ];
+
+        return $this->ig->request('highlights/create_reel/')
+            ->addPost('source', $module)
+            ->addPost('_uuid', $this->ig->uuid)
+            ->addPost('_uid', $this->ig->account_id)
+            ->addPost('_csrftoken', $this->ig->client->getToken())
+            ->addPost('title', $title)
```
valga: You don't check for title length. Is it ok to have an empty one?

src/Request/Archive.php
```diff
@@ -0,0 +1,214 @@
...
+            ->addPost('cover', json_encode($cover))
+            ->addPost('added_media_ids', json_encode($addedMediaIds))
+            ->addPost('removed_media_ids', json_encode($removedMediaIds))
+            ->getResponse(new Response\HighlightFeedResponse());
+    }
+
```
valga: There is no `deleteHighlight()` method here. I checked the app, and it has the `highlights/%s/delete_reel/` endpoint.

src/Request/Archive.php
```diff
@@ -0,0 +1,214 @@
...
+     */
+    public function editHighlight(
+        $highlightReelId,
+        $title,
+        $mediaIdForCover,
+        array $addedMediaIds = [],
+        array $removedMediaIds = [],
+        $module = 'story_viewer')
+    {
+        if ($mediaIdForCover === null) {
```
valga: You are changing the cover in every edit. Is it ok? I mean, does the app always send this param or only when changed?

src/Request/Archive.php
```diff
@@ -0,0 +1,214 @@
...
+     * @param mixed  $mediaIdForCover One media ID in Instagram's internal format (ie ["3482384834_43294"]).
+     * @param array  $addedMediaIds   One or more media IDs in Instagram's internal format (ie ["3482384834_43294"]).
+     * @param array  $removedMediaIds One or more media IDs in Instagram's internal format (ie ["3482384834_43294"]).
+     * @param mixed  $module          (optional) From which app module (page) you're performing this action.
+     *
+     * @throws \InstagramAPI\Exception\InstagramException
+     *
+     * @return \InstagramAPI\Response\ReelsMediaResponse
+     */
+    public function editHighlight(
```
valga: Too much params for one method. I think it is better to change the signature to
```php
public function editHighlight(
    $highlightReelId,
    array $params,
    $module = 'story_viewer')
```
And `$params` will have `title`, `cover_media_id`, `add_media` and `remove_media` keys.

src/Request/Archive.php
```diff
@@ -0,0 +1,214 @@
...
+            ->addPost('cover', json_encode($cover))
+            ->addPost('media_ids', json_encode($mediaIds))
+            ->getResponse(new Response\HighlightFeedResponse());
+    }
+
+    /**
+     * Edit a highlight reel.
+     *
+     * @param string $userId          Numerical UserPK ID.
+     * @param mixed  $highlightReelId Highlight ID, using internal format (ie ["highlight:12345678901234567"]).
```
valga: This param goes into request URI. Why is it marked as mixed (should be a string) and why did you mention array as an example?

src/Request/Story.php
aitte2: Great review (all of these comments)! But don't you mean `@see Archive::addToHighlight() To create a highlight.` here?

src/Request/Archive.php
mgp25: Fixed :white_check_mark:

src/Request/Archive.php
mgp25: Fix doc. :white_check_mark:

Yes, all of them are strings or string[].

I don't have a complete list, but i'll try to provide one.

src/Request/Archive.php
mgp25: Fixed. :white_check_mark:

src/Request/Archive.php
mgp25: Fixed. :white_check_mark:

src/Request/Archive.php
mgp25: I have to check wether this are fixed or not. But so far, it took the same crop properties in each time, so i considered as a default crop. Will recheck this.

Yes, it is double JSON-encoded. That is right.

src/Request/Archive.php
mgp25: Fixed. :white_check_mark:

I

src/Request/Archive.php
mgp25: Fixed. :white_check_mark:

If you set an empty one, a default name will be set `Highlights`, this is what Instagram does. Max characters 16.

src/Request/Archive.php
mgp25: Yes, this is okay

src/Request/Archive.php
mgp25: I have done the changes, i don't know if the style is okay. If you could provide me some feedback on this, would be great.

src/Request/Story.php
mgp25: Fixed. :white_check_mark:

src/Request/Archive.php
mgp25: Added. :white_check_mark:

src/Request/Collection.php
mgp25: Done. :white_check_mark:

src/Request/Archive.php
mgp25: @valga there is already a function called `editCollection()` for a different purpose. Any idea?

aitte2:
What is a story highlight and what is a media collection? Do they really belong together in one class? Seems like one is a way to save story highlights, and one is a way to make collection-groups of liked media? I am confused :p

mgp25:
@aitte2 

- **Story highlight:** Creates archives (highlights) where you can save stories. It is accessible from the user feed. This is public and can be seen by anyone.

- **Collections:** Creates archives where you can save any media (except stories). It is accesible from your own user feed. This is private to each user.

Since both features, are doing something in common (archive stuff), @valga proposed it would be better to have all related function into a single class.

aitte2:
Hmm ok... @valga Hmm can you comment on this? https://help.instagram.com/274531543007118

The feature is officially named Collections, shouldn't we keep that in `$ig->collections`?

And story highlights could go as 2 functionss in $ig->story?

valga:
@aitte2 Somehow I thought that the app has one module for both highlights and collections. I just checked that, and I was wrong: they belong to different modules.

@mgp25 I am terribly sorry, but you have to bring `Collection` class back and leave it as is.

> And story highlights (archiving) could go as 2 functionss in $ig->story?

Highlights for stories is something like collections for medias (the only difference it that you can only add your own stories), so it is ok to have a separate class for them.

mgp25:
@valga @aitte2 all changes done. Since `Archive` class now only contains `Highlight` related functions, wouldn't it be better to rename it to `Highlight`?

aitte2:
@valga wow ok!

@mgp25 I agree, I vote for $ig->highlight->create(), edit(), delete(), getFeed()

aitte2:
Btw under https://github.com/mgp25/Instagram-API/pull/1785/files look closely, there are still edits in collection.php. intentional or accident? and in story.php :)

mgp25:
@aitte2 was unintentional. I think everything should be okay now. Sometimes i left things behind. That's why reviews are so important.

src/Request/Collection.php
```diff
@@ -86,6 +86,9 @@ public function edit(
         $collectionId,
         array $params)
     {
+        if (($params['name'] === null) && (!is_array($params['add_media']) || (count($params['add_media']) === 0))) {
```
aitte2: .

src/Request/Collection.php
```diff
@@ -73,51 +73,39 @@ public function delete(
...
      * @return \InstagramAPI\Response\EditCollectionResponse
      */
     public function edit(
         $collectionId,
-        $name)
+        array $params)
     {
-        return $this->ig->request("collections/{$collectionId}/edit/")
-            ->addPost('name', $name)
+        if (($params['name'] === null) && (!is_array($params['add_media']) || (count($params['add_media']) === 0))) {
```
aitte2: I think this line crashes if "name" or "add_media" are not in the array. Needs isset(), right? @valga do you know?

src/Request/Collection.php
```diff
@@ -73,51 +73,39 @@ public function delete(
     }
 
     /**
-     * Edit the name of a collection.
+     * Edit the name of a collection or add more saved media to an existing collection.
      *
      * @param string $collectionId The collection ID.
-     * @param string $name         New name for the collection.
+     * @param array  $params       User-provided key-value pairs. string 'name', string[] 'add_media', string module_name (optional).
```
aitte2: There are no quotes around module_name here :)

src/Request/Highlight.php
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * Get highlight feed.
+     *
+     * @param string $userId Numerical UserPK ID.
+     *
+     * @throws \InstagramAPI\Exception\InstagramException
+     *
+     * @return \InstagramAPI\Response\ReelsMediaResponse
+     */
+    public function getUserFeed(
```
aitte2: Should this also have a getSelfUserFeed()? Like in timeline class.

src/Request/Highlight.php
```diff
@@ -0,0 +1,143 @@
...
+    {
+        if (count($mediaIds) < 1) {
+            throw new \InvalidArgumentException('You must provide at least one media ID.');
+        }
+        if ($mediaIdForCover === null) {
+            $mediaIdForCover = $mediaIds[0];
+        }
+        if ($title === null || $title === '') {
+            $title = 'Highlights';
+        } else {
```
aitte2: Could write this as elseif instead, and would get the same effect :)

src/Request/Highlight.php
```diff
@@ -0,0 +1,143 @@
...
+            ->addPost('cover', json_encode($cover))
+            ->addPost('media_ids', json_encode($mediaIds))
+            ->getResponse(new Response\HighlightFeedResponse());
+    }
+
+    /**
+     * Edit a highlight reel.
+     *
+     * @param string $highlightReelId Highlight ID, using internal format (ie "highlight:12345678901234567").
+     * @param array  $params          User-provided highlight key-value pairs. string 'title', string 'cover_media_id', string[] 'add_media', string[] remove_media.
```
aitte2: Missing quotes around remove_media.

src/Request/Highlight.php
```diff
@@ -0,0 +1,143 @@
...
+        array $params,
+        $module = 'story_viewer')
+    {
+        if ($params['cover_media_id'] === null) {
+            throw new \InvalidArgumentException('You must provide one media ID for the cover.');
+        }
+        if ($params['title'] === null || $params['title'] === '') {
+            $params['title'] = 'Highlights';
+        } else {
+            if (strlen($params['title']) > 16) {
```
aitte2: Elseif?

src/Request/Collection.php
mgp25: Fixed!

src/Request/Collection.php
mgp25: I was thinking, what about adding this to Utils?

```php
function getValue($var, $key) {
    if (!isset($var[$key])) {
        return false;
    }
    return $var[$key];
}
```
If not, should i wrap all that condition doing an `isset()` to `name` and `add_media` first?

src/Request/Highlight.php
mgp25: Yes, good idea! Im going to add it now

src/Request/Highlight.php
mgp25: You are right, didn't know why i did that xD. Fixed :)

src/Request/Highlight.php
mgp25: Fixed!

src/Request/Highlight.php
mgp25: Fixed!

src/Request/Collection.php
aitte2: Nice work! But that function is not needed since isset already checks for null too :) So you can just write this: `if (!isset($params['name']) && (!isset($params['add_media']) || !is_array($params['add_media']) || empty($params['add_media']))) {` and after that validation, you can assume that `if (isset($params['name'])) {}` is enough to know if a value was provided.

src/Request/Collection.php
```diff
@@ -73,51 +73,39 @@ public function delete(
...
 
     /**
-     * Edit the name of a collection.
+     * Edit the name of a collection or add more saved media to an existing collection.
      *
      * @param string $collectionId The collection ID.
-     * @param string $name         New name for the collection.
+     * @param array  $params       User-provided key-value pairs. string 'name', string[] 'add_media', string 'module_name' (optional).
      *
      * @throws \InstagramAPI\Exception\InstagramException
```
aitte2: Needs `@throws \InvalidArgumentException` :)

src/Request/Highlight.php
```diff
@@ -0,0 +1,151 @@
...
+
+    /**
+     * Create a highlight reel.
+     *
+     * @param array       $mediaIds        Array with one or more media IDs in Instagram's internal format (ie ["3482384834_43294"]).
+     * @param string      $title           Title for the highlight.
+     * @param null|string $mediaIdForCover One media ID in Instagram's internal format (ie "3482384834_43294").
+     * @param string      $module
+     *
+     * @throws \InstagramAPI\Exception\InstagramException
```
aitte2: Needs `@throws \InvalidArgumentException` :)

src/Request/Highlight.php
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /**
+     * Edit a highlight reel.
+     *
+     * @param string $highlightReelId Highlight ID, using internal format (ie "highlight:12345678901234567").
+     * @param array  $params          User-provided highlight key-value pairs. string 'title', string 'cover_media_id', string[] 'add_media', string[] 'remove_media'.
+     * @param string $module          (optional) From which app module (page) you're performing this action.
+     *
+     * @throws \InstagramAPI\Exception\InstagramException
```
aitte2: Needs `@throws \InvalidArgumentException` :D

src/Request/Highlight.php
```diff
@@ -0,0 +1,151 @@
...
+        return $this->ig->request("highlights/{$userId}/highlights_tray/")
+            ->getResponse(new Response\HighlightFeedResponse());
+    }
+
+    /**
+     * Get self highlight feed.
+     *
+     * @throws \InstagramAPI\Exception\InstagramException
+     *
+     * @return \InstagramAPI\Response\ReelsMediaResponse
```
aitte2: `@return \InstagramAPI\Response\HighlightFeedResponse`

src/Request/Highlight.php
```diff
@@ -0,0 +1,151 @@
...
+class Highlight extends RequestCollection
+{
+    /**
+     * Get highlight feed.
+     *
+     * @param string $userId Numerical UserPK ID.
+     *
+     * @throws \InstagramAPI\Exception\InstagramException
+     *
+     * @return \InstagramAPI\Response\ReelsMediaResponse
```
aitte2: `@return \InstagramAPI\Response\HighlightFeedResponse`

mgp25:
@aitte2 done! 😄

src/Request/Highlight.php
```diff
@@ -0,0 +1,153 @@
...
+        array $mediaIds,
+        $title = 'Highlights',
+        $mediaIdForCover = null,
+        $module = 'story_viewer')
+    {
+        if (count($mediaIds) < 1) {
+            throw new \InvalidArgumentException('You must provide at least one media ID.');
+        }
+        if ($mediaIdForCover === null) {
+            $mediaIdForCover = $mediaIds[0];
```
valga: If you want to get first element of an array, use `reset()`: `$first = reset($array);` because first element is not always at `0`.

src/Request/Highlight.php
```diff
@@ -0,0 +1,153 @@
...
+        }
+        if ($title === null || $title === '') {
+            $title = 'Highlights';
+        } elseif (strlen($title) > 16) {
```
valga: Use `mb_strlen()` instead of `strlen()` and don't forget to pass `utf-8` as its second argument. Also in other methods.

src/Request/Highlight.php
```diff
@@ -0,0 +1,153 @@
...
+    {
+        if ($params['cover_media_id'] === null) {
+            throw new \InvalidArgumentException('You must provide one media ID for the cover.');
+        }
+        if ($params['title'] === null || $params['title'] === '') {
+            $params['title'] = 'Highlights';
+        } elseif (strlen($params['title']) > 16) {
+            throw new \InvalidArgumentException('Title length must be between 1 and 16 characters.');
+        }
+        if (!isset($params['add_media']) && !is_array($params['add_media'])) {
```
valga: I think you meant `||` not `&&`, also below.

src/Request/Highlight.php
```diff
@@ -0,0 +1,153 @@
...
+     * @throws \InstagramAPI\Exception\InstagramException
+     *
+     * @return \InstagramAPI\Response\HighlightFeedResponse
+     */
+    public function edit(
+        $highlightReelId,
+        array $params,
+        $module = 'story_viewer')
+    {
+        if ($params['cover_media_id'] === null) {
```
valga: The rule is simple: if you want to check a variable against `NULL`, use `$var === null`; if you want to check array index, use `!isset($array['index'])`. Also in other places.

src/Request/Collection.php
```diff
@@ -73,51 +73,40 @@ public function delete(
...
      * @return \InstagramAPI\Response\EditCollectionResponse
      */
     public function edit(
         $collectionId,
-        $name)
+        array $params)
     {
-        return $this->ig->request("collections/{$collectionId}/edit/")
-            ->addPost('name', $name)
+        if (!isset($params['name']) && (!isset($params['add_media']) || !is_array($params['add_media']) || empty($params['add_media']))) {
```
valga: You are doing double checks here. I think it is better to rewrite this part as follows:
```php
$postData = [];
if (isset($params['name']) && $params['name'] !== '') {
    $postData['name'] = $params['name'];
}
if (isset($params['add_media'])...) {
...
}
if (!count($postData)) {
    throw new \InvalidArgumentException('You must provide a name or at least one media ID.');
}
foreach ($postData as $key => $value) {
    $request->addPost($key, $value);
}
```

valga:
@aitte2 Thanks for the help!

src/Request/Collection.php
aitte2: @valga any reason why you use `if count($arr)` instead of `if (!empty($arr))` ? The empty function is made for that checking for empty arrays :)

valga: @aitte2 Bad habit, I guess. Fixed.

mgp25: Fixed!

src/Request/Highlight.php
mgp25: True! I have changed it and took notes of it :)

src/Request/Highlight.php
mgp25: In which cases do we use `mb_strlen()` instead of the other one? Just to know it for the future. Fixed btw :)

src/Request/Highlight.php
mgp25: Fixed!

src/Request/Highlight.php
mgp25: True, that was my fault. Fixed

src/Request/Highlight.php
valga: Basically you should always use `mb_strlen()`, because almost everything runs with multibyte encodings nowadays. Use `strlen()` for binary data and ANSI strings.

src/Request/Collection.php
```diff
@@ -73,51 +73,44 @@ public function delete(
...
-        $name)
+        array $params)
     {
-        return $this->ig->request("collections/{$collectionId}/edit/")
-            ->addPost('name', $name)
+        if (isset($params['name']) && $params['name'] !== '') {
+            $postData['name'] = $params['name'];
+        }
+        if (isset($params['add_media']) && (is_array($params['add_media']) && !empty($params['add_media']))) {
+            $postData['add_media'] = $params['add_media'];
```
valga: It seems that `json_encode()` has been lost in edits.

aitte2: I also don't see any `$postData = [];` anywhere ;)

src/Request/Highlight.php
```diff
@@ -0,0 +1,153 @@
...
+     *
+     * @throws \InvalidArgumentException
+     * @throws \InstagramAPI\Exception\InstagramException
+     *
+     * @return \InstagramAPI\Response\HighlightFeedResponse
+     */
+    public function create(
+        array $mediaIds,
+        $title = 'Highlights',
+        $mediaIdForCover = null,
```
aitte2: It seems this pull request is almost done. Maybe rename this `$coverMediaId` for a shorter var name?

src/Request/Collection.php
mgp25: I'll do it later tonight. Sorry for so many changes, but i didnt had time to review full code and test it. But im learning good habits from this. 😄

aitte2: @mgp25 lol no problem xD happy to help!

src/Request/Highlight.php
mgp25: Yup. I like the idea. Changed

src/Request/Collection.php
mgp25: Fixed!

valga: @mgp25 Nope, it is not fixed. You should call `json_encode()` in this line, because your check `$key === 'add_media'` is pointless.

And one last thing: you should call `array_values()` before calling `json_encode()` on arrays, because JSON allows only arrays starting with `0` and without gaps. For example:
```
php > echo json_encode(['a', 'b', 'c']);
["a","b","c"]
php > echo json_encode([1 => 'a', 'b', 'c']);
{"1":"a","2":"b","3":"c"}
php > echo json_encode(['a', 2 => 'b', 'c']);
{"0":"a","2":"b","3":"c"}
```
As you can see, only first array is encoded into JSON array, other two are objects. As alternative, you can do it with `implode()`: `'["'.implode('","', $array).'"]'`

src/Request/Collection.php
```diff
@@ -73,51 +73,45 @@ public function delete(
...
-        $name)
+        array $params)
     {
-        return $this->ig->request("collections/{$collectionId}/edit/")
-            ->addPost('name', $name)
+        $postData = [];
+        if (isset($params['name']) && $params['name'] !== '') {
+            $postData['name'] = $params['name'];
+        }
+        if (isset($params['add_media']) && (is_array($params['add_media']) && !empty($params['add_media']))) {
```
aitte2: The parenthesis around the 2nd part is pointless since every condition is && :)

src/Request/Collection.php
```diff
@@ -73,51 +73,45 @@ public function delete(
...
+            $postData['add_media'] = json_encode(array_values($params['add_media']));
+            if (isset($params['module_name']) && $params['module_name'] !== '') {
+                $postData['module_name'] = $params['module_name'];
+            } else {
+                $postData['module_name'] = 'feed_saved_add_to_collection';
+            }
+        }
+        if (!count($postData)) {
```
aitte2: Let's use If empty($postData)

src/Request/Highlight.php
```diff
@@ -0,0 +1,153 @@
...
+     *
+     * @return \InstagramAPI\Response\HighlightFeedResponse
+     */
+    public function create(
+        array $mediaIds,
+        $title = 'Highlights',
+        $coverMediaId = null,
+        $module = 'story_viewer')
+    {
+        if (count($mediaIds) < 1) {
```
aitte2: In this place we can use if empty() instead

aitte2:
This one is gonna need a squash merge ^^

mgp25:
@aitte2 fur sure xD this took longer than i thought, but it was my fault not looking step by step the code for pausible issues, typos and so. I have commited the latest changes you requestes.

aitte2:
@mgp25 lol yep :P speaking about reading carefully, look here xD

https://github.com/mgp25/Instagram-API/pull/1785/commits/a05c216e71c9b5a2c1428c7c0e14820595ad27c8

When you took out the extra `()` it accidentally removed part of `is_array(` xD hehe

When I said extra parenthesis I meant to change:
`if (isset($params['add_media']) && (is_array($params['add_media']) && !empty($params['add_media']))) {`
into:
`if (isset($params['add_media']) && is_array($params['add_media']) && !empty($params['add_media'])) {`

after you do that final fix, this looks done to me. what do you think @valga?

mgp25:
@aitte2 you are free to squash and merge if @valga agrees

valga:
@mgp25 You have 3 `json_encode()` calls without `array_values()`. We are good to go after fixing that. And be careful with `$cover`, because it must be an object.

aitte2:
@valga You mean adding json_encode(array_values()) to every json encoding place except the $cover objects?

mgp25:
Commit done, i found other calls where i added `array_values()` +  a fix in the doc. It should be okay now.

Waiting for your approvals

aitte2:
@mgp25 good job! I read all the final lines of changed code and it looks great now. :) Do you agree, @valga?

valga:
@mgp25, @aitte2 I already gave my approval. Just don't forget to rename and squash the PR.

aitte2:
Ok! Finally done, great work! 👍  :)

PR #1831, opened 2018-02-28
Generate video thumbnail from specific time

tje3d:
Title says all ( not tested )

```php
$filePath  = "video.mp4";
$caption   = "My Awesome Caption!";

$ig    = new \InstagramAPI\Instagram();
$video = new \InstagramAPI\Media\Video\InstagramVideo($filePath, [
    'thumbnailTimestamp' => '00:00:05'
]);
$ig->timeline->uploadVideo($video, compact('caption'));
```


aitte2:
Thanks for starting on this! Weird though that it can only get thumbnail from 00:00:00 to 00:00:59, it does not seem safe or useful this way.
 I think i remember seeing a chat in some todo list around here about taking thumbnails from a percent location, such as 50%, auto calculated for any video. That's a more safe way than this risky idea of trying and failing to say "get thumbnail at 50 seconds, in this 10 second long video". Maybe both methods should be allowed. If int is used as param, use seconds (without this 59 second limit in this one), if string assume the format is X% and use video info duration (it is already stored internally somewhere since the project can look at and block too long videos) and calculate offset automaticallly? But take into account the video playback start pull Request, which says video duration is nor always correct.....

src/Media/InstagramMedia.php
```diff
@@ -1323,4 +1326,11 @@ protected function _debugDimensions(
             $width, $height, $width / $height
         );
     }
+
+    public function thumbnailFrom($second = 1)
```
valga: There is no need for a separate method, you should use `$options` argument in `InstagramThumbnail` constructor.

src/Media/InstagramMedia.php
```diff
@@ -92,6 +92,9 @@
      * input being outside allowed min/max range. */
     protected $_forceTargetAspectRatio;
 
+    /** @var int From where make the thumbnail */
+    protected $_thumbnail = 1;
```
valga: This property should be moved into `InstagramThumbnail` class, because it doesn't make sense for `InstagramMedia` class. Also, please choose more specific name, like `thumbnailTimestamp`.

src/Media/Video/InstagramThumbnail.php
```diff
@@ -17,7 +17,9 @@ protected function _shouldProcess()
     /** {@inheritdoc} */
     protected function _getOutputFormat()
     {
-        // TODO Allow custom timestamp.
-        return '-f mjpeg -ss 00:00:01 -vframes 1';
+        $second = $this->_thumbnail < 0 ? 0 : ($this->_thumbnail > 59 ? 59 : $this->_thumbnail);
```
valga: This check should be moved into constructor where we have `VideoDetails` object with video duration, so you can compare thumbnail timestamp with actual video length.

src/Media/Video/InstagramThumbnail.php
```diff
@@ -17,7 +17,9 @@ protected function _shouldProcess()
...
+        $second = str_pad($second, 2, 0, STR_PAD_LEFT);
+
+        return '-f mjpeg -ss 00:00:'.$second.' -vframes 1';
```
valga: Just use `sprintf('-f mjpeg -ss 00:%02d:%02d -vframes 1', $timestamp / 60, $timestamp % 60)` here. I think it is safe to skip hours here.

aitte2:
@valga GREAT FEEDBACK! :)

mgp25:
@tje3d if you do: `composer codestlye` it fill fix the style automatically.

src/Media/Video/InstagramVideo.php
```diff
@@ -43,6 +46,12 @@ public function __construct(
         }
 
         $this->_details = new VideoDetails($this->_inputFile);
+
+        $maximumVideoTime = $this->_details->getDuration() > $this->_constraints->getMaxDuration() ? $this->_constraints->getMaxDuration() : $this->_details->getDuration(); // Result is a Float number like 230.138000
```
tje3d: $maximumVideoTime = min($this->_details->getDuration(), $this->_constraints->getMaxDuration()); // Result is a Float number like 230.138000

src/Media/Video/InstagramVideo.php
```diff
@@ -43,6 +46,12 @@ public function __construct(
...
+        $maximumVideoTime = strtotime(sprintf('00:%02d:%02d', $maximumVideoTime / 60 % 60, $maximumVideoTime % 60));
```
aitte2: I really dislike these usages of `strtotime` which is turning a timestamp such as `strtotime('00:01:01')` to a "unix timestamp today" such as `1521244861`.

`$maximumVideoTime` is already a float such as `806.3`, I think?

Keep it like that and make `thumbnailTimestamp` into a number (int) too. I added a function for you to do that: https://github.com/mgp25/Instagram-API/commit/4032a5c015d164e8955c9827f809fc07e2e411ee

Do:

`$thumbnailTimestamp = Utils::hmsTimeToSeconds(isset($options['thumbnailTimestamp']) ? $options['thumbnailTimestamp'] : '00:00:01');`

:)

src/Request/Direct.php
```diff
@@ -715,13 +716,18 @@ function () use ($internalMetadata, $recipients, $options) {
      */
     public function sendDisappearingVideo(
         array $recipients,
-        $videoFilename,
```
aitte2: This change is wrong. Please revert `Request/Direct.php`.

In fact I just noticed this whole pull request is doing something nasty now and is massively changing the functions without documenting that... Every function that takes `@param string $videoFilename` is now expecting a Video object, and also has a ton of "backward compatibility converters". It looks dirty and needs more work...

@valga Shouldn't video timestamp be put in `$externalMetadata` and handled by the internal uploader instead of by the user? That way the user can continue sending string filenames. And I don't have time to review this anymore for a while, mind taking over?

davisjackie321:
Yo

valga:
@tje3d Somehow this PR went wrong, please revert your changes to `$videoFilename` params.

When speaking about `InstagramThumbnail` class and its constructor I meant something like this:
```php
class InstagramThumbnail extends InstagramVideo {
...
    protected $_thumnailTimestamp;
...
    public function __construct(
        $inputFile,
        array $options = [],
        FFmpegWrapper $ffmpegWrapper = null)
    {
        parent::__construct($inputFile, $options, $ffmpegWrapper);
        
        // We have video details in $this->_details now, so we can check timestamp against video duration.
        if (isset($options['thumbnail_timestamp']) {
            $this->_thumnailTimestamp = (int) max(0, min($options['thumbnail_timestamp'], $this->_details->getDuration()));
        } else {
            $this->_thumnailTimestamp = 0;
        }
    }
...
```

valga:
@aitte2 Well, it is complicated. Direct videos don't require thumbnail at all, (direct) stories always have thumbnail at 00:00, so basically thumbnail timestamp is needed just for the feed videos (not sure about album ones).

Moreover, there is an experiment to move thumbnail extraction to the server. Just send `extract_cover_frame=1` and `cover_frame_seek=time_in_seconds_float` params and their server will make a thumbnail for you.

aitte2:
@valga Hmm okay that is complicated indeed.

Does this sound right?

1. It should only be supported in `timeline->uploadVideo()` (which already has `$externalMetadata`).
2. Change `internal->uploadSingleVideo()` to pass its `$externalMetadata` to `uploadVideoThumbnail()`.
3. Change `uploadVideoThumbnail()` to add the `$externalMetadata` parameter.
4. Change `uploadVideoThumbnail()` to say "if targetFeed === timeline", then look for timestamp in externalmetadata, otherwise always use `00:00` for all other target feeds.
5. Timestamp parsing: If is_int/is_float, use number as-is, otherwise use https://github.com/mgp25/Instagram-API/commit/4032a5c015d164e8955c9827f809fc07e2e411ee function to convert it.
6. Then, cap it so it cannot be longer than duration of video.

How does this plan sound?

valga:
@aitte2 2-4 aren't ok. I am going to add those experiments some day, so we must handle thumbnail timestamp in `uploadSingleVideo()` because it's going either to `uploadVideoThumbnail()` (local thumbnail extraction) or to `configureSingleVideo()` (remote thumbnail extraction).

aitte2:
@valga I see. Perhaps that is a good reason to put it in `InstagramThumbnail`? That class knows the length of the video. We can make that class validate the timestamp and have 3 things: A getter to get the thumbnail as an image `getThumbnailImage()`, a getter to get it as a float `getTimestampAsFloat()`, and a getter to get it as `00:00:01.150` (`getTimestampAsString()`) timestamp.

To configure a video to make the server make the image we would do:

```php
$thumbnail = new InstagramThumbnail()...;
configure(['cover_frame_seek' => $thumbnail->getTimestampAsFloat()]);
```

src/Media/Video/InstagramThumbnail.php
```diff
@@ -7,6 +7,18 @@
...
+    /** @var InstagramVideo The video. */
+    protected $video;
+
+    public function __construct(
+        InstagramVideo $video,
+        array $options = [])
+    {
+        $this->video = $video;
+
+        parent::__construct($video->getFile(), $options);
```
aitte2: Just saw this, which is very bad (triggers video conversion).

Either way it doesn't matter anymore, I am taking the working parts of this pull request + valga's feedback and am nearly done with a fixed version of this idea. I will credit you and valga for your parts in this :)

src/Request/Internal.php
```diff
@@ -389,7 +395,7 @@ public function uploadVideo(
         } catch (\Exception $e) {
             // Wrap runtime errors.
             throw new UploadFailedException(
-                sprintf('Upload of "%s" failed: %s', basename($videoFilename), $e->getMessage()),
+                sprintf('Upload of "%s" failed: %s', basename($video->getFile()), $e->getMessage()),
```
aitte2: Again this thing which triggers video conversion ;) be careful of `getFile()`! ;p

tje3d:
I don't have enough knowledge about the Core, so it does not seem this PR to be very helpful.
Closing...

aitte2:
@tje3d I think you did a great job getting everything started. I am confused by the internals too. :)

Anyway the job is done now after 3 hours in front of the tv with some casual coding. :)

@valga https://github.com/mgp25/Instagram-API/commit/96b02ec73256c865cfc5930d8b629d8dc7762bda

All validation of target feed and timestamp position is in the thumbnail class, so you only need to create a class instance and then `getTimestamp()` to know what to send to the server for cover generation later. :)

PR #1854, opened 2018-03-09
Internal: add missing quotes to defined() call

crewmango:
it only accepts a string (the name of the constant) as parameter


valga:
Thank you for noticing and fixing that.

PR #1893, opened 2018-04-03
Media: Refactor ffmpeg wrapper

valga:


src/Media/Video/FFmpeg.php
```diff
@@ -0,0 +1,232 @@
...
+        if ($ffmpegBinary === null) {
+            return static::_autoDetectBinary();
+        }
+
+        if (isset(self::$_instances[$ffmpegBinary])) {
+            return self::$_instances[$ffmpegBinary];
+        }
+
+        $instance = new static($ffmpegBinary);
+        self::$_instances = $instance;
```
aitte2: Bug? I expected `self::$_instances[$ffmpegBinary] = $instance`?

examples/liveBroadcast.php
```diff
@@ -39,8 +39,9 @@
     // https://github.com/mgp25/Instagram-API/issues/1488#issuecomment-333365636
 
     // Get FFmpeg handler and ensure that the application exists on this system.
-    // NOTE: Look in `\InstagramAPI\Utils` if you want a custom ffmpeg path.
-    $ffmpeg = \InstagramAPI\Utils::getFFmpegWrapper();
+    // NOTE: You can supply custom path to the ffmpeg binary.
```
aitte2: Maybe say `, or use NULL to autodetect.`

src/Media/Video/FFmpeg.php
```diff
@@ -0,0 +1,232 @@
...
+     */
+    protected function __construct(
+        $ffmpegBinary)
+    {
+        $this->_ffmpegBinary = $ffmpegBinary;
+
+        try {
+            $this->version();
+        } catch (\Exception $e) {
+            throw new \RuntimeException(sprintf('It seems that the path to ffmpeg binary is invalid. Please double verify it.'));
```
aitte2: "Please double verify it" is a bit strange. Something like "Please check your path to ensure that it is correct."

src/Media/Video/FFmpeg.php
```diff
@@ -0,0 +1,232 @@
...
+        }
+    }
+
+    /**
+     * Create a new instance or use a cached one.
+     *
+     * @param string|null $ffmpegBinary
```
aitte2: I would add param description: `Path to a ffmpeg binary, or NULL to autodetect.`

src/Media/Video/FFmpeg.php
```diff
@@ -0,0 +1,232 @@
...
+        $process = $this->runAsync($command);
+
+        try {
+            $exitCode = $process->wait();
+        } catch (\Exception $e) {
+            throw new \RuntimeException(sprintf('Failed to run the ffmpeg binary: %s', $e->getMessage()));
+        }
+        if ($exitCode) {
+            $errors = str_replace("\n", '"], ["', $process->getErrorOutput());
+            $errorMsg = sprintf('FFmpeg Errors: ["%s"], Command: "%s".', $errors, $command);
```
aitte2: Maybe this is an issue on windows, if ffmpeg outputs \r\n which leaves \r which would move cursor to the 1st dos column or do other weirdness.

I would use something like: `preg_replace('/[\r\n]+/', ...)` to totally replace _all_sequences of newline characters with the separator.

src/Media/Video/FFmpeg.php
```diff
@@ -0,0 +1,232 @@
...
+
+            throw new \RuntimeException($errorMsg, $exitCode);
+        }
+
+        return explode("\n", $process->getOutput());
```
aitte2: Same issue here with newline handling. Use preg_split and the pattern above. Maybe with PREG_SPLIT_NO_EMPTY.

src/Media/Video/FFmpeg.php
```diff
@@ -0,0 +1,232 @@
...
+     *
+     * @return string
+     */
+    public function version()
+    {
+        return $this->run('-version')[0];
+    }
+
+    /**
+     * Get a path to the ffmpeg binary.
```
aitte2: "Get a" is weird since this is the current instance's path. So "Get the" sounds more correct.

valga:
@aitte2 Thanks for the feedback.

I think we need something similar for ffprobe.

aitte2:
No problem. All code looks correct now. I haven't run this code but I assume you have and that it works. :)

About ffprobe: I agree. But how? As a shared base class perhaps? Lots of the code here looks like it can be reused as a `BinaryWrapper` class, with `class FFprobe extends BinaryWrapper`, and with `$this->_binary` and `->getBinary()` as a shared method. And for error handling, the shared `run()` can call some `(array) $errors = $this->_formatErrors(preg_split(trim(getErrorOutput())))` which is an abstract method that takes an array of error lines, and every wrapper has to implement an error formatter. These are just some ideas...

valga:
@aitte2 Let's merge this PR as is, so mgp can work on live issue. Will think about additional refactoring later.

aitte2:
@valga I agree. Merge it :) 👍

PR #1895, opened 2018-04-04
Internal: check for audio stream

alexok:
ffmpeg wrapper throws an exception when loading video without audio stream


aitte2:
@alexok Please describe why this change is necessary? I can guess why, and it seems fine if it does what I suspect, but please tell us. ;) 

And merging will need to wait until @valga has finished the ffmpeg wrapper rewrite (#1893).

alexok:
@aitte2 I updated the description

aitte2:
This change means no audio segment files are created, which is logical since no audio exists. So disabling splitting to audio makes sense.

But I just wonder if the later encoder/uploader code will look for those missing audio files and is broken by this change?

alexok:
Ready

aitte2:
@alexok Thanks a lot alex! 👍 :)

Question: Since you obviously tested this change, did the videos (without audio) upload correctly to Instagram after this change?

alexok:
Yes, video without audio is loaded normally

aitte2:
Thanks, perfect news :)

PR #1898, opened 2018-04-06
Examples: Fix liveBroadcast example

aitte2:
- Some places said `$lastCommentTS` and others said `$lastCommentTs`.

- Cleaned up "last comment timestamp" algorithm by making it check if an
  array element exists (via `&& end($array)`) before trying to use the
  value, and also changing ternary to a normal if-statement.

- Change the `while()` loop to a `do{} while()` so that the code is
  always guaranteed to run at least once.


aitte2:
@mgp25 @valga does this look ok?

mgp25:
@aitte2 sorry for the typo 😉 

Yep, it looks great.

Just one thing, just to understand and improve my coding for future times. Why are you checking `&& end($array)`? I mean, if `$array` is `null` you wont go into the if condition and if its different from `null`,  it will go into the if and get last item in the array using `end()`

aitte2:
@mgp25 Ok good! :) And I wasn't sure about needing `end()`. I did it because: if the array is empty, `end()` returns false (instead of object), which means the old code would break if the response was:

```json
"system_comments":[],
"comments":[
   {comment...}
]
```

(Or if it was the opposite: there's system comment but not regular comment).

The old code assumed that there's at least 1 entry in each array and just tries to read it directly: `end($systemComments)->getCreatedAt();` which would be the same as trying to write `false->getCreatedAt()`, since `end()` returns false if there are no array entries.

Hmm actually, I think I am wrong :D I just remembered that `[]` counts as false. So if the value is `null` or `[]` it wouldn't do anything in the old code either.

I'll edit it to remove the extra `end()` checks.

aitte2:
Ok fixed. But instead of just `if ($comments)` I changed it to `if (!empty($comments))`. This was just to make the code more readable for newbies. The effect is the same.

valga:
I have a couple of questions.

1. What happens when the video file is missing?
2. Does the app start sending comments/likes request right after going live, or there is some delay between broadcast start and a first request?

aitte2:
@valga 

I don't know about the 2nd question. As for 1st: Good question. FFmpeg will exit with non-0 code if input video is missing or fails in any other way. So it depends on what symfony `Process` does when a program has an error code when you try to `start()` it. I have not read their docs so I don't know.

https://github.com/mgp25/Instagram-API/blob/5838b24cf3a94ef4af2d56ddc8ee276e04ec414d/src/Media/Video/FFmpeg.php#L109-L118

PR #1920, opened 2018-04-17
Update ISSUE_TEMPLATE.md

zzet:
@mgp25 just an idea


SteveJobzniak:
Hi. Making it clear not to ask about those things is a good idea. But 3 checkboxes is too much *and* it's too far down in the text which means the lazy people who post those tickets won't even read it. 😆 It needs to be near the top. I'll write something in a moment...

PR #1924, opened 2018-04-17
Share media (sendPost) for Realtime

Maxim-Mazurok:
Added option to share media using Realtime.
The `realtimeHttp.php` can be extended with:
```php
case '/post':
    return $this->_handleClientContext($this->_rtc->sendPostToDirect($params['threadId'],
                                       $params['mediaId'], 
                                       ['media_type' => 'photo']));
```


src/Realtime/Command/Direct/SendPost.php
```diff
@@ -0,0 +1,35 @@
...
+        $threadId,
+        $mediaId,
+        array $options = [])
+    {
+        parent::__construct($threadId, self::TYPE, $options);
+
+        if (!preg_match('#^\d+_\d+$#D', $mediaId)) {
+            throw new \InvalidArgumentException(sprintf('"%s" is not a valid media ID.', $mediaId));
+        }
+        if (!isset($options['media_type'])) {
```
valga: `media_type` is not used in realtime, so you should remove its validation.

src/Realtime.php
```diff
@@ -196,6 +196,35 @@ public function sendTextToDirect(
...
 
+    /**
+     * Share an existing media post to a given direct thread.
+     *
+     * @param string $threadId Thread ID.
+     * @param string $mediaId  The media ID in Instagram's internal format (ie "3482384834_43294").
+     * @param array  $options  An associative array of additional parameters, including:
+     *                         "media_type" (required) - either "photo" or "video";
+     *                         "client_context" (optional) - predefined UUID used to prevent double-posting;
+     *                         "text" (optional) - text message.
```
valga: Why `text` param is not used anywhere?

src/Realtime.php
```diff
@@ -196,6 +196,35 @@ public function sendTextToDirect(
...
+     *
+     * @return bool|string Client context or false if sending is unavailable.
+     */
+    public function sendPostToDirect(
+        $threadId,
+        $mediaId,
+        array $options = [])
+    {
+        try {
```
valga: An ability to send a post via realtime is controlled by `ig_android_rtc_reshare/is_rtc_reshare_enabled` experiment, so you should check whether it is enabled before sending anything.

And as far as I know, it is not enabled yet.

valga:
@Maxim-Mazurok Thank you for your contribution.

It would be nice if you could update `examples/realtimeHttp.php` with your example.

valga:
@Maxim-Mazurok I decided to implement realtime sharing by myself. See ea27d0141319ded06f1bda51424b1550aae52dc3, but take in mind that this feature is still experimental.

PR #2007, opened 2018-06-03
Zero rating rewrites implementation

valga:
It has been simplified: for example, the app stores multiple tokens (per carrier) and switches between them when you change your SIM card. As we don't have a mobile network, it is ok to keep just one token (that is always empty anyway).

@mgp25 Do not squash this PR: it meant to be in 2 individual commits. Do a rebase.


aitte2:
Wow looks like a big job. Impressive! :)

I only have one feedback: Why are getters `$ig->client->zeroRewrite()->update()` instead of `$ig->client->getZeroRewrite()->update()`? The word `get` is missing from your new getters... which makes them look like "action" functions rather than getters.

Since there's no "get" word I read `$ig->client->zeroRewrite()` as "trigger a client zero-rewrite call"... 😮

mgp25:
@valga let me know when this is ready to merge.

Regards

valga:
@aitte2 I renamed it to `zeroRating()` to cause less confusion. There is no point in `get` prefix when we have no setter.

@mgp25 It's almost ready.

valga:
@mgp25 It's ready.

PR #2040, opened 2018-06-17
Share media to story feed

mgp25:
This commits add the feature to share media to your own story feed.

This PR is not finished and has some issues and probably needs to be factorized.

The main issue is:

- How to manage temp files (the implementation in this PR is not okay) since the function needs to download the media and reupload it.
- Should we add a fixed external metadata or should we add an argument to the function to override some values of the metadata.


src/Request/Story.php
```diff
@@ -55,6 +56,65 @@ public function uploadVideo(
...
+     * @throws \RuntimeException
+     * @throws \InstagramAPI\Exception\InstagramException
+     * @throws \InstagramAPI\Exception\UploadFailedException If the video upload fails.
+     *
+     * @return \InstagramAPI\Response\ConfigureResponse
+     *
+     * @see Internal::configureSinglePhoto() for available metadata fields.
+     * @see Internal::configureSingleVideo() for available metadata fields.
+     */
+    public function shareMedia(
```
valga: I don't think that we need a new method for that. You should convert it into 2 examples: one for photo and one for video.

src/Request/Story.php
```diff
@@ -55,6 +56,65 @@ public function uploadVideo(
...
+        switch ($mediaInfo->getMediaType()) {
+            case '1': // PHOTO
+                $mediaUrl = $mediaInfo->getImageVersions2()->getCandidates()[0]->getUrl();
+                $outputFile = Utils::createTempFile(sys_get_temp_dir(), 'IMG');
+                file_put_contents($outputFile, file_get_contents($mediaUrl));
+                break;
+            case '2': // VIDEO
+                $mediaUrl = $mediaInfo->getVideoVersions()[0]->getUrl();
+                $outputFile = Utils::createTempFile(sys_get_temp_dir(), 'VID');
+                file_put_contents($outputFile, file_get_contents($mediaUrl));
```
valga: While it is ok to use `file_get_contents()` to download a photo, it is unacceptable for a video, because it places the whole file into memory. cURL (and Guzzle) can stream response data into a file. So you should use our `$ig->request()` with `sink` custom Guzzle option: http://guzzle.readthedocs.io/en/latest/request-options.html#sink-option

src/Utils.php
```diff
@@ -690,6 +690,31 @@ public static function throwIfInvalidStoryHashtags(
...
+        $missingKeys = array_keys(array_diff_key(['media_id' => 1], $attachedMedia));
+        if (count($missingKeys)) {
+            throw new \InvalidArgumentException(sprintf('Missing keys "%s" for mention array.', implode(', ', $missingKeys)));
+        }
+
+        if (!is_string($attachedMedia['media_id']) && !is_numeric($attachedMedia['media_id'])) {
+            throw new \InvalidArgumentException(sprintf('Invalid value "%s" for media_id.', $attachedMedia['media_id']));
+        }
+
+        self::_throwIfInvalidStoryStickerPlacement(array_diff_key($attachedMedia, array_flip($requiredKeys)), 'attahced media');
```
valga: Typo (attahced).

src/Request/Internal.php
```diff
@@ -306,6 +309,11 @@ public function configureSinglePhoto(
                         ->addPost('internal_features', 'polling_sticker')
                         ->addPost('mas_opt_in', 'NOT_PROMPTED');
                 }
+                if ($attachedMedia !== null) {
+                    Utils::throwIfInvalidAttachedMedia($attachedMedia);
+                    $request
```
valga: I think it is better to place it on one line, same below.

src/Request/Internal.php
```diff
@@ -306,6 +309,11 @@ public function configureSinglePhoto(
...
+                        ->addPost('attached_media', json_encode($attachedMedia));
```
valga: I saw `story_sticker_ids` param in your email. It contains shared media ID, so I think we should add it.

src/Request/Story.php
```diff
@@ -55,6 +56,65 @@ public function uploadVideo(
...
+                'y'             => 0.5004223,
+                'z'             => 0,
+                'width'         => 0.8,
+                'height'        => 0.6224662,
+                'rotation'      => 0.0,
+                'media_id'      => $mediaInfo->getPk(),
+                'is_sticker'    => true,
+            ];
+
+        if ($mediaInfo->getMediaType() === 1) {
```
valga: Is it possible to share an album?

src/Request/Story.php
mgp25: Oks, I'm going to remove this and add examples instead.

src/Utils.php
mgp25: Fixed.

src/Request/Story.php
mgp25: Yes and no. 

Yes because it lets you share an item of the album, in essence, we are treating the items as if they were videos or images.

No because you can't share a whole album.

src/Request/Internal.php
mgp25: Added :)

mgp25:
@valga How about that? https://github.com/mgp25/Instagram-API/pull/2040/commits/90fadb6cc61fd3a13f3aaef66014f9d1bd23cc09

examples/sharePhotoToStoryFeed.php
```diff
@@ -0,0 +1,75 @@
...
+    exit(0);
+}
+
+// NOTE: This code will make the credits of the media area 'clickable', but YOU need to
+// manually draw the credit to the user or a sticker-image on top of your image yourself
+// before uploading, if you want the credit to actually be visible on-screen!
+
+// If we want to attach a media, we must find a valid media_id first.
+try {
+    $mediaInfo = $ig->media->getInfo($mediaId)->getItems()[0];
```
valga: `Media::getInfo()` docs says that the media ID is required in `3482384834_43294` format. Does it work with the short one?

examples/sharePhotoToStoryFeed.php
```diff
@@ -0,0 +1,75 @@
...
+    // This example will upload the image via our automatic photo processing
+    // class. It will ensure that the story file matches the ~9:16 (portrait)
+    // aspect ratio needed by Instagram stories. You have nothing to worry
+    // about, since the class uses temporary files if the input needs
+    // processing, and it never overwrites your original file.
+    //
+    // Also note that it has lots of options, so read its class documentation!
+    $photo = new \InstagramAPI\Media\Photo\InstagramPhoto($outputFile, ['targetFeed' => \InstagramAPI\Constants::FEED_STORY]);
+    $ig->story->uploadPhoto($photo->getFile(), $metadata);
+    unlink($outputFile);
```
valga: It is better to put `unlink()` into a `finally` block, as it guarantees that the tmp file will be removed even in case of an error:
```php
$outputFile = Utils::createTempFile(sys_get_temp_dir(), 'IMG');
try {
    ...
} catch (\Exception $e) {
    ...
} finally {
    @unlink($outputFile);
}
```

examples/sharePhotoToStoryFeed.php
```diff
@@ -0,0 +1,75 @@
...
+    // NOTE: Providing metadata for story uploads is OPTIONAL. If you just want
```
valga: Providing metadata is required in this case, so you should remove this note.

examples/sharePhotoToStoryFeed.php
```diff
@@ -0,0 +1,75 @@
...
+    ],
+];
+
+try {
+    $client = new GuzzleHttp\Client();
+    $outputFile = Utils::createTempFile(sys_get_temp_dir(), 'IMG');
+    $response = $client->request('GET', $mediaInfo->getImageVersions2()->getCandidates()[0]->getUrl(), ['sink' => $outputFile]);
+    // This example will upload the image via our automatic photo processing
+    // class. It will ensure that the story file matches the ~9:16 (portrait)
+    // aspect ratio needed by Instagram stories. You have nothing to worry
```
valga: I think that everything starting from "You have nothing to worry" should be removed from this comment, because we have no original file to keep.

examples/sharePhotoToStoryFeed.php
mgp25: In fact, the correct way to do it is in that format `pk_userid`, but using media pk is a valid option. Yes, it works.

PR #2041, opened 2018-06-18
Cutout duplicate. $this->_setUserWithoutPassword($username); call alr…

4n70w4:
`$this->_setUserWithoutPassword($username);` call already exists inside `$this->userLookup($username);`


4n70w4:
@valga I did not expect that the githab automatically adds other commits from my branch to the pullrequest. If critical - I can remake.

I emulate the work of different versions of the application, in different versions different requests.

PR #2042, opened 2018-06-19
Refactoring: make getters for constants.

4n70w4:
Never call a Constants class directly.
Use a wrapper to access constants.
It is necessary for dynamic constants overwrite.


valga:
@4n70w4 Although our `Constants` class is disgusting, this PR makes it even worse:
1. Most important thing: you can't change a version without reversing the app, because they change a lot of things between versions. New endpoints, new or updated params, deprecated endpoints, etc.
2. Some of these consts are really consts, so they can't be overridden. For example, there is no way for `Content-Type` header to change without changing request body encoding.
3. It is dangerous to override some consts. For example, our exception mapper uses messages to detect fatal errors. If you change user locale, you could miss some important errors.
4. Absence of interface implies that the override can be done only with inheritance. It's a bad practice.
5. Your constants class is too wide. For example, there is no need for `Fbns` to know anything about signatures and HTTP requests.

You should use https://github.com/vaimo/composer-patches if you really want to override something.

4n70w4:
@valga I have a need to redefine some groups of constants in runtime. This is a good way for dependency injection. Use static methods, variables and const - makes dynamic override difficult. Let's think of another way to change constants in runtime?

mgp25:
@4n70w4 I don't think changing constants in runtime is a good idea (it depends on many things, and above everything we need to make sure the user is "safe"). Although it would be nice to have something for example to change between platforms (iOS/Android). This is not something im planning to add soon, but we'll (maybe) in the future.

What are you thoughts about this? @valga

4n70w4:
@mgp25 I work with several users in parallel in one installation. Each user needs different client settings. I already do it in my version. And it works. But when you make changes in lib, I sit for a long time and merge conflicts.

mgp25:
@4n70w4 It is okay if you want to apply that changes to your code, but for this repo it will bring a lot of issues because of different versions. Each version isn't just constants, there are more things and that takes a lot of time to track every single change for every single version. 

I personally think we should stick the latest Instagram stable release (unique version), but I am open to hear different opinions. @valga @aitte2

4n70w4:
@mgp25 by default all users will get your stable constants and queries.
For fear and risk, developer can override the constants and queries.
Everyone is happy.

It is necessary to give possible overwrite of constants and behavior.
This is a modern and common approach. Dependency injection, etc.
I have another commits to simplify overriding requests classes.

P.S> Current version of android application 49.0.0.15.89, but in this lib - 42.0.0.19.95

valga:
@4n70w4 Correct me if I am wrong. You want to use the most recent version of the app, but we always are a month or two behind it, so you want to override our version/keys/requests. If yes, I think I have a better solution: just open a PR called "Update to _version_" or something like that and put everything you need there. Then you can use your fork in your `composer.json`.

We will update and merge it when necessary. That way:
1) you don't have to override anything;
2) double work (yours and ours) on requests update is eliminated.

@mgp25 I don't think the we need an ability to override everything, because it leads to fragmentation and useless work.

4n70w4:
@valga I work with several users in parallel in one installation. Each user needs different client settings. Different versions give different advantages. I need to use several versions at the same time. Hard-coded constants and requests are not suitable in my case.

4n70w4:
All these improvements will make your library code more flexible and progressive.
See: https://en.wikipedia.org/wiki/Dependency_injection
https://en.wikipedia.org/wiki/SOLID
https://en.wikipedia.org/wiki/GRASP_(object-oriented_design)

mgp25:
@4n70w4 I am not arguing that the system you are proposing is bad, but for this situation I don't think it is the best solution. Each version contains several changes and it will require support for all of them,  and that can lead to particular issues of every version.

As @valga said, is more convenient to update everything to a newer version with all its changes rather than changing only contains which can end in different errors.

valga:
@4n70w4 We started to work on IGTV, so I am closing this, because there is a better way to have multiple apps.

PR #2044, opened 2018-06-19
Add ability to set $responseObject as property of Request class and u…

4n70w4:
Add ability to set `$responseObject` as property of Request class and use it value if `getResponse()` method receive empty `$responseObject`.

This is necessary to dynamically redefine the `$responseObject` for queries.


valga:
@4n70w4 Please provide a valid user case for this PR.

4n70w4:
@valga ok.

// return Request instance instead Response instance
// inside getSomething() calls set setResponseObject() with needed $responseObject
$request = $ig->anothers->getSomething();

// dynamically update request
$request->addPost('something', $value);

// execute request without passing $responseObject
$result = $request->getResponse();

valga:
@4n70w4 I see no difference between
```php
$responseObject = new SomeResponse();
$response = $request->getResponse($responseObject);
```
and
```php
$responseObject = new SomeResponse();
$request->setResponseObject($responseObject);
$response = $request->getResponse();
```

PR #2047, opened 2018-06-20
Add ConsentRequiredException

4n70w4:


valga:
@4n70w4 And how to express a consent? Catching this exception is useless if you can't do anything with it.

4n70w4:
Firstly, you can catch concrete exception and show custom message.
The first situation is easier and more logical to solve with editing the library core. right?

Secondly, you can express a consent in client code.
For second case, there is no need for editing in the core of library.

4n70w4:
@valga Also similar exceptions FeedbackRequiredException, ChallengeRequiredException and CheckpointRequiredException are not processed in the library. They to is useless if you can't do anything with it.

valga:
@4n70w4 That's a different situation: feedbacks, challenges and checkpoints are usually related with spam. As we don't support spamers, we don't have and we don't accept a code related with their solving. But consent is just GDPR in action: anyone could face it.

Another thing: it should be added into our uploaders to break the loops.

4n70w4:
@valga I can add requests to express a consent.
But library-user must call this requests from his code.

mgp25:
Although I agree with adding GPDR policy consent, I think it would be better if the user accepts it manually (by checking his device), so the user can read everything and accept it if he wants.

GPDR consent request consists in three step requests like this one:

```php
    public function sample()
    {
        return $this->ig->request('consent/existing_user_flow/')
            ->addPost('_uuid', $this->ig->uuid)
            ->addPost('_uid', $this->ig->account_id)
            ->addPost('_csrftoken', $this->ig->client->getToken())
            ->addPost('current_screen_key', 'tos_and_two_age_button')
            ->addPost('updates', json_encode(['age_consent_state' => 2, 'tos_data_policy_consent_state' => 2]))
            ->getResponse(new Response\GenericResponse());
    }
```

sajad-sst:
@mgp25 
When not logged in, how to use this method?

PR #2049, opened 2018-06-21
Realtime: Handle \Throwable in the same way as \Exception

dm0x1:
In php 7 Mqtt client can throw `\Throwable` and it can cause app crash. This commit fixes it with backward compatibility to php 5.x. 

Example:
https://3v4l.org/mWS3I


valga:
@dm0x1 Did you experience a crash when sending a command? If yes please provide a backtrace.

I am asking that because it is very unlikely that the RT client can crash there. The best place for it to crash is when you are handling an incoming event with your custom code: https://github.com/mgp25/Instagram-API/blob/f1ac06d915bb01b0a7825f01fe67369fec0ba7e6/src/Realtime/Mqtt.php#L788-L794

dm0x1:
I have some troubles not with RT client directly, but with net-mqtt-client-react, which in turn crashes RT client (throws \Error in php 7.x) in some cases. 
Unfortunately this commit doesn't fix it completely, so I'll investigate it and will contribute later. I'm sorry to bother you.

PR #2054, opened 2018-06-28
Update Constants.php

ch4kn:
Update to 52.0.0.5.83


mgp25:
Updating constants without doing a full analysis of the APK and checking what other changes happened is a HUGE error, that can lead to many issues. We won't be updating to anything newer until all the stuff in between is done.

Regards

PR #2057, opened 2018-06-30
Story: Get poll voters

mgp25:
This PR adds a new function that lets the user owning a story poll to get it results filtered by the voting option. 

Example:

- Story Poll has two options: A and B.

You will be able to get the voters (users) who has voted A or B, depending the option you pass to the function.

0 means A (first option)
1 means B (second option)


valga:
@mgp25 It looks good to me, but you should not mix new features with irrelevant models update into one commit.

mgp25:
@valga I comitted them by mistake in the first commit, then decided to add other updates too. I'll be more careful next time.

Regards

PR #2066, opened 2018-07-05
Only reset curl when the proxy has changed

WoLpH:
A little addendum to https://github.com/mgp25/Instagram-API/commit/fa6057c16d7ab646abf2a943254edcce4ef3c04d
Resetting the CURL connection is a pretty heavy operation so it shouldn't be done unless the proxy actually changed.


valga:
@WoLpH You can do the same with this:

```php
if ($ig->getProxy() !== $proxy) {
    $ig->setProxy($proxy);
}
```

PR #2070, opened 2018-07-07
Instagram TV

mgp25:
This PR contains requests of the new Instagram feature (Instagram TV) aka **IGTV** (Instagram module).

Requests:

- getTvGuide()
- getChannels()
- suggestedSearch()
- writeSeenState()

This PR also contains responses and models for each request.


src/Request/TV.php
```diff
@@ -0,0 +1,115 @@
...
+     * @throws \InstagramAPI\Exception\InstagramException
+     *
+     * @return \InstagramAPI\Response\TVChannelsResponse
+     */
+    public function getChannels(
+        $id = 'for_you',
+        $maxId = null)
+    {
+        if (!in_array($id, ['for_you', 'chrono_following', 'popular', 'continue_watching'])
+        && strpos($id, 'user_')) {
```
valga: 1. `strpos('random_thing', 'user_')` evaluates to `false`, so this check will not catch invalid channel ID. `strpos($id, 'user_') !== 0` is the right way.
2. But it is better to use regex here to ensure that user ID will be a valid number.

src/Request/TV.php
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @param string      $id    ID used to filter channels.
+     * @param null|string $maxId Next "maximum ID", used for pagination.
+     *
+     * @throws \InvalidArgumentException
+     * @throws \InstagramAPI\Exception\InstagramException
+     *
+     * @return \InstagramAPI\Response\TVChannelsResponse
+     */
+    public function getChannels(
```
valga: Method name is misleading. It should be named `getChannel()` or something like that, because it returns a single channel with its media. Description must be updated to match this.

src/Request/TV.php
```diff
@@ -0,0 +1,115 @@
...
+        return $this->ig->request('igtv/suggested_searches/')
+            ->addParam('query', $query)
+            ->getResponse(new Response\TVSuggestedSearchResponse());
+    }
+
+    /**
+     * Write seen state on a video.
+     *
+     * @param string $impression      Format: 1813637917462151382
+     * @param int    $viewProgress    Video view progres progress in seconds.
```
valga: `progres progress`

src/Request/TV.php
```diff
@@ -0,0 +1,115 @@
...
+    public function suggestedSearch(
+        $query = '')
+    {
+        return $this->ig->request('igtv/suggested_searches/')
+            ->addParam('query', $query)
+            ->getResponse(new Response\TVSuggestedSearchResponse());
+    }
+
+    /**
+     * Write seen state on a video.
```
valga: How often should it be called? Every N seconds or just after the end?

src/Request/TV.php
```diff
@@ -0,0 +1,115 @@
...
+     * @return \InstagramAPI\Response\TVWriteSeenStateResponse
+     */
+    public function writeSeenState(
+        $impression,
+        $viewProgress = 0,
+        $gridImpressions = [])
+    {
+        if (!ctype_digit($viewProgress) && (!is_int($viewProgress) || $viewProgress < 0)) {
+            throw new \InvalidArgumentException(
+                'View progress must be a positive integer.');
```
valga: I think it the message should be on one line with `throw`.

src/Request/TV.php
mgp25: The response contains `Item[]` which can have multiple `Item` of different channels. In the request I just made, it contains 4 `Item` in which 2 of them are media of a channel (owned by a user), the other Item is owned by other user, and the last Item the same.

Thats why I named it `getChannels`. Should I rename it?

src/Request/TV.php
mgp25: Typo fixed! 😄

mgp25:
@valga I am going to do media upload today. One thing I find strange is channel creation, I didn't find any request related to it. Maybe all of the users has their own channel but they just need to enable it in the GUI. I'll recheck this.

src/Request/TV.php
mgp25: Fixed.

src/Request/TV.php
mgp25: Fixed. Maybe there is a better regex, but so far the one I did worked.

src/Request/TV.php
mgp25: I think it is every seconds but let me check this again

src/Request/TV.php
valga: It's `^user_[1-9]\d*$`:
1. `user$` will also match `abuser`, so you have to add the caret.
2. There is no need to capture `user_`, so you should remove the parentheses.
3. `\d` is shortcut for `[0-9]`.

src/Request/TV.php
valga: `popular`, `for_you` and others are meta-channels, because they contain items from other channels.

The purpose of this method is to get a media from the channel (no matter if it is meta or not), so its name should be either `getChannel()` or `getChannelMedia()`.

src/Request/TV.php
valga: By the way, seen state format is invalid. It should be:
```php
[
    'impressions' => [
        '1813637917462151382' => [
            'view_progress_s' => 0,
        ],
        ...
    ],
    'grid_impressions' => [],
]
```

src/Request/TV.php
mgp25: Ah, I see. I misunderstood this endpoint. So the purpose for this is to obtain the media of user channels.

src/Request/TV.php
mgp25: You are right. At this point I don't think we need `view_progress_s` so I am not adding an unpredictible container.

mgp25:
@valga I think upload to TV should work okay, I need to verify if all values are okay. I think param `supported_capabilities_new` is not set anymore and i have to check if there is anything new in headers and so.

One thing we should have is the aspect ratios for this kind of media. Probably there are similar to story media?

balthazarneveu:
Hi @valga  and @mgp25  , i'm trying this branch (and help you debug) because i'm interested in this feature. I'm using Story, Video albums & live broadcast features in combination with Python,  works great! Good job .

So far i got getChannel working smoothly   ` $ig->tv->getChannel($id='popular'); `

I'm trying to upload a 20 second mp4 1080x1920 video named $videoFilename


```
$videoFilename = 'igtv_test.mp4'
$metadata = [
'caption' => 'NICE VIDEO',
'title' => 'test video',
];
$video = new \InstagramAPI\Media\Video\InstagramVideo($videoFilename, ['targetFeed' => \InstagramAPI\Constants::FEED_TV]);
$ig->tv->uploadVideo($video->getFile(), $metadata);
```


I cannot get it to work.... thumbnail's missing but i'm not sure how to get it right. 
Best regards
Balth

```
[1;34mPOST:  [0mhttps://i.instagram.com/rupload_igvideo/1531235364487_0_1256009852?target=FRC
[1;33mÔåÆ 13.69mB[0m
[0;32mÔåÉ 200   39B[0m
[0;36mRESPONSE: [0m{"xsharing_nonces": {}, "status": "ok"}

[1;34mGET:  [0mhttps://i.instagram.com/rupload_igphoto/1531235364487_0_-728504556
[1;33mÔåÆ 0B[0m
[0;32mÔåÉ 200   12B[0m
[0;36mRESPONSE: [0m{"offset":0}

[1;34mPOST:  [0mhttps://i.instagram.com/rupload_igphoto/1531235364487_0_-728504556
[1;33mÔåÆ 17.02kB[0m
[0;32mÔåÉ 200   69B[0m
[0;36mRESPONSE: [0m{"upload_id": "1531235364487", "xsharing_nonces": {}, "status": "ok"}

Something went wrong: Upload of "INS130E.tmp" failed: All configuration retries have failed.
```

mgp25:
@balthazarneveu it is not yet finished, plus you are using something it is not implemented:

`$video = new \InstagramAPI\Media\Video\InstagramVideo($videoFilename, ['targetFeed' => \InstagramAPI\Constants::FEED_TV]);`

We recommend you to wait until we finish it.

balthazarneveu:
Ok no worries, i'll be waiting for sure and i'll test it in depth whenever you notify me. Thanks a lot @mgp25

balthazarneveu:
Hi, a bit of first feedback
By providing the direct path to the 1080x1920 mp4 video (converted with moviepy because the rest of my backend is python), i was able to upload a full 3minutes IGTV so i'm pretty satisfied ! Thanks a lot. 
By the way, I checked a bit the code if i would have wanted to do the conversion/checks in PHP and definitely there's still a bit of work to finish that PR properly in all cases  (constraints and videoDetails but I wouldn't dare modifying php too much).

Best regards

valga:
@mgp25 Here is a checklist for uploads:

- [ ] Create a set of constraints for TV. I think it is ok to inherit it from story ones, because only min and max durations are different.
- [ ] Add these constraints into the factory https://github.com/mgp25/Instagram-API/blob/36a92c944ac5b3117331a85a71b40a8220f6d349/src/Media/Constraints/ConstraintsFactory.php#L22
- [ ] Check whether segmented uploader is used for TV and adjust thresholds here https://github.com/mgp25/Instagram-API/blob/36a92c944ac5b3117331a85a71b40a8220f6d349/src/Request/Internal.php#L1997
- [ ] Adjust segment duration for TV (if needed to) https://github.com/mgp25/Instagram-API/blob/36a92c944ac5b3117331a85a71b40a8220f6d349/src/Request/Internal.php#L2314
- [ ] Check whether resumable uploader is used for TV https://github.com/mgp25/Instagram-API/blob/36a92c944ac5b3117331a85a71b40a8220f6d349/src/Request/Internal.php#L2076

Leadro:
Hi all,
I've made few tests for TV branch and found strange behavior for "suggestedSearch" request.
It's always return same result.
Doesn't matter what query I've put to request.

Here response for empty query:
[Response with empty query](https://pastebin.com/rNEWuhmT)

Here response for query "nnnn1w":
[Response with query **nnnn1w**](https://pastebin.com/ZqAN770A)

Hope this will help.

mgp25:
@valga I have been checking experiments to check wether there is a different experiment for the TV media, and I think IG is using experiments from Story. I would add in the `switch-case` statement:

```php
            case Constants::FEED_TV:
```

To fit with the `STORY` and `DIRECT_STORY`.

I can confirm TV uses resumable uploader.

regards

valga:
@mgp25 They just hardcoded all values for the TV, so no experiments for it. See 7f18620.

valga:
@mgp25 Mind the difference between `self` and `static`:
```php
class Root {
  const ID = 'root';
  
  public function id() {
    return self::ID;
  }
}

class Child extends Root {
  const ID = 'child';
}

echo (new Child())->id(); // Will output "root", because self means the declared class.
```

```php
class Root {
  const ID = 'root';
  
  public function id() {
    return static::ID;
  }
}

class Child extends Root {
  const ID = 'child';
}

echo (new Child())->id(); // Will output "child", because static means the current class.
```

So your `TvConstraints` will not work as intended. Also, you forgot to override the `getTitle()` method.

src/Media/Constraints/TvConstraints.php
```diff
@@ -0,0 +1,25 @@
...
+    const MIN_DURATION = 15.0;
+
+    /**
+     * Maximum allowed video duration.
+     *
+     * For verified accounts, this duration can be up to 3600.0.
+     *
+     * @var float
+     */
+    const MAX_DURATION = 600.0;
```
valga: I think we should change it to 3600, because there is a chance that someone with a verified account will use our library.

By the way, what happens when a non-verified account tries to upload super long video? Do they block it on configure?

valga:
@Leadro It should be fixed in 15b980b.

trungducng:
Hey guys, Just a question: Is  "upload to TV" function ready to use?

valga:
@trungducng I will merge it tomorrow.

PR #2073, opened 2018-07-08
Update Live.php

sorset90:
Adds burst like in live broadcast....


mgp25:
Hello @sorset90,

The function already exists: https://github.com/mgp25/Instagram-API/blob/master/src/Request/Live.php#L325

You only need to update it and add that param (`user_like_burst_count`). You also need to know which values must have that param in order to document it.

Your other PR (https://github.com/mgp25/Instagram-API/pull/2072) is not required since the response already exists: https://github.com/mgp25/Instagram-API/blob/master/src/Request/Live.php#L338

And you need to be careful about the changes you make: https://github.com/mgp25/Instagram-API/pull/2073/files#diff-7b72e166d9f6411ea83e0307319da436R352

Feel free to open a new pull request with the required changes.

Regards

PR #2078, opened 2018-07-10
Can not override already exists user-agent header

4n70w4:
Case:
$ig->request($url)->setAddDefaultHeaders(false)->addHeader('User-Agent', 'my user agent')....

Expectation: make request with my user agent


valga:
@4n70w4 You can't override `User-Agent` header because it could break other things.

PR #2082, opened 2018-07-11
add setter and getter for \InstagramAPI\Client::$_userAgent

4n70w4:


valga:
@4n70w4 Same reason as for #2078.

PR #2085, opened 2018-07-12
ability to redefine radio_type in requests

4n70w4:


valga:
@4n70w4 The problem with (almost) all your PRs is you don't want to improve or fix the library. You just want to patch it to use it in your personal interests (whatever they are). We can't accept proposed patches, because they will make the library fragile.

For example, you can't change `radio_type` alone in requests, because current network has a lot of other effects, including:
1. `Connection-Type` header in HTTP requests https://github.com/mgp25/Instagram-API/blob/36a92c944ac5b3117331a85a71b40a8220f6d349/src/Request.php#L380
2. Network type and network subtype in the Realtime client: https://github.com/mgp25/Instagram-API/blob/36a92c944ac5b3117331a85a71b40a8220f6d349/src/Realtime/Mqtt.php#L530-L533
3. Network type and network subtype in the Push client: https://github.com/valga/fbns-react/blob/master/src/Connection.php#L55-L58
4. You also have to call MSISDN-related endpoints when "using" cellular network.
5. Whatever I forgot to mention.

You have to change all of them at once, so the proper solution would be to create `NetworkInterface` that will look something like that:
```php
interface NetworkInterface {
  public function connectionType();
  public function networkType();
  public function networkSubtype();
  public function radioType();
}
```
Then just inject it via `Instagram` constructor and use it when needed. Also `WiFiNetwork` and `CellularNetwork` are required as default implementations. That is harder then your patches, because it requires some effort, so consider using https://github.com/cweagans/composer-patches or https://github.com/vaimo/composer-patches if you don't want to collaborate.

4n70w4:
@valga it does not make anything fragile, are you out of your mind? This allows to override the settings for those who need it. By default, everything works the same as in your version. It's true improve. Who does not need it, do not use it.

PR #2088, opened 2018-07-14
cpu_abi is not a constant

4n70w4:


valga:
@4n70w4 The proper way is to modify the `Device` class to include ABI into its definition. Also see https://github.com/mgp25/Instagram-API/pull/1105#issuecomment-294375717

4n70w4:
@valga proper? seriously? are you aware of the trends in programming, or do you live in the 1980s?

PR #2092, opened 2018-07-16
Add ability to realtime modify Guzzle Handlers and Middleware

4n70w4:


JRoy:
Not only does this have conflicts with master, but it's dangerous to allow users to modify this and may lead to issues with bug reports.

Regards.

Issue #2145, opened 2018-08-08
Transcode error: Video file does not contain duration


taha:
### Before submitting an issue make sure you have:
- [x] [Searched](https://github.com/mgp25/Instagram-API/search?type=Issues) the bugtracker for similar issues including **closed** ones
- [x] [Read the FAQ](https://github.com/mgp25/Instagram-API/wiki/FAQ)
- [x] [Read the wiki](https://github.com/mgp25/Instagram-API/wiki)
- [x] [Reviewed the examples](https://github.com/mgp25/Instagram-API/tree/master/examples)
- [x] [Installed the api using ``composer``](https://github.com/mgp25/Instagram-API#installation)

### Purpose of your issue?
- [ ] Bug report (encountered problems/errors)
- [ ] Feature request (request for new functionality)
- [x] Question
- [x] Other

---

Hi,

I'm using the Timeline::uploadVideo method as described in the example. However, Instagram rejects ~50% of the videos I've tested with (the error below). My code is literally identical to the example's (also using `\InstagramAPI\Media\Video\InstagramVideo` for video processing).

I've tried converting the failed videos with various tools (e.g. HandBrake on macOS), even tried to manually set the duration for the video in ffmpeg (using `-t` argument) with no luck.

Any idea how to work around this issue?

**EDIT:** I forgot to mention that the same rejected videos by Timeline::uploadVideo work fine when fed to Timeline::uploadAlbum

**Code**
```php
$video = new \InstagramAPI\Media\Video\InstagramVideo($videoFilename);
$ig->timeline->uploadVideo($video->getFile(), ['caption' => $captionText]);
```

**Error Response**
```
400 	 86B
RESPONSE: {"message": "Transcode error: Video file does not contain duration", "status": "fail"}
RuntimeException: All configuration retries have failed.
```

mgp25:
Try this:

```php
$video = new \InstagramAPI\Media\Video\InstagramVideo($videoFilename, ['targetFeed' => \InstagramAPI\Constants::FEED_TIMELINE]);
```

You can change the target feed to fit your requirements. Let me know if that worked.

valga:
@taha Please share the video file you have problems with.

taha:
@mgp25 that didn't work :/

@valga [test.mp4.zip](https://github.com/mgp25/Instagram-API/files/2272097/test.mp4.zip) (A video I grabbed of an Instagram video post)

valga:
[...]

PR #2155, opened 2018-08-11
GLOB_BRACE workaround for non GNU systems

taha:
GLOB_BRACE flag used is not supported on non-GNU systems, this leaves out Alpine. Came across it while troubleshooting https://github.com/mgp25/Instagram-API/issues/2145 (which was indeed the cause of this issue as it wasn't uploading the video segments)


valga:
@taha Thanks for your finding, but knowing that audio track always goes into one segment, we can replace your 2nd `glob()` with `is_file()` check. I will do it in 1-2 days.

By the way, is `GLOB_BRACE` defined in PHP build for Alpine?

taha:
@valga Thanks. `GLOB_BRACE` was not defined, but it only threw a warning in my environment which I missed.

PR #2172, opened 2018-08-23
added ext-bcmath to composer.json

kozintsev:


valga:
@kozintsev Thanks. Is it also required in our Dockerfile? 

https://github.com/mgp25/Instagram-API/blob/2fee96d3107bb23ec791e476ab04215fec89325e/Dockerfile#L14-L17

kozintsev:
Yes, Probably.

PR #2379, opened 2018-12-14
Fix maxDuration in InstagramVideo.php

nilsalexberg:
I uploaded a video for instagram story with exactly 15s and 25 FPS. 
When creating a new `\InstagramAPI\Media\Video\InstagramVideo` instance, `$this->_details->getDuration()` was equal to `$this->_constraints->getMaxDuration()`.
But when I called `$Instagram->story->uploadVideo`, I got this error: "Instagram only accepts story videos that are between 3.000 and 15.000 seconds long. Your video "INSTAVID_ZAPhOS" is 15.040 seconds long."
Probably, after the cropping process, my video got 1 frame longer.
To solve this, `$maxDuration` should be assigned to `$this->_constraints->getMaxDuration() - 0.1` before the `if` statement.


JRoy:
Thanks for your contribution!

PR #2428, opened 2019-01-17
Update VideoDetails.php

aliazizi:
Instagram is not checking the floating point of duration so it should cast otherwise it may throw `Instagram only accepts timeline videos that are between 3.000 and 60.000 seconds long. Your video  is 60.060 seconds long`


JRoy:
Hi @aliazizi

Can you please conform to the style guidelines?

Regards

JRoy:
Closing due to the submitter not adhering to style guidelines.

PR #2452, opened 2019-01-26
Vote slider story

pashmak73:


JRoy:
Commit already added in #2451

PR #2463, opened 2019-01-29
Add support of thumbnail in albums

e4se:
Added support of thumbnail in albums


JRoy:
@e4se could you please include examples and have you tested this?

mgp25:
Does Instagram allows setting your own thumbnail? Because if not, we shouldn't add this.

Regards

JRoy:
@mgp25 I'm not familiar on how the thumbnail system works from an API standpoint but I know you can pick frames from the video in the app to set as the thumbnail (for post video and album posts)

e4se:
Add example for this. Of cause I tested this feature and Instagram allows pick frames from the video to set as the thumbnail for every video in albums)

JRoy:
Well if it works, it works. I know Instagram allows for this behavior and feature looks good to me. I'll wait for mgp's opinion before merge tho...

mgp25:
If this works and album supports it, then perfect. There is a comment that you be removed:  `// NOTE: This is NOT for albums!` since now it also for albums

e4se:
Remove outdated comment

JRoy:
Thanks for your contribution!

PR #2476, opened 2019-02-02
Live: Add wave feature

JRoy:
While doing my research for Live with Friends, I found that the wave feature wasn't added, as it's a simple request I added real quick, hence this pull request. This pull request also fixes some style guidelines from #2475 (whoops).

Example:
```php
// Detect a user joins the livestream
$ig->live->wave($broadcastId, $userId)
```


mgp25:
Everything seem okay. Just a question: Is `vievwer_id` the same as `user_id`?

JRoy:
@mgp25 Yep just tested this and it requires user ids. The reason I named it viewer id is so that people know that the person must be viewing the stream.

JRoy:
Merging due to no objections

PR #2497, opened 2019-02-11
Update Location.php

yussufs:
Don't sign the post request, otherwise we will get the same response (with ranked items) no matter what.


JRoy:
Thanks for your contribution!

mgp25:
Has this been verified with official request?

JRoy:
@mgp25 Yeah, I tested it before I merged. When signed the request *does* work but as the submitter pointed out, the responses are the same. When unsigned, however, the behavior is as expected.

PR #2519, opened 2019-02-18
Bugfix/story slider on videos

tobiasmanroth:
Added metadata to the actual story video post for story sliders


JRoy:
@mgp25: Pull-request matches android app behavior, I also added `emoji_slider_` prefix to the `story_sticker_ids` parameter as that's how the app does it now.

@tobiasmanroth: Thanks for your contribution!

PR #2500, opened 2019-02-12
Add Question Sticker Support

JRoy:
This pull request adds the ability to answer a question sticker as well as updates the Item model to support viewing question sticker responses (as well as fetch their ids).

The new endpoint has been fully tested and works as expected.

Examples:
```php
$storyId = "1977230768527962393";
$questionId = "18036081463043385";
$ig->story->answerStoryQuestion($storyId, $questionId, "Hello World!")
```

PR #2520, opened 2019-02-18
Add Story Questions Upload Support

JRoy:
This pull request finishes what #2500 started by allowing question stickers to be uploaded via metadata. This sticker supports both photo and video uploads and these endpoints have been fully tested! Additionally, while researching, I found this sticker uses the `z` axis as one of its parameters (Maybe AR Stickers?) but it is always `0` as of now.

This pull request also adds a corresponding example for this feature as users are not expected to just "know" how to make a questions sticker.

Only pay attention to 23d2486, 3ac326f, and a04d4b8 as all the other commits are from PhpStorm freaking out, once the commits are squashed it won't matter anyway :P


mgp25:
@JRoy everything seems okay, but `throwIfInvalidStoryQuestion()` doc is missing. Add it before merging.

Great work 👍 

Regards

PR #2539, opened 2019-02-27
Add Story Countdowns Support

JRoy:
This pull request adds support for the following actions:
* Get a story's countdown data via `Item#getStoryCountdowns()`
* Allow story viewers to subscribe/unsubscribe to/from countdown notifications via `Story#followStoryCountdown()` and `Story#unfollowStoryCountdown()` respectivly
* Upload/create countdowns to story
  * Example with `examples/uploadStoryCountdown.php`
* Fetch created countdowns via `Story#getStoryCountdowns()`

Everything has been fully tested as always and this pull request is **ready** for merge.


examples/uploadStoryCountdown.php
```diff
@@ -0,0 +1,67 @@
...
+            'width'                  => 0.6564815, // Clickable area size, as percentage of image size: 0.0 - 1.0
+            'height'                 => 0.22630993, // ...
+            'rotation'               => 0.0,
+            'text'                   => 'NEXT API RELEASE', // Name of the countdown. Make sure it's in all caps!
+            'text_color'             => '#ffffff',
+            'start_background_color' => '#ca2ee1',
+            'end_background_color'   => '#5eb1ff',
+            'digit_color'            => '#7e0091',
+            'digit_card_color'       => '#ffffff',
+            'end_ts'                 => time() + 432000, // UNIX Epoch of when the countdown expires.
```
mgp25: Maybe you could change this line to:

```php
'end_ts'                 => time() + strtotime('1 day', 0),
```

For me it is more understandable in a very first sight.

mgp25:
@JRoy So far seems good! I only commented one thing in the exaple. Have a look and let me know if you think its a good proposal. It is ready for merge.

Great work 👍

examples/uploadStoryCountdown.php
JRoy: Makes sense for users who are not familiar with the epoch.

PR #2566, opened 2019-03-14
Direct: added new parameter for sending hidden links

iwex:
Sometimes it's useful to send links which are hidden in the main text.


iwex:
`composer codestyle` is different than `styleci`

JRoy:
Hi, is this feature on the Android app, if so how do I trigger this?

iwex:
@JRoy Hi!
It works only through API, you can't send it directly over the app.

iwex:
I will look like
image

JRoy:
I swear I've seen this in the app before, but I don't remember how to do it. Until we can figure out how this is triggered via the app, I don't want to add this as this API is supposed to emulate the Android App.

iwex:
@JRoy It's backward compatible, so why not to add this functionality?

JRoy:
Whether it works or not is against the point, if it doesn't actually happen in the app, we like to avoid it

iwex:
@JRoy, @mgp25 said "However, we can document these changes in the wiki so anyone under their own risks can perform this."
I don't have access to Wiki pages, so should I add just a comment to PhpDoc?

iwex:
```php
$ig->direct->sendText($recipients, $text, array $options = [], $hiddenUrls = []);
```
If `$text` contains a link, it will be processed and sent as a link.

Instagram allows sending links which will be hidden from the message body. **This functionality is not present in the official Instagram application, so use it on your own risk!**
To send hidden links - just pass them as $hiddenUrls (array) parameter.

Issue #2540, opened 2019-02-28
gzip some endpoints


ivanyadeshko:
*improvement*
Hi, to every one!
I'm detecting strange Instagram behavior on feed/timeline endpoint.
On my phone i see many pagination requests, an see pagination requests, But POST body encripted O_o ... and cannot be read.

In Instagram-API script, POST body send on Instagram not cripted. I think, next time will cause considerable problems.

APP: 76.0.0.15.395
Device: Nexus 5

ivanyadeshko:
And i am dont see pagination point on instagram response
http://joxi.ru/12MWx9OhlgLyWr
but in my phone over 100 medias in timeline feed

JRoy:
Will look into this

valga:
@ivanyadeshko It's not encrypted. It's just compressed with gzip. 

@JRoy This behavior is controlled with `ig_android_request_compression_universe` experiment. It has following params:
* `use_for_feed` - `feed/timeline/` endpoint;
* `use_for_qe` - `qe/sync/` endpoint;
* `use_for_stories` - ??

When enabled, POST body must be compressed with gzip, and headers must contain `Content-Encoding: gzip` one.

JRoy:
@valga I just tested those experiments and they appear to not be present in the account I know to be using gzipped content. Maybe in v82, gzip is the norm?

valga:
@JRoy This experiment has been changed in 82. Now it's called `ig_android_request_compression_launcher` and has only 2 params:
* `use_for_feed`;
* `use_for_stories`.

Also gzip compression is now always enabled for `address_book/link/` endpoint and some ads-related one.

PR #2568, opened 2019-03-15
Update Highlight Requests/Models

JRoy:
Changes:
* Added Params for `highlights_tray/`
* Removed `crop_rect` param from covers
  * The app *still* sends these, but the server will accept and solve these without us having to do anything
* Updated StoryTray models

Everything has been tested and this fixes #2560

PR #2569, opened 2019-03-15
Changes for v85

JRoy:
This pull request is ready for merge

---

This pull request is basically just for discussion for the changes needed for v82 to work. (`v82-changes` branch)

Things that need to be done:
* Constants:
  * ~Experiments~
  * ~Signature Key~
* ~GZip Endpoints~
* Maybe test every endpoint :D

### GZipping:
As pointed out in @valga's comment: https://github.com/mgp25/Instagram-API/issues/2540#issuecomment-468937559 feed and stories have an experiment and respective parameters for using gzip compression and address book linking is enabled all the time. 

As for implementation of this, I'm not aware of any built-in guzzle gz-compressor so we'd either need to compress the post data to gzip in `Request#_buildHttpRequest()` or `Request#_getRequestBody()`. As for how those methods should obtain the data, we should probably just add a function to the `Request` class like `setGzCompressedBody()` and set it to `true` when the experiments and their respective parameters are detected.

woah that was a lot of text for 10 pm


valga:
@JRoy You can find PSR-7 gzip stream [here](https://github.com/cyberdummy/gzstream).

But I think it will be also ok to use `zlib_encode()` with `ZLIB_ENCODING_GZIP`, because those requests sizes aren't that big.

JRoy:
This pull request now concerns version `85.0.0.21.100` of the instagram app.

GZipping, Signature Key, and Experiments still need to be done before we can consider this ready for merge.

@mgp25: Any updates on the new signature key location?

JRoy:
Alrighty, the experiments from `85.0.0.21.100` have been added after I just shoved a base64 compressor into the `qe/sync/` smali code to fit the experiments into the logcat. After we find where the new signature key is, we'll be good for constants.

@valga: I was going to implement request gzipping but found `ig_android_request_compression_launcher` not in `85.0.0.21.100`'s experiments. Is gzipping enabled all the time now or is there *another* new experiment.

I just spent about 5 hours researching dalvik byte-code and my now brain hurts 😄 so I'm going to goto sleep.

Cheers.

valga:
@JRoy Yeah, it has been changed again:

* `address_book/link/ ` - always;
* `feed/timeline/` - always;
* `media/seen/` - always;
* `feed/injected_reels_media/` - always - it's that ads-related one (quote: "Ad$ Revenue Impacted: FIX IMMEDIATELY!");
* `address_book/get_contact_hashes/` - only when `enabled` param is on in `ig_android_enable_request_compression_ccu` experiment.

JRoy:
GZipping has been added and this pull request now solves the issue that sparked it: #2540

All changed code is tested and is ready for merge. Updated signature key is still needed but all features are done. 

cc @mgp25 @valga Please review and share any feedback you have.

mgp25:
@JRoy I am reviewing your code. I will post 85.0.0.21.100 signature in a couple of minutes

JRoy:
Sounds good 👍

mgp25:
@JRoy 

`85.0.0.21.100`: `937463b5272b5d60e9d20f0f8d7d192193dd95095a3ad43725d494300a5ea5fc`

I am finishing other things at the moment but will check all the code tonight ;)

mgp25:
@JRoy so far seems good to me. @valga any words?

valga:
@JRoy You forgot about `Content-Encoding` header. Also it seems that there is some threshold for body size, so small one won't be gzipped. Will check that later.

@mgp25 It seems that they don't care about signature key anymore. It gives us two options:
1. They don't need anymore. For example, they have something more powerful now.
2. They silently flag clients with invalid signatures. 

By the way, I saw a couple of new headers: `X-Pigeon-Session-Id` and `X-Pigeon-Rawclienttime`. Second one is obvious, first one needs to be investigated.

mgp25:
@mgp25 I am already working on that ;) but going slowly because of other side projects :/

JRoy:
@valga Added the content encoding header. As for the threshold, I doubt we would ever reach it, the smallest request that's gzipped, `media/seen/`, is still gzipped by the app with only one story. Therefore I doubt we would run into any issues in the library.

src/Request.php
```diff
@@ -493,6 +503,23 @@ public function setIsMultiResponse(
...
+     * @param bool $isBodyCompressed
+     *
+     * @return self
+     */
+    public function setIsBodyCompressed(
+        $isBodyCompressed = false)
+    {
+        $this->_isBodyCompressed = $isBodyCompressed;
+
+        $this->_headers['Content-Encoding'] = 'gzip';
```
valga: It' a bug: one can call `setIsBodyCompressed(false)` causing uncompressed body with gzip encoding.

src/Request/Live.php
```diff
@@ -482,8 +483,9 @@ public function getPostLiveLikes(
      *
      * @param int    $previewWidth     (optional) Width.
      * @param int    $previewHeight    (optional) Height.
-     * @param string $broadcastMessage (optional) Message to use for the broadcast.
+     * @param string $broadcastMessage (optional) Title to use for the broadcast.
```
valga: As far as I know, it is impossible to set a title for more than year. Am I wrong?

src/Request/Timeline.php
```diff
@@ -226,12 +227,15 @@ public function getTimelineFeed(
             ->addPost('phone_id', $this->ig->phone_id)
             ->addPost('device_id', $this->ig->uuid)
             ->addPost('client_session_id', $this->ig->session_id)
-            ->addPost('battery_level', '100')
-            ->addPost('is_charging', '1')
+            ->addPost('battery_level', mt_rand(25, 100))
+            ->addPost('is_charging', '0')
```
valga: Why did you change that? I think that always-on-charge mode is more suitable for this library.

src/Request/Live.php
JRoy: I don't know what you mean by year but Instagram is using broadcast message to title livestreams now. Was a new feature in 80 I believe.

Please clarify.

JRoy: Actually just checked the app, this feature has been removed from Instagram again :(

Will revert this

src/Request/Timeline.php
JRoy: I changed this to be consistent with changes mgp suggested here: https://github.com/mgp25/Instagram-API/pull/2568#issuecomment-473269160

8lbiasian: I think even is_charging should be random with a high percent of appearing as not charging for example, as 80% chance should be not charging while 20% chance it's charging. 

idk if IG system analyzes this also but it's even more realistic. 

and sorry to interrupt.

bolds07: > I changed this to be consistent with changes mgp suggested here: [#2568 (comment)](https://github.com/mgp25/Instagram-API/pull/2568#issuecomment-473269160)

`->addPost('battery_level', mt_rand(25, 100)`

To have a full random battery level is really not a good idea in the "realistic mode" two consecutive requests will have very different battery levels what would never happen in a real device

imagine now the random is 99, then 5 seconds later the random is 30

if you want to have a random value for that i would recomend create a random at session begin (when login refresh login) and then keep it

8lbiasian: > > I changed this to be consistent with changes mgp suggested here: [#2568 (comment)](https://github.com/mgp25/Instagram-API/pull/2568#issuecomment-473269160)
> 
> `->addPost('battery_level', mt_rand(25, 100)`
> 
> To have a full random battery level is really not a good idea in the "realistic mode" two consecutive requests will have very different battery levels what would never happen in a real device
> 
> imagine now the random is 99, then 5 seconds later the random is 30
> 
> if you want to have a random value for that i would recomend create a random at session begin (when login refresh login) and then keep it

but if the session is up for 1 month without rel ogin or any expiry so it will do all actions with that value which can`t be a real user also is that correct?

bolds07: maybe i didn't express very well...

login is refreshed every 1800 seconds, i meant this when i said session

 ```
protected function _login(
        $username,
        $password,
        $forceLogin = false,
        $appRefreshInterval = 1800)
    {
```

Comment (8lbiasian) [reply src/Request/Timeline.php]: > maybe i didn't express very well...
> 
> login is refreshed every 1800 seconds, i meant this when i said session
> 
> ```
> protected function _login(
>        $username,
>        $password,
>        $forceLogin = false,
>        $appRefreshInterval = 1800)
>    {
> ```

ah ok then yes I agree with you it's more logical and real waiting for big guys to check this.

Comment (JRoy) [reply src/Request/Timeline.php]: @bolds07 brings up a good point and considering that the timeline endpoint can be called rapidly due to pagination. Will revert changes

**Comment by JRoy** [comment]
@valga Your changes have been applied, how does it look now?

**Comment by itsMoji** [comment]
@JRoy I've tested some requests with Instagram version 85 and find out that some endpoints like `media/{$mediaId}/like/`, `friendships/create/{$userId}/` and maybe more has the `device_id` post param that is not exist in current version of this api or your PR.

**Comment by JRoy** [comment]
@mgp25 I've changed a lot of requests so if you would like to review my recent changes again and let me know how they look.

PS: I understand there are conflicts, I am able to resolve them when it's time to merge.

File: src/Request.php [hunk 4]
```diff
@@ -493,6 +503,27 @@ public function setIsMultiResponse(
...
+     * @return self
+     */
+    public function setIsBodyCompressed(
+        $isBodyCompressed = false)
+    {
+        $this->_isBodyCompressed = $isBodyCompressed;
+
+        if ($isBodyCompressed === true) {
+            $this->_headers['Content-Encoding'] = 'gzip';
+        } elseif ($isBodyCompressed === false && $this->_headers['Content-Encoding'] === 'gzip') {
```
valga: Should be changed to 
```php
elseif (isset($this->_headers['Content-Encoding']) && $this->_headers['Content-Encoding'] === 'gzip')
```

src/Request/Account.php
```diff
@@ -680,8 +659,10 @@ public function setContactPointPrefill(
         return $this->ig->request('accounts/contact_point_prefill/')
             ->setNeedsAuth(false)
             ->addPost('phone_id', $this->ig->phone_id)
-            ->addPost('usage', $usage)
             ->addPost('_csrftoken', $this->ig->client->getToken())
+            ->addPost('_uid', $this->ig->account_id)
+            ->addPost('_uuid', $this->ig->uuid)
```
valga: Anonymous requests do not require `_uid` and `_uuid` params, as these are only for authed.

src/Request/Collection.php
```diff
@@ -59,26 +60,24 @@ public function getFeed(
...
         return $this->ig->request('collections/create/')
             ->addPost('module_name', $moduleName)
-            ->addPost('added_media_ids', json_encode(array_values($mediaIds)))
+            ->addPost('collection_visibility', '0') //Instagram is planning for public collections soon
+            ->addPost('_csrftoken', $this->ig->client->getToken())
+            ->addPost('_uid', $this->ig->account_id)
             ->addPost('name', $name)
             ->addPost('_uuid', $this->ig->uuid)
-            ->addPost('_uid', $this->ig->account_id)
-            ->addPost('_csrftoken', $this->ig->client->getToken())
```
valga: Was it really necessary to move those params? I mean, we will rearrange them later in `Utils::reorderByHashCode()` anyway.

src/Request/Live.php
```diff
@@ -482,8 +484,9 @@ public function getPostLiveLikes(
      *
      * @param int    $previewWidth     (optional) Width.
      * @param int    $previewHeight    (optional) Height.
-     * @param string $broadcastMessage (optional) Message to use for the broadcast.
+     * @param string $broadcastMessage (optional) Don't use.
```
valga: What's the point in keeping this argument?

src/Request/Collection.php
JRoy: No, Sorry :P Was mostly just a sanity thing when researching.

JRoy:
@valga Your request changes have been added, now waiting on @mgp25 for merge permission.

mgp25:
@JRoy I have taken a look to the changes and seems good so far! So you have my approval to merge this PR.

Great work @JRoy! 👍 and thanks as always @valga for your reviews!

JRoy:
Will resolve conflicts and merge shortly!

PR #2582, opened 2019-03-25
Solve concurrency issues in atomic write of the cookie

atyachin:


valga:
@atyachin You did not solve concurrency issues, you just worked around them. The problem with `Utils::atomicWrite()` is it uses static suffix for temp filename. I see 2 scenarios when it fails.

1. `rename()` is called when another thread already renamed tmp file - with 2 threads:
     * thread 1 calls `file_put_contents()` - success;
     * thread 2 calls `file_put_contents()` - success;
     * thread 1 calls `rename()` - success;
     * thread 2 calls `rename()` - fail.
2. `unlink()` is called after another thread called `file_put_contents()` - with 3 threads:
     * thread 1 calls `file_put_contents()` - success;
     * thread 2 calls `file_put_contents()` - success;
     * thread 1 calls `rename()` - success;
     * thread 2 calls `rename()` - fail;
     * thread 3 calls `file_put_contents()` - success;
     * thread 2 calls `unlink()` - success;
     * thread 3 calls `rename()` - fail.

So, the proper solution is to always use random suffix (for example, `uniqid()`), so every thread will use its own temp file. But it could produce a lot of garbage when script is killed between `file_put_contents()` and `rename()`.

atyachin:
@valga You are correct in the sense that it's not avoiding the issue, but rather working around it using a busy-waiting approach. The scenario we are facing is that we have multiple scripts (different php processes) that run simultaneously and perform login to the same user. The failure is during the login process (that writes to the cookie files). I'm not sure if your suggestion will work as all processes need to use the same cookie files during the login process.

valga:
@atyachin `rename()` call is atomic, so it is safe to use it with random filenames.

But take in mind that our cookie storage is not designed for parallel jobs, because it doesn't pick up changes from the file when it was changed by another process.

atyachin:
@valga Can you suggest a better approach to run scripts in parallel that are handling the same IG user?

valga:
@atyachin The best way is to create an HTTP server that will accept certain commands to be executed on internally running `Instagram` objects. Something like I did in [realtimeHttp](https://github.com/mgp25/Instagram-API/blob/4c18d7f2c4f17a0d9c1cfe5d7c6fa0c9b0ff754b/examples/realtimeHttp.php) example. But unfortunately our library doesn't have async functionality (although guzzle has), so it will block execution on outgoing requests.

Another approach is to create a job processing server, where multiple instances will pick up jobs from the queue. It could work for you.

Anyway, this library was made for personal use, so you shouldn't consider it as enterprise-ready solution.

atyachin:
@valga Is there any reason why not to merge my suggested changes?

valga:
@atyachin I have to run some tests. I will merge your patch if it survives them.

atyachin:
@valga Thanks, please let me know if there are any issues.

atyachin:
@valga any update regarding the tests?

src/Settings/StorageHandler.php
```diff
@@ -509,7 +509,14 @@ public function setCookies(
                 // Perform an atomic diskwrite, which prevents accidental
                 // truncation if the script is ever interrupted mid-write.
                 $this->_createCookiesFileDirectory(); // Ensures dir exists.
-                $written = Utils::atomicWrite($this->_cookiesFilePath, $rawData);
+                $timeout = 15;
```
valga: You should decrease it to some reasonable value, because race conditions are indistinguishable from storage failures, so it would be wasteful to wait for 15 seconds when you have something like wrong permissions.

I think 4 seconds would be fine.

src/Settings/StorageHandler.php
```diff
@@ -509,7 +509,14 @@ public function setCookies(
...
+                $init = time();
+                while (!$written = Utils::atomicWrite($this->_cookiesFilePath, $rawData)) {
+                    usleep(500000);  // 0.5 sec
```
valga: You should randomize delay, because when more than one thread fail, they will retry (almost) in the same time, causing one more failure.

src/Settings/StorageHandler.php
atyachin: i went with 5 secs.

atyachin:
@vlaga - your suggestions were implemented: timeout = 5 secs, randomized the retry delay. Please let me know if we're good to merge. Thanks!

valga:
@atyachin Merged, thanks for your work on it.

PR #2583, opened 2019-03-25
Update type of profile_chaining_secondary_label

Mineroboter:
Error thrown when json_encode() of $ig->people->getInfoById($id)->getUser()

> LazyJsonMapper\Exception\LazyJsonMapperException: Unable to convert data to stdClass: Bad JSON property definitions in "InstagramAPI\Response\Model\ChainingSuggestion": (['Unexpected inner array value in non-array inner property for "profile_chaining_secondary_label", where we expect value type "string".']). in /home/pi/instagram/vendor/lazyjsonmapper/lazyjsonmapper/src/LazyJsonMapper.php:1174

This value can also be string[] instead of just string
e.g.:

> ["profile_chaining_secondary_label"]=>
>         array(2) {
>           [0]=>
>           string(9) "Suggested"
>           [1]=>
>           string(79) "Suggested because you may be interested in following. It's the default context."
>         }

Therefore the comment above ("This currently always contains the same value as "full_name".") is wrong and should be removed.


Mineroboter:
> Therefore the comment above ("This currently always contains the same value as "full_name".") is wrong and should be removed.

Or maybe someone has a proposal how this value can be properly documented.

JRoy:
Can you provide an example user id where you get the `profile_chaining_secondary_label` param? I haven't been able to find an account that produces it.

Mineroboter:
If you request the User-Info of an account, you'll also get the value "chaining_suggestions". Only one account out of 42 in this value provides profile_chaining_secondary_label as string[]. Seems to be very rare, and only if Instagram can't correctly identify the context. Maybe it is an error or old value, which isn't used anymore.

I don't know if this value depends on which user crawls the User-Info, but here is an example: 

Example Account for User-Info:
353355020
Example Account, mentioned in chaining_suggestions of 353355020 having string[] as profile_chaining_secondary_label:
2264194707

JRoy:
Thanks for your contribution!

PR #2611, opened 2019-04-08
Update InstagramMedia and Utils to accept ints and not just floats

diegomansua:
Updated the three places in code where a float was expected but an int would work just fine:

- Sticker rotation
- Min and max aspect ratio in media
- Story poll tally font size


szokeptr:
Why don't you just cast the `int` to a `float` before passing it to the library? This seems a bit awkward if you ask me, and from my experience the app is always sending floats.

diegomansua:
@szokeptr 

It's because we interact with this library from another service via a generic JSON proxy/facade. Even if we send 1.0 the `json_decode` function will deserialize the number as an integer. This simple change would solve our issues, and I'd say sending an integer in those cases would work just fine.

szokeptr:
Just use `JSON_PRESERVE_ZERO_FRACTION` option when encoding the data (`json_decode` preserves floating points by default, so it's probably `json_encode` that's stripping it in your case).
I don't think the library should be changed for this, it's obvious that these values should always be floats, it's up to you to provide them properly.

diogofcunha:
@szokeptr Diego just pointed out that the library is consumed as JSON RPC, there is no way to do that in javascript (https://stackoverflow.com/questions/5520399/how-to-prevent-removing-decimal-point-when-parsing-json).

You're assuming the library is always consumed by other PHP clients and that is not always the case. Most languages do not handle floats vs integers the same way php does, and converting between floats and integers should be handled by serialization and not require dirty hacks on the consumer side. I don't see nothing wrong with this change as integers are an acceptable parameter for this requests

JRoy:
You cannot just send an int to Instagram's servers it would be invalid and/or create a vector for Instagram to detect a non genuine session.

szokeptr:
@JRoy Yes, that's exactly my point here. This should be handled outside of the library, no matter how you consume it.

diegomansua:
The library doesn't seem to be using JSON_PRESERVE_ZERO_FRACTION. So `json_encode(1.0)` will be serialized as 1 anyway. Or am I missing something?

diogofcunha:
@JRoy and also @szokeptr 

This a snippet from a request made from the library ATM:

`[1;33mDATA: [{"internal_features":"polling_sticker","story_polls":"[{\"x\":0.45810314932767165,\"y\":0.27902291500796184,\"width\":0.6986,\"height\":0.112,\"rotation\":0,\"question\":\"\",\"viewer_vote\":0,\"viewer_can_vote\":true,\"is_sticker\":true,\"tallies\":[{\"text\":\"Tally 1 \",\"count\":0,\"font_size\":28},{\"text\":\"TOMORROW\",\"count\":0,\"font_size\":26.25}]}]","`

As you can see the **mentioned fields are sent as integers anyway**. And it's up and running and never caused any issues. I can't understand the resistance on this.

diegomansua:
In Javascript (and thus JSON) there's no distinction between integers and floats, they're all floating point numbers.

Since PHP's `json_decode` seems to be deserializing numbers without decimal part as integers rather than floats (I guess as an optimisation as in most platforms an int will be 32 bit and a float/double 64) the library should accept integers as well IMO. And as shown above the float 1.0 will become 1 anyway.

JRoy:
This is something that should be handled client side, not library side. Just convert your int to a float. I'm too scared that accepting this PR will add unintended consequences that I'm not thinking of. 

Regards.

diegomansua:
@JRoy please reconsider. I don't think what you just said is a good enough reason to disregard this PR.

This change will:

- Enable others to use the library in different ways and thus expand its reach.
- It's even an optimisation when using it in traditional PHP code. Why am I forced to use a 64 bit float to specify font size 25 when I can simply use a 32 bit integer, considering that the float 25.0 will be serialised as 25 when sending the request to Instagram servers?

diegomansua:
Can't we simply compare the requests sent to Instagram API when:

- using a float `25.0` in the current master branch
- using an integer `25` with this PR code

I'm fairly certain you will find out they're identical.

szokeptr:
@diegomansua that doesn't mean it's correct. I am now researching what is the app doing in case the values are without precision.

diogofcunha:
@szokeptr what do you mean by "correct"? That's just **how JSON works**, the app will do the same because they are **REST apis**. I'm shocked with this tbh. Please read the json spec at https://www.json.org/

szokeptr:
@diegomansua I just did some tests and the Instagram app is NOT converting such values to an `int`.

`..."story_polls":"[{"x":0.50046295,"y":0.5002787,"z":0,"width":0.58981484,"height":0.10981048,"rotation":0.0,"viewer_vote":0,"viewer_can_vote":true,"tallies":[{"text":"YES","count":0,"font_size":35.04762},{"text":"NO","count":0,"font_size":35.04762}],"is_shared_result":false,"finished":false,"is_sticker":true}]",...`

This is from a request made by the actual Instagram app. Check the "rotation" value, it's clear that they are not converting the float 0.0 to int 0.

This means that the library has to be updated to use JSON_PRESERVE_ZERO_FRACTION when such values can occur.

diegomansua:
If that's the case then yes the app seems to be using some JSON library (Gson probably I guess) and it's preserving decimals. Was that response captured by some proxy?

Now adding JSON_PRESERVE_ZERO_FRACTION _is_ a scary change I guess...

szokeptr:
I used Charles Proxy to capture that request. I am not sure about the implications of using JSON_PRESERVE_ZERO_FRACTION, let's wait for others to comment on that. (@JRoy, @valga, @mgp25 )

diegomansua:
In any case if that's how things are I can change my PR so that an int is cast to a float when it's needed. Would you consider that?

szokeptr:
The only change this library needs is adding that flag to `json_encode`, the rest should be handled on your side.

JRoy:
I would consider a pull request with the `JSON_PRESERVE_ZERO_FRACTION` although I'd have to do a lot of testing because I'm not aware of how much that would unintentionally change other things.

PR #2617, opened 2019-04-11
fix(location): use correct param type for markStoryMediaSeen

cojj90:
A quick fix: Wrong param type was being used for `markStoryMediaSeen`.

To be clear, current `markStoryMediaSeen` does not work at all because `LocationFeedResponse` does not have StoryTray object in it (correctly).

I'm quite sure `markStoryMediaSeen` should use `LocationStoryResponse` as this response contains a StoryTray object. I tested it and it works!


mgp25:
Does all properties matches?

cojj90:
> Does all properties matches?

yes it does - one quick example: 
`$locationFeed->getStory() instanceof Response\Model\StoryTray`

LocationStoryResponse -> has getStory method as shown in the signature (https://github.com/mgp25/Instagram-API/blob/master/src/Response/LocationStoryResponse.php#L12)

but 

LocationFeedResponse doesn't have it (https://github.com/mgp25/Instagram-API/blob/master/src/Response/LocationFeedResponse.php)

cojj90:
I also made the call and checked the shape of Instagram response matches the response model in this PHP lib

PR #2682, opened 2019-06-03
Instagram: Fix usernameHandler being unused

super-ienien:
Hi,

this PR fix the $usernameHandler argument of the finishTwoFactorLogin method and also add this argument to the sendTwoFactorLoginSMS method.

The behavior of the finishTwoFactorLogin method was wrong when using $usernameHandler, the $usernameHandler variable was never added to the request.


JRoy:
Thanks for your contribution!

Issue #2712, opened 2019-06-19
Wrong import Realtime/Mqtt.php


RainGrid:
## Prerequisites
- You will be asked some questions and requested to provide some information, please read them **carefully** and answer completely.
- Put an `x` into all the boxes [ ] relevant to your issue (like so [x]).
- Use the *Preview* tab to see how your issue will actually look like, before sending it.
- Understand that we will *CLOSE* (without answering) *all* issues related to `challenge_required`, `checkpoint_required`, `feedback_required` or `sentry_block`. They've already been answered in the Wiki and *countless* closed tickets in the past!
- Do not post screenshots of error messages or code.

---

### Before submitting an issue make sure you have:
- [ x] [Searched](https://github.com/mgp25/Instagram-API/search?type=Issues) the bugtracker for similar issues including **closed** ones
- [ x] [Read the FAQ](https://github.com/mgp25/Instagram-API/wiki/FAQ)
- [x ] [Read the wiki](https://github.com/mgp25/Instagram-API/wiki)
- [x ] [Reviewed the examples](https://github.com/mgp25/Instagram-API/tree/master/examples)
- [ x] [Installed the api using ``composer``](https://github.com/mgp25/Instagram-API#installation)
- [ x] [Using latest API release](https://github.com/mgp25/Instagram-API/releases)

### Purpose of your issue?
- [ x] Bug report (encountered problems/errors)
- [ ] Feature request (request for new functionality)
- [ ] Question
- [ ] Other

---

I'm writing a new function (blurred border). What I did:
1) composer require mgp25/instagram-php dev-master
2) cd vendor/mgp25/instagram-php
3) composer update -o
4) composer codestyle
Got this error on the last step:
Fatal error: Cannot use BinSoul\Net\Mqtt\Message as Message because the name is already in use in C:\xampp\htdocs\instagramapi-dev\vendor\mgp25\instagram-php\src\Realtime\Mqtt.php

This is a duplicate Message definition , because the class is defined in Realtime\Message.php in the same namespace.

RainGrid:
@JRoy I can import BinSoul\Net\Mqtt\Message as MqttMessage and edit all uses in this class. Should I create a PR?
And I have a problem with line endings, this automatically changes during the codestyle operation.

JRoy:
@RainGrid You can do that as a PR

As for line endings, use the codestyle line endings.

JRoy:
[...]

PR #2713, opened 2019-06-19
Realtime: Fix import conflict (#2712)

RainGrid:
Fix import conflict Realtime\Message with BinSoul\Net\Mqtt\Message


JRoy:
Thanks for your contribution!

PR #2804, opened 2019-08-06
Added ability to send replayable stories to user's inbox

sh2ezo:
As I can see Instagram application can send 'replayable' and 'disappearing' stories to user's direct inbox, but library can send only disappearing versions. So I decided to add functionality for replayable versions.


JRoy:
Thanks for your contribution!

PR #2814, opened 2019-08-19
There was double content field.

Miljan9602:
Double content field when mapping json in Creative model


JRoy:
Thanks for your contribution!

PR #2815, opened 2019-08-19
Direct: Add photo validation

sh2ezo:
Hello!

As I can see the validation is performed for video, video story and photo story, but not for simple photo. So I decided to add this missing validation code.


src/Request/Direct.php
```diff
@@ -578,10 +580,11 @@ public function sendPhoto(
         $photoFilename,
         array $options = [])
     {
-        if (!is_file($photoFilename) || !is_readable($photoFilename)) {
-            throw new \InvalidArgumentException(sprintf('File "%s" is not available for reading.', $photoFilename));
```
JRoy: Good change but this check should still be here

JRoy:
Thanks for your contribution

PR #2832, opened 2019-08-26
Updated sig key and version codes

charlie759:
Maybe it's helpful?


mgp25:
Thanks but it would break a lot of things since it is tied to a lot of requests and many more things. So it can't be pushed.

I still appreciatte your intention 😄 

Best regards

PR #2856, opened 2019-09-04
Remove duplicate has_more in FBSearchResponse

SHELA:
Remove has_more key from array


JRoy:
Thanks for your contribution

PR #2878, opened 2019-09-18
Facebook share & thumbnail image file for videos

frmeran:
Add the ability to share the post to Facebook Business Page & add the ability to select a thumbnail image file for videos


mgp25:
How you share in Facebook if you are not authenticated ?

frmeran:
User needs to enable the option first on the official Instagram's App. After that it will be able to post via the API by just adding share_to_facebook to the request.

JRoy:
Please add that `can_crosspost_without_fb_token` as you can't trust the user to do that.

heitorsampaio:
what do you mean in 'adding share_to_facebook to the request' can you where is this request? like this one '$ig->story->uploadPhoto' ?

Cheers!

PR #2887, opened 2019-09-19
update FelixShare model

Jho00:
When we integrated this library to our Instagram Direct depented project we found this bug. Instagram always respond with one Item object in this property. With this configuration Json-mapper not worked.

PR solve this problem


JRoy:
Thanks for your contribution!

PR #2905, opened 2019-10-01
Story::getReelsMediaFeed edit of returned value

aerdnar:
current version of getReelsMediaFeed returns a broken lazyjsonmapper structure because of unpredictable property's name, this is a problem which require a workaround like:

`$reelsMedia = $ig->story->getReelsMediaFeed([userIDs or highlight IDs])`
                            `->getReels()->asArray();`
`foreach($ReelsMedia as $key=>$value){`
   `//if is highlight ID`
    `$highlights[$key] = new InstagramAPI\Response\Model\StoryTray($value);`
    `//if is UserID`
    `$stories[$key] = new InstagramAPI\Response\Model\StoryTray($value);`
`}`

now it can be done in this way:

`$reelsMedia = $ig->story->getReelsMediaFeed([userIDs or highlight IDs])->getReels();`
`$highlights  = $reelsMedia->getHighlights();`
`$stories = $reelsMedia->getStories();`

I've maintained backward compatibility with first method so no one have to edit his code


mgp25:
For this cases there is a way to do this in a cleaner way. @steveJobzniak did some cool things with the JsonMapper and you can actually use impredictible container to get values from within a unpredictible key.

Take as example: 

https://github.com/mgp25/Instagram-API/blob/master/src/Response/FriendshipsShowManyResponse.php

https://github.com/mgp25/Instagram-API/blob/master/src/Response/Model/UnpredictableKeys/FriendshipStatusUnpredictableContainer.php

Regards

aerdnar:
Thanks, I've should read this before! Sorry to bother you. 
https://github.com/mgp25/Instagram-API/blob/fa6057c16d7ab646abf2a943254edcce4ef3c04d/src/Response/Model/UnpredictableKeys/CoreUnpredictableContainer.php

If getReelsMediaFeed()->getReels()->getData()[ userID or highlight ID ] is a reel type, then there are two new properties that miss:
        'has_pride_media'      => 'bool', 
        'media_count'          => 'int',

PR #2909, opened 2019-10-01
Client, Debug, & Request: Debugger improvements 

JRoy:
This pull request adds a few new features to aid with debugging.

### Changes:
* The client class now stores the most recently processed `Request` object.
  * Useful for seeing which request fails in an exception try/catch without having to enable debug mode.
  * Adds a field in `Client` class: `_lastRequest` which can be modified with `Client#setLastRequest()` and fetched with `Client#getLastRequest()`
* Added `Request#getUrl()`
* You can now have debug logs be sent to a file rather than just the console.
  * Can be enabled by setting the `debugLog` static field in the `Debug` class to true.
  * The default file location can be changed from `debug.log` to anything else by changing the `debugLogFile` static field in the `Debug` class

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -36,7 +36,7 @@ INSERT YOUR CODE HERE
 ``` 
 
 ### Debug Log
-Please post your debug log in the section below; You can enable the debug log by using `new Instagram(true)` instead of `new Instagram()`:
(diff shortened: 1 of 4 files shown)


mgp25:
So far seems good, as I am seeing debug log file is only enabled if:

```
\InstagramAPI\Instagram\Debug::debugLog
```

Right?

Do you have a sample output of the debug log file?

JRoy:
@mgp25 correct; the debug log file looks exactly like the console output

mgp25:
@JRoy Seems good to me. To avoid people asking it would be recommended to add the steps in the issue template and the wiki. As per this PR, seems ready to be merged. This should be commited upstream as well. I will manage that if you want.

👍

PR #2912, opened 2019-10-03
Repair sendPhoto method for direct (Now you can use method without …

antonforwork:
- Repair sendPhoto method for direct (Now you can use method without any "challenge require" problem)
- php cs fixer based on .php_cs.dist


mgp25:
Hello @antonforwork,

Thanks for your collaboration, however the solution provided is not correct at all and will cause issues. If you need help with that part, feel free to contact with me and I will point you which parts should be changed in order to be everything correct

Best regards

antonforwork:
okay

which part of pr is wrong ?

wolrd:
@mgp25 ?

PR #2926, opened 2019-10-10
Settings: Add redis storage handler

Juan-de-Costa-Rica:
Redis in-memory key-value storage backend.

You need https://github.com/phpredis/phpredis for this class


JRoy:
Thanks for your contribution!

Issue #2910, opened 2019-10-02
Response\Model\User property error


disco-mayhem:
# Bug Report

---
### Description
Making a people->getInfoById request on certain profiles can lead to LazyJsonMapperException error.
The solution is to change "[is_call_to_action_enabled_by_surface](https://github.com/mgp25/Instagram-API/blob/master/src/Response/Model/User.php#L696)" model's property, however I am not sure which definition I should use (empty is just works well).

### Code
An example of response (redacted for user's privacy):
```{
  "user": {
    "pk": "0",
    "username": "x",
    "full_name": "x",
    "is_private": false,
    "profile_pic_url": "x",
    "profile_pic_id": "x",
    "is_verified": false,
    "has_anonymous_profile_picture": false,
    "media_count": 0,
    "geo_media_count": 0,
    "follower_count": 0,
    "following_count": 0,
    "following_tag_count": 6,
    "biography": "x",
    "can_link_entities_in_bio": true,
    "biography_with_entities": {
      "raw_text": "x",
      "entities": []
    },
    "external_url": "x",
    "external_lynx_url": "x",
    "has_biography_translation": true,
    "can_boost_post": true,
    "can_see_organic_insights": true,
    "show_insights_terms": false,
    "can_convert_to_business": false,
    "can_create_sponsor_tags": false,
    "can_be_tagged_as_sponsor": true,
    "total_igtv_videos": 1,
    "has_igtv_series": false,
    "total_ar_effects": 0,
    "reel_auto_archive": "on",
    "is_profile_action_needed": false,
    "usertags_count": 0,
    "usertag_review_enabled": false,
    "is_needy": false,
    "is_interest_account": true,
    "has_recommend_accounts": false,
    "has_chaining": true,
    "hd_profile_pic_versions": [
      {
        "width": 320,
        "height": 320,
        "url": "x"
      },
      {
        "width": 640,
        "height": 640,
        "url": "x"
      }
    ],
    "hd_profile_pic_url_info": {
      "url": "x",
      "width": 1080,
      "height": 1080
    },
    "has_placed_orders": false,
    "can_tag_products_from_merchants": false,
    "show_shoppable_feed": false,
    "shoppable_posts_count": 0,
    "can_be_reported_as_fraud": false,
    "merchant_checkout_style": "none",
    "show_conversion_edit_entry": false,
    "aggregate_promote_engagement": true,
    "allowed_commenter_type": "any",
    "is_video_creator": false,
[...]

PR #2938, opened 2019-10-15
Models: Update User Model

disco-mayhem:
Fix 'is_call_to_action_enabled_by_surface' property. Bug #2910


JRoy:
Thanks for your contribution!

PR #2944, opened 2019-10-19
Redis Storage Handler: Add ability to set hash prefix

Juan-de-Costa-Rica:
The previous implementation used plain Instagram usernames to name the hashes of our stored settings data. This new implementations prepends `InstagramAPI:` or a user definable string to the user name.

Advantages of this change are:

1. Avoids conflicts of hashes being named the same by someone else
2. Allows for hashes made by InstagramAPI to be searched for and retrieved all at once without knowing all the specific hash names. This is great for when you need to migrate your settings between to or from another storage type.

This should be the last for the Redis storage handler!


mgp25:
Seems okay, however we will keep this open for a couple of days in case any change its required.

Regards

JRoy:
Thanks for your contribution!

PR #2966, opened 2019-10-28
Settings: Add postgres storage handler

fanatov37:


docker/php.ini
```diff
@@ -0,0 +1,16 @@
...
+memory_limit = 536870912M
+
+xdebug.remote_enable=1
+xdebug.remote_handler=dbgp
+xdebug.remote_port=9000
+xdebug.remote_autostart=1
+xdebug.remote_connect_back=0
+xdebug.idekey=docker
+;IMPORTANT!!! need use your currently host. V
+xdebug.remote_host=10.254.254.254
```
JRoy: Does this need to be hardcoded/included?

docker/Dockerfile
```diff
@@ -29,9 +41,4 @@ RUN wget https://johnvansickle.com/ffmpeg/releases/ffmpeg-release-amd64-static.t
 RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/bin/ --filename=composer
 
 # Set the WORKDIR to /app so all following commands run in /app
-WORKDIR /var/www/html
-
-COPY . ./
-
-# Install dependencies with Composer.
-RUN composer install --prefer-source --no-interaction
```
JRoy: Why were theses removed

src/Settings/Storage/Components/PDOStorage.php
```diff
@@ -267,7 +267,7 @@ public function openUser(
 
         // Retrieve and cache the existing user data row if available.
         try {
-            $sth = $this->_pdo->prepare("SELECT id, settings, cookies FROM `{$this->_dbTableName}` WHERE (username=:username)");
+            $sth = $this->_pdo->prepare("SELECT id, settings, cookies FROM {$this->_dbTableName} WHERE (username=:username)");
```
JRoy: Why was this changed?

src/Settings/Storage/Components/PDOStorage.php
```diff
@@ -247,7 +247,7 @@ public function deleteUser(
     {
         try {
             // Just attempt to delete the row. Doesn't error if already missing.
-            $sth = $this->_pdo->prepare("DELETE FROM `{$this->_dbTableName}` WHERE (username=:username)");
+            $sth = $this->_pdo->prepare("DELETE FROM {$this->_dbTableName} WHERE (username=:username)");
```
JRoy: Why was this changed?

src/Settings/Storage/Components/PDOStorage.php
```diff
@@ -227,7 +227,7 @@ public function moveUser(
             }
 
             // Now attempt to rename the old username column to the new name.
-            $sth = $this->_pdo->prepare("UPDATE `{$this->_dbTableName}` SET username=:newusername WHERE (username=:oldusername)");
+            $sth = $this->_pdo->prepare("UPDATE {$this->_dbTableName} SET username=:newusername WHERE (username=:oldusername)");
```
JRoy: Why was this changed?

src/Settings/Storage/Components/PDOStorage.php
```diff
@@ -192,7 +192,7 @@ public function hasUser(
         $username)
     {
         // Check whether a row exists for that username.
-        $sth = $this->_pdo->prepare("SELECT EXISTS(SELECT 1 FROM `{$this->_dbTableName}` WHERE (username=:username))");
+        $sth = $this->_pdo->prepare("SELECT EXISTS(SELECT 1 FROM {$this->_dbTableName} WHERE (username=:username))");
```
JRoy: Why was this changed?

src/Settings/Storage/Components/PDOStorage.php
```diff
@@ -159,10 +159,10 @@ protected function _setUserColumn(
             // Update if the user row already exists, otherwise insert.
             $binds = [':data' => $data];
             if ($this->_cache['id'] !== null) {
-                $sql = "UPDATE `{$this->_dbTableName}` SET {$column}=:data WHERE (id=:id)";
+                $sql = "UPDATE {$this->_dbTableName} SET {$column}=:data WHERE (id=:id)";
                 $binds[':id'] = $this->_cache['id'];
             } else {
-                $sql = "INSERT INTO `{$this->_dbTableName}` (username, {$column}) VALUES (:username, :data)";
+                $sql = "INSERT INTO {$this->_dbTableName} (username, {$column}) VALUES (:username, :data)";
```
JRoy: Why was this changed?

src/Settings/Storage/Components/PDOStorage.php
```diff
@@ -159,10 +159,10 @@ protected function _setUserColumn(
             // Update if the user row already exists, otherwise insert.
             $binds = [':data' => $data];
             if ($this->_cache['id'] !== null) {
-                $sql = "UPDATE `{$this->_dbTableName}` SET {$column}=:data WHERE (id=:id)";
+                $sql = "UPDATE {$this->_dbTableName} SET {$column}=:data WHERE (id=:id)";
```
JRoy: Why was this changed?

src/Settings/Storage/Postgresql.php
```diff
@@ -0,0 +1,74 @@
...
+
+namespace InstagramAPI\Settings\Storage;
+
+use InstagramAPI\Settings\Storage\Components\PDOStorage;
+use PDO;
+
+/**
+ * Class Postgresql.
+ */
+class Postgresql extends PDOStorage
```
JRoy: Should be `PostgreSQL`

src/Settings/Storage/Postgresql.php
```diff
@@ -0,0 +1,74 @@
...
+{
+    /**
+     * Postgresql constructor.
+     *
+     * {@inheritdoc}
+     */
+    public function __construct()
+    {
+        parent::__construct('Postgresql');
```
JRoy: Should be `PostgreSQL`

src/Settings/Storage/Components/PDOStorage.php
fanatov37: bc it will be error for postgres.
  **`** is wrong syntax for postgres

http://joxi.ru/5mdoggJU3WVpgr

src/Settings/Storage/Components/PDOStorage.php
fanatov37: bc it will be error for postgres.

src/Settings/Storage/Components/PDOStorage.php
fanatov37: for mysql and sqlite is OK without **`**

src/Settings/Storage/Postgresql.php
fanatov37: ok. gonna to rename this class

src/Settings/Storage/Postgresql.php
fanatov37: ok +

docker/php.ini
fanatov37: So I use mac  and this remote host for xdebug will be normal. 
Just need do alis like this 
`sudo ifconfig lo0 alias 10.254.254.254`

for ubuntu alias did not work and my collegues use dockerbridge ip.
cmd `ifconfig`. 

So I really don't know 1 way for mac and ubuntu xdebug settings **remote_host**

docker/Dockerfile
fanatov37: **COPY** unnecessary bc we use docker-compose **volumes**  

**composer install** use after **docker-compose up -d**

composer.json
```diff
@@ -47,6 +47,8 @@
         "ext-gd": "*",
         "ext-exif": "*",
         "ext-zlib": "*",
+        "ext-pdo": "*",
+        "ext-redis": "*",
```
JRoy: Please move these to the `suggest` section of the `composer.json` as they are not needed 100% of the time

docker/Dockerfile
```diff
@@ -19,7 +19,19 @@ RUN echo "deb http://deb.debian.org/debian jessie main" > /etc/apt/sources.list
     && docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ \
```
JRoy: Please keep the Dockerfile in the root directory of the repository

fanatov37: It's not problem I can move.

but we will use just **docker-compose**  in root directory
`docker-compose up -d`

it's better flow for project than run single DockerFile :)

So your choose ?

JRoy: I just don't want to break people's existing workflows with docker so if you could move it back that would be best :P

fanatov37: ok. I understand.
 no problem. 

I think in the future we will use only docker-compose :)

composer.json
fanatov37: +

docker/Dockerfile
fanatov37: `+`

JRoy:
Thanks for your contribution :)

Issue #2972, opened 2019-10-30
API - Post (photo, video, story) not working any more.


hoemon:
I know this does not fit the format but as far as I follow the issues, the recent issue (probobly about the same problem) is not getting any updates. I would love to know if anyone else is still having the same problem..

During the time since the api stopped working, I've realized a few things:

1. some old versions of the api is still working. For instance, for one of the acounts I have (that's using this api) I'm still using the old v4.1.0 and it seems to post just fine.

2. other private api's (python, nodejs, etc) seem to be working.

Are we a minority with this api failing to work since yesterday? Or is everyone having the same issues?

Can this be account related rather than api related?

Thank you.

bepossitive:
Got same situation with "Invalid request options" then upload photo. It was started yesterday, and that method not work for 1 hour.

messilao:
Same problem

bepossitive:
@hoemon yesterday it stops for hour, then resolved. Today again stops

kapil3790:
Same here. It stopped for about 7-8 hours then it started working for another 6-7 hours and it's again throwing the same error.

hoemon:
Hm, that's interestig. For me it was gone one time, and still not working. This means it may be account based?

bepossitive:
@hoemon tested on many accounts (different country's based), all the same error

hoemon:
Thank you for that information. Weirdly old API still works for me though.

bepossitive:
@hoemon plz share release version number, that works fine

usamaejaz:
I am also getting this issue again.

mgp25:
https://github.com/mgp25/Instagram-API/issues/2971#issuecomment-547562245

PR #2977, opened 2019-10-31
Bug for #2972 And update Sig Key

deoryz:
/upload/video/ not used by IG
just use resumable upload for video and image


mgp25:
Please contact with me for for info

PR #2984, opened 2019-11-02
Fix #2972

### LINKED ISSUE #2972: API - Post (photo, video, story) not working any more.

vostlerts:
The current code allows you to restore the script without any corrections.

Must be added to constants "EXPERIMENTS":
`ig_android_photo_fbupload_universe`


mgp25:
@vostlerts please contact with me. Best regards

PR #2989, opened 2019-11-04
Tests: Add phpunit.xml

fanatov37:
add phpunit.xml

after we can simple run phpUnit tests by cmd

`vendor/bin/phpunit --configuration /var/www/html/phpunit.xml`

after we will get output like this

http://joxi.ru/8234ddPU90OZXA


JRoy:
Thanks for your contribution!

Issue #3003, opened 2019-11-12
Video & Photo sizes constant might be incorrect


Mavyre:
# Question

### Description
It seems that the API is not in sync anymore with the requirements of Instagram in term of video and photo sizes. It now seems that it's minimum of 600 for both photo and videos, and maximum of 1080px for videos, and 1936px for photos.

Sources to confirm: 
https://blog.tailwindapp.com/required-instagram-video-sizes-and-dimensions/
https://louisem.com/37587/whats-the-best-instagram-image-size

Shall we do a Pull Request to update PhotoDetails and VideoDetails to conform to the new specs?

JRoy:
I'd you want to do a pull request, I'll likely merge it :p

Mavyre:
PR #3004 created! :)

JRoy:
Fixed in 0c81a1f22d40d863d59a2d214a31b743ffd31b42

PR #3004, opened 2019-11-12
Changed min and max width to fit Instagram new Specs

Mavyre:
As stated in issue #3003 ; sizes are wrong and it doesn't allow us to upload at max size. This pull request updates the new allowed sizes by Instagram


JRoy:
Thanks for the contribution :)

PR #3017, opened 2019-11-20
Hashtag: Migrate markStoryMediaSeen() to TagsStoryResponse

epicwinsolutions:
Fixed the `markStoryMediaSeen()` function. The `getFeed()` functionality was moved to `getSection()` which does not send stories. So the only option to get stories is `getStory()` function. The `markStoryMediaSeen()` was using the response of `getFeed()` function. Now its update to use the response of `getStory()` function.


JRoy:
Thanks for your contribution!

PR #3018, opened 2019-11-20
Hashtag: Correctly pass maxId to getSection() to getFeed()

epicwinsolutions:
Fixed the params passed to `getSection()` function from `getFeed()` function.

The signature of `getSection` function is `public function getSection($hashtag, $rankToken, $tab = null, $nextMediaIds = null, $maxId = null)` where the `$maxId` is the 5th param not the 4th. With this fix, `null` is sent to the 4th param (`$nextMediaIds`) and the `$maxId` is sent as 5th param.


JRoy:
Thanks for your contribution!

PR #3080, opened 2019-12-26
FFmpeg: Update to Symfony 5

mysamimi:
when i update composer,
my symfony/process updated from v4.4.1 to v5.0.2.
Process method input change 2 array from string
then I write this changes


JRoy:
Thanks for your contribution!

